PR #12, opened 2026-06-28
Heuristic call-resolution fixes + optional LSP path & divergence diagnostic

ashtom:
Three tree-sitter call-resolution fixes (each found by a new LSP divergence diagnostic), plus an optional, parallel LSP-backed resolver. The heuristic stays the default; the LSP is opt-in.

## Heuristic fixes (measured vs brain-bench's independent compiler oracles)
1. **Trait/interface-dispatched calls resolve to the declaration, not every impl.** A same-name call that fanned out to a trait declaration *plus* all its impls was the dominant false-positive source on trait-heavy code. Now narrows to the unique trait/interface declaration — same-file and cross-file. *Rust byteorder call-graph F1 0.49 → 0.74.*
2. **Construction is not a call.** A direct call resolving to a type (class/struct/enum) is object construction; the compilers never emit it as a call to the type symbol. Excluding type-like targets is a **uniform precision gain across all four languages, zero regressions**: Python httpx 0.68 → 0.82, TS zod 0.55 → 0.64, Go errors 0.77 → 0.82 (+12 more), Rust byteorder → 0.75 (== CBM).

## Optional LSP path (`internal/sem/lsp`)
- A minimal LSP client (framing, request correlation, server-request auto-ack, `$/progress` readiness with warmup fallback, per-request skip-on-stall) + a call-hierarchy resolver, with a server registry for rust-analyzer / gopls / typescript-language-server / pyright (degrades silently to the heuristic when absent).
- `entire-sem lsp-calls --repo . --language <lang>` emits the LSP-resolved call graph; `--diff` reports heuristic-vs-LSP divergence — the diagnostic that surfaced all three fixes above.

All four-language tests pass; new unit tests cover each resolver fix; the one affected golden is regenerated.


suhaanthayyil:
Closing this **for now** — it's not rejected, it's *stale against the merged stack* and needs a reconciliation + a benchmark gate before any of its always-on changes can land. The branch and all the work are preserved; reopen (or a fresh PR) anytime.

**What's genuinely good here and should come back as its own PR:** the optional LSP subsystem (`internal/sem/lsp/{client,resolver}.go`, `internal/cli/lsp.go`, the `lsp-calls` command) and the divergence diagnostic. They're ~930 of the 1071 lines, fully **off by default** (sole entry point is the opt-in `lsp-calls` subcommand, spawn-safe + timeout-bounded + read-only), so they can't regress the always-on graph. These should be split into a clean PR cut from current `main`.

**Why the rest can't merge as-is — three independent reasons:**

1. **It collides with the already-merged CONSTRUCTS feature (#10).** Both this branch and `main` rewrite `resolveCallTargets`, but with opposite policy for type-targeted calls. `main` (#10) **routes** type-targets into the `CONSTRUCTS` relation; this PR adds `typeLikeKind(to.Kind)` to the *skip* list and **discards** them. Merging this silently starves the CONSTRUCTS branch that's live in `main` — a regression that would pass as green if the goldens were regenerated to the new state.

2. **It changes the always-on graph by an unmeasured amount.** The `provider.go` cross-file trait/interface dispatch adds new edges at confidence **0.66** to every run (not opt-in). Whether that's net precision-positive is exactly what brain-bench measures — and it hasn't been run against this. Per the audit standard for this repo ("never a blind merge"), an unmeasured default-behavior change can't land.

3. **The branch predates #10, so the merge isn't even clean.** This `provider.go` doesn't contain the CONSTRUCTS code; a real merge conflicts in `resolveCallTargets`, and rebasing it onto current `main` would rewrite the commit history here.

**Path to revive:**
- **PR A (mergeable now):** split out the opt-in LSP subsystem + divergence diagnostic, cut from current `main`. No always-on behavior change → no benchmark needed.
- **PR B (gated):** reconcile the heuristic *on top of* #10's CONSTRUCTS routing (keep CONSTRUCTS, layer the trait-dispatch resolution), regenerate goldens, and land it behind a brain-bench precision pass.

Everything else in the stack — #9, #10, #11, #13, #15 — is merged into `main` and green.

PR #14, opened 2026-06-28
Resolve method calls on package vars of package-qualified types

ashtom:
Method calls on a package-level var whose type is package-qualified (`enc = json.Encoder{}`) were dropped when the bare type name is ambiguous across packages. zerolog defines `Encoder` in both `internal/json` and `internal/cbor` with identical method names, so the unique-method-name fallback couldn't choose and every `enc.AppendX` call was lost.

## Change
Collect package-level composite-literal var declarations per directory (a Go package spans a dir; the var is often declared in one file, used in another), then resolve the qualified type by the Go convention that a package's import alias equals its directory basename (`json.Encoder` → the `Encoder` in `.../json/`). Requires a **unique** match, so an ambiguous alias resolves to nothing rather than wrongly — the conservative guard that keeps it regression-free.

## Result
zerolog call-graph F1 **0.516 → 0.646** (recall 0.36 → 0.49, precision 0.93 → **0.95**). **No other Go repo changes** in the full sweep. Unit tests cover collection (incl. grouped `var (...)` blocks and function-body exclusion) and the unique-match resolver.

Stacks on the build-constraint PR (which makes `enc`'s declaration itself unambiguous); the two together take zerolog 0.51 → 0.65.

### DIFF
--- a/internal/sem/pkgvar_test.go
+++ b/internal/sem/pkgvar_test.go
@@ -1,0 +1,49 @@
+package sem
+
+import "testing"
+
+func TestCollectPackageVarTypes(t *testing.T) {
+	content := `package zerolog
+
+import "example.com/m/internal/json"
+
+var (
+	_   encoder = (*json.Encoder)(nil)
+	enc         = json.Encoder{}
+)
+
+var solo = &json.Encoder{}
+
+func useEnc() {
+	local := json.Encoder{} // inside a func body: must NOT be collected as a package var
+	_ = local
+}
+`
+	got := collectPackageVarTypes(content)
+	if qt, ok := got["enc"]; !ok || qt.alias != "json" || qt.typeName != "Encoder" {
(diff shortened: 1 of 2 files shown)


suhaanthayyil:
Closing note (this PR was auto-closed when its stacked base branch was deleted — recording why, for the trail): **superseded by #15**, which carries this exact change. #15 was opened from the same `feat/go-pkgvar-imported-type` branch, rebased onto the latest `main`, and merged on 2026-06-28 (merge commit `d83db6b`) — including the identical `collectPackageVarTypes`/`resolveQualifiedType` resolver, the `pkgQualType` type and `pkgQualVarRe` regex helpers, and the `internal/sem/pkgvar_test.go` tests (that test file is byte-for-byte identical to this PR's). All of that content is now in `main`, so no work is lost — it landed via #15 to avoid a duplicate merge.

PR #21, opened 2026-07-03
sem: fix C#/PHP call and definition extraction gaps found by the brain-bench C#/PHP batch

ashtom:
Five extraction bugs diagnosed from the brain-bench C#/PHP significance batch (dotnet/runtime, dotnet/roslyn, laravel/framework, WordPress/wordpress-develop), rebased onto main. Each fix commit carries a regression test.

## C#

1. **Vendor tree-sitter-c-sharp v0.23.5 at ABI 14** (`76d59ec`) — the bundled grammar predates C# 12/13; a single `params ReadOnlySpan<Task>` overload in runtime's `Task.cs` swallowed the enclosing `class Task` and every later member (177 of ~800 symbols). Corpus parse-failure counts: runtime 3920, roslyn 1593, efcore 945, aspnetcore 544. `Task.cs` now parses with zero ERROR nodes; roslyn's failures drop 1593 → 286. Masks for primary constructors and collection expressions are removed (grammar-native now); preprocessor masking stays, plus a new length-preserving mask for `*(T*)&x` pointer-deref casts, the one construct the new grammar still rejects.
2. **Out-declaration / nullable-local receiver types** (`8f9b3e4`) — `out HttpConnectionPool? pool` and `Type? local;` now feed receiver-type inference; parameter modifiers (`out/ref/in/params/this/scoped/readonly`) no longer corrupt signature-derived types.
3. **Static calls across partial classes** (`d01e5aa`) — type-name resolution prefers the partial declaration that actually defines the called method; roslyn's `Contract.ThrowIfFalse(...)` (declared in Contract.cs, shadowed by the lexically-first Contract.InterpolatedStringHandlers.cs) now resolves `type_inferred` 0.82.

## PHP

4. **Closure-body receiver inference** (`cfe0e78`) — an untyped receiver named after the enclosing type resolves to it (method-existence gated): laravel's `Container.getClosure` now emits CALLS to `Container.build` and `Container.resolve`.
5. **Ambiguous bare calls + docblock return types** (`568a423`) — ambiguous name-only calls emit ranked candidate edges instead of silence (WordPress `apply_filters`, 3 declarations); the chained-call regex family accepts `->`; `@return` docblock types drive `rest_get_server()->dispatch()` resolution; nullable `: ?Type` hints are stripped correctly.

Plus `1ec76d0`: regression tests pinning two bugs found by self-measuring the benchmark corpus (jq's C typedef-return misnaming, Go self-module package imports) whose fixes had already landed on main independently.

## Verification

Full `go test ./...` passes on the rebased branch. Brain-bench re-measurement with this binary: roslyn 17/18 → **18/18**, framework 20/21 → **21/21**, wordpress-develop 18/20 → **20/20** (runtime re-measure in flight). Score artifacts land in a separate brain-bench PR, cross-linked here.

https://claude.ai/code/session_01EaiZeA5H6EvrrG8jjtXTRB

### DIFF
--- a/internal/sem/call_scanners.go
+++ b/internal/sem/call_scanners.go
@@ -1,0 +1,360 @@
+package sem
+
+import (
+	"regexp"
+	"strings"
+)
+
+var (
+	clojureCallHeadRe       = regexp.MustCompile(`\(\s*([A-Za-z0-9_.$!?*+\-<>=/]+)`)
+	sqlRoutineCallRe        = regexp.MustCompile(`(?i)\b((?:@?[A-Za-z_][A-Za-z0-9_]*@?|"[^"]+")(?:\s*\.\s*(?:@?[A-Za-z_][A-Za-z0-9_]*@?|"[^"]+"))*)\s*\(`)
+	objectiveCMessageSendRe = regexp.MustCompile(`\[\s*([A-Za-z_][A-Za-z0-9_]*)\s+([A-Za-z_][A-Za-z0-9_]*)`)
+	jsDottedCallRe          = regexp.MustCompile(`\b([A-Za-z_$][A-Za-z0-9_$]*(?:\s*\.\s*[A-Za-z_$][A-Za-z0-9_$]*)+)\s*\(`)
+	fsharpDottedCallRe      = regexp.MustCompile(`\b([A-Za-z_][A-Za-z0-9_']*(?:\s*\.\s*[A-Za-z_][A-Za-z0-9_']*)+)\s*\(`)
+	fsharpDottedApplyRe     = regexp.MustCompile(`\b([A-Za-z_][A-Za-z0-9_']*(?:\s*\.\s*[A-Za-z_][A-Za-z0-9_']*)+)\s+[A-Za-z_({\["'0-9]`)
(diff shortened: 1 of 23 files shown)


ashtom:
Reviewed `334ac8f` (agentic benchmark call resolution). Full test suite passes, and the frozen C#/PHP batch surfaces can't regress — the only cross-cutting change (optional generics in `receiverCallRe`) is purely additive, and all five batch regression tests stay green.

Two precision flags worth watching on broad corpora (both language-scoped, both bounded by downstream resolution gates):

1. `jsCallableArgumentIdentifiers` (call_scanners.go) treats every bare identifier in a call's argument list as a potential callable. Common names that are both variables and workspace functions (`next`, `error`, `handler`) can mint false CALLS edges in large JS/TS repos.
2. `appendSQLDerivedDuplicateTargets` (provider.go) intentionally breaks the single-target invariant for SQL routines under migration/update paths, fanning out `name_only` 0.54 edges to same-qualified-name routines.

Neither blocks merge; the brain-bench verification pass runs against this head, so any score movement lands in the paired artifacts PR. The PHP `create`/`Factory` and SQL directory-name conventions are the most benchmark-shaped heuristics — fine as conventions, but worth generalizing if they accrete.

PR #36, opened 2026-07-16
fix(gitutil): stop misclassifying git errors as file-absent

### LINKED ISSUE #31: gitutil.ShowFile swallows genuine git errors for any path containing "Path"
## Summary

`gitutil.ShowFile` misclassifies genuine git errors as "file absent" whenever
the file path contains the substring `"Path"` (capital P). It returns
`("", false, nil)` — swallowing the real error — because the path is always
embedded in the error string.

## Root cause

`internal/gitutil/git.go`, `ShowFile` (line 289) classifies "file absent" by
substring-matching the error text:

```go
// internal/gitutil/git.go:291-296
if strings.Contains(err.Error(), "exists on disk, but not in") ||
	strings.Contains(err.Error(), "Path") ||        // <-- over-broad
	strings.Contains(err.Error(), "does not exist") ||
	strings.Contains(err.Error(), "not found") {
	return "", false, nil
}
```

But `run` formats every git error with the full argv, which includes
`rev+":"+path`:

```go
// internal/gitutil/git.go:448
return "", fmt.Errorf("%s %s: %s", name, strings.Join(args, " "), msg)
```

So for `ShowFile(ctx, repo, "BADREV", "src/PathHelper.go")` the error is
`git show BADREV:src/PathHelper.go: fatal: invalid object name 'BADREV'.`,
whose `PathHelper` matches `strings.Contains(err.Error(), "Path")`. The bare
`"Path"` clause is also redundant: its intended message (`fatal: Path 'X'
exists on disk, but not in 'REV'`) is already caught by the
`"exists on disk, but not in"` clause.

## Impact

`ShowFile` is the blob reader for the diff/dependents paths. `internal/sem/analyze.go`
treats `!ok` (its `beforeOK`/`afterOK`) as "the file didn't exist on that side",
so a **modified** file whose path contains `"Path"` that hits a transient error
(bad rev, IO error, context cancellation) renders as fully **added or deleted**
instead of surfacing the error; `dependents.go` silently skips it.

## Reproduction

Verified live against `main` (`385cc7e`) with a throwaway test in
`package gitutil` calling the unexported `ShowFile` against a temp repo
containing `src/PathHelper.go`:

```
RAW git error (PathHelper): "fatal: invalid object name 'BADREV'.\n"
Case1 (src/PathHelper.go, BADREV): content="" ok=false err=<nil>
    >>> real git error swallowed (ok=false, err=nil)
Case2 (src/helper.go, BADREV):     content="" ok=false err=git show BADREV:src/helper.go: fatal: invalid object name 'BADREV'.
    >>> non-Path path correctly returns err
```

Same bad revision, two paths: the one containing `"Path"` swallows the error;
the one without it surfaces it.

## Suggested fix

Do not classify absence by scanning free-form error text that includes the
path. Prefer one of:

- Check the git **exit status** / a structured signal rather than substrings.
- Match only git's actual "missing path/object" wording anchored away from the
  argv echo (e.g. build the classification from `stderr` alone, not the wrapped
  `err.Error()` that includes the path), and drop the redundant bare `"Path"`
  clause.

## Areas to update

- `internal/gitutil/git.go` — `ShowFile` (line 289); consider having `run`
  expose stderr separately so callers classify on stderr, not the path-bearing
  wrapped error.

## Suggested tests

- Bad revision on a path containing `"Path"` → returns a non-nil error (not
  `ok=false, err=nil`).
- A genuinely missing path still returns `("", false, nil)`.
- A path containing `"Path"` that legitimately does not exist still returns
  `("", false, nil)`.

karthik-rameshkumar:
## Summary

`ShowFile` in `internal/gitutil/git.go` classified a git failure as *file-absent* (returning `("", false, nil)`) whenever the error text contained `"Path"`, `"does not exist"`, `"not found"`, or `"exists on disk, but not in"`. But `run()` wraps errors as `git show REV:PATH: <stderr>`, so the path (argv) is **always** part of the error string. Any real error — bad rev, IO failure, context cancellation — on a path that happens to contain the substring `"Path"` (e.g. `src/PathHelper.go`) was silently swallowed as a missing file.

## Fix

Classify against git's **stderr only**, never the wrapped error that echoes the argv. Added a `runWithStderr` helper that returns stdout and stderr separately; `run()` now delegates to it and keeps its existing signature, so no other callers change. The redundant `"Path"` clause is gone (git's real capital-P message is already caught by `"exists on disk, but not in"`).

- Genuinely missing paths at a valid rev still return `("", false, nil)`.
- Real failures (including bad rev on a `"Path"` file) now propagate as a non-nil error.

Fixes #31

## Test evidence

Added `TestShowFileClassifiesErrorsByStderrNotPath` (in `internal/gitutil/git_test.go`), which commits `src/PathHelper.go` and asserts:
- `ShowFile(ctx, repo, "BADREV", "src/PathHelper.go")` returns a non-nil error (regression).
- A genuinely missing path at `HEAD` returns `("", false, nil)`.
- An existing file at `HEAD` returns its content, `ok=true`, `nil`.

Verification (all pass):
```
gofmt -s -w . ; gofmt -l internal   # empty
go vet ./...                          # ok
go build ./...                        # ok
go test -race ./internal/gitutil/     # ok
```

### DIFF
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
 	"os/exec"
 	"sort"
 	"strconv"
@@ -121,7 +120,11 @@ func GrepIndexMatches(ctx context.Context, repo string, patterns []string, maxPe
 		return []GrepMatch{}, nil
 	}
 	args = append(args, "--")
-	cmd := newCmd(ctx, repo, "git", args...)
+	// Preserve the caller's locale here. Unlike the other git commands in this
+	// package, `git grep -i` uses LC_CTYPE for non-ASCII case folding; forcing
+	// the C locale would make Unicode matches disappear.
+	cmd := exec.CommandContext(ctx, "git", args...)
+	cmd.Dir = repo
 	var stdout bytes.Buffer
 	var stderr bytes.Buffer
 	cmd.Stdout = &stdout
@@ -291,11 +294,13 @@ func ShowFile(ctx context.Context, repo, rev, path string) (string, bool, error)
 	// the argv (which includes rev+":"+path). Matching the full error text made
 	// any real failure on a path containing a marker substring (e.g. "Path" in
 	// src/PathHelper.go) look like a missing file, swallowing the error.
-	out, stderr, err := runWithStderr(ctx, repo, "git", "show", rev+":"+path)
+	// Peel the revision to a tree before resolving the path. Without the type
+	// constraint, a missing full object ID or a blob object can produce the
+	// same path-looking diagnostic as a genuinely absent file.
+	objectSpec := rev + "^{tree}:" + path
+	out, stderr, err := runWithStderr(ctx, repo, "git", "show", objectSpec)
 	if err != nil {
-		if strings.Contains(stderr, "exists on disk, but not in") ||
-			strings.Contains(stderr, "does not exist") ||
-			strings.Contains(stderr, "not found") {
+		if isMissingPathDiagnostic(stderr) {
 			return "", false, nil
 		}
 		msg := stderr
@@ -305,6 +310,15 @@ func ShowFile(ctx context.Context, repo, rev, path string) (string, bool, error)
 		return "", false, fmt.Errorf("git show %s:%s: %s", rev, path, msg)
 	}
 	return out, true, nil
+}
+
+func isMissingPathDiagnostic(stderr string) bool {
+	// ShowFile runs git under the C locale, so only classify Git's specific
+	// missing-path diagnostics. Broad substring checks can match a bad revision
+	// or an unrelated error merely because an argv value contains the phrase.
+	return strings.HasPrefix(stderr, "fatal: path '") &&
+		(strings.Contains(stderr, "' does not exist in '") ||
+			strings.Contains(stderr, "' exists on disk, but not in '"))
 }
 
 // BatchFileReader reads blobs from one revision through a persistent
@@ -451,7 +465,9 @@ func run(ctx context.Context, dir, name string, args ...string) (string, error) 
 	return stdout, nil
 }
 
-// newCmd builds the exec.Cmd every subprocess in this package runs through.
+// newCmd builds the exec.Cmd used by subprocesses whose diagnostics must be
+// stable. GrepIndexMatches intentionally preserves the caller's locale because
+// git grep uses LC_CTYPE for case folding.
 // It pins the subprocess locale to C (LC_ALL=C overrides LANG and any LC_*;
 // LANG=C is set as a belt-and-braces default) so git's stderr messages are
 // always the English ones our error classification matches — e.g. ShowFile's
@@ -459,7 +475,9 @@ func run(ctx context.Context, dir, name string, args ...string) (string, error) 
 func newCmd(ctx context.Context, dir, name string, args ...string) *exec.Cmd {
 	cmd := exec.CommandContext(ctx, name, args...)
 	cmd.Dir = dir
-	cmd.Env = append(os.Environ(), "LC_ALL=C", "LANG=C")
+	// Cmd.Environ observes Dir and updates PWD accordingly. Starting from
+	// os.Environ would leave child processes with the parent's stale PWD.
+	cmd.Env = append(cmd.Environ(), "LC_ALL=C", "LANG=C")
 	return cmd
 }
 
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -287,15 +287,22 @@ func FileCochanges(ctx context.Context, repo string, maxCommits int) ([]FileCoch
 }
 
 func ShowFile(ctx context.Context, repo, rev, path string) (string, bool, error) {
-	out, err := run(ctx, repo, "git", "show", rev+":"+path)
+	// Classify against git's stderr only, never the wrapped error that echoes
+	// the argv (which includes rev+":"+path). Matching the full error text made
+	// any real failure on a path containing a marker substring (e.g. "Path" in
+	// src/PathHelper.go) look like a missing file, swallowing the error.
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

- **[Medium]** `internal/gitutil/git.go:296-298` — classification is still a locale-fragile stderr string-match; confirmed live (git 2.54.0, `LC_ALL=fr_FR.UTF-8 git show HEAD:missing.go` → `fatal : le chemin 'missing.go' n'existe pas dans 'HEAD'`), which matches none of `"exists on disk, but not in"` / `"does not exist"` / `"not found"`. A genuinely-absent file under a non-English git locale now returns a hard error instead of `("", false, nil)`. This is **not a regression** — the old code was equally fragile for any filename not containing the accidental `"Path"` substring — so it doesn't block this PR, but it's exactly the gap Pass A/B asks about and it's a live, reproducible miss, not theoretical. Worth a follow-up: pin `cmd.Env` to `LC_ALL=C`/`LANG=C` on the git subprocess (cheap, deterministic), or use an exit-code-based existence check (e.g. `git cat-file -e rev:path`) instead of stderr text matching.

- **[Medium]** `internal/sem/analyze.go:33,39` and `internal/sem/dependents.go:65` (not touched by this PR, but directly affected by it) — these callers do `if err != nil { return Result{}/nil, err }` on `ShowFile`. Before this fix, a real error on a `"Path"`-containing filename was silently swallowed as absent — wrong for that one file, but the batch continued. After this fix, that same error now aborts the *entire* `AnalyzeGitRange` call / the *entire* `buildReferenceIndex` build (discarding every delta/reference already collected for every other file in the commit range or repo). `analyze.go` already has a graceful-degradation convention a few lines below (a hard parse failure becomes a `result.Warnings` entry + `continue`, not an abort) — worth confirming whether a transient/single-file `ShowFile` failure (flaky IO, a tight `ctx` deadline hit deep in a large loop) deserves the same per-file isolation rather than an all-or-nothing hard fail. Not a bug in this diff's own two files, but a real widening of blast radius caused by it, worth a deliberate call.

- **[Low/Nit]** `internal/gitutil/git.go:298` — I could not get the `"not found"` arm to fire on real git 2.54.0 for any `ShowFile`-relevant failure (bad rev, missing path, uncommitted-but-on-disk path, corrupted/missing blob object, submodule gitlink path, blob-treated-as-tree traversal) — all land on `"does not exist"` / `"exists on disk, but not in"` or correctly propagate as real errors. May be a defensive catch for an older/different git message I didn't reproduce, but it's untested and possibly dead, in the same spirit as the redundant `"Path"` clause this PR just removed. `(speculative — did not exhaustively test across git versions)`

**Solid:**
- Root cause is correctly diagnosed and fixed exactly as scoped: classification now sees git's own `stderr` only, never the wrapped `run()` message that echoes `rev+":"+path` — verified this closes the reported hole (bad rev on `src/PathHelper.go` no longer swallowed).
- `run()`'s refactor onto `runWithStderr` is byte-for-byte behavior-preserving (same `TrimSpace`, same fallback-to-`err.Error()`, identical wrapped message format) for all 9 other call sites — zero blast radius there.
- Regression test directly encodes the bug report (bad-rev-on-Path-filename, genuinely-missing-file, existing-file) and CI is green on macOS/Linux/Windows for build+test+lint. Diff matches the stated +69/-14 with no hidden hunks.

**Overall:** Correct, minimal, well-tested fix for the reported bug — safe to merge. The two Medium items (residual locale-fragility, widened failure blast radius in two downstream callers) are pre-existing/consequential rather than introduced-and-broken, and worth tracking as follow-ups rather than blocking this PR.

karthik-rameshkumar:
Thanks for the audit review.

- Locale fragility: fixed in 9d0b0e2. All git subprocesses in `internal/gitutil` now go through a shared `newCmd` helper that appends `LC_ALL=C` and `LANG=C` to the inherited environment, so git stderr is always the English text the classifier matches regardless of the host locale. Added a unit test asserting the pinning wins over inherited `LC_ALL`/`LANG` on the constructed command.
- Downstream blast radius (per-file isolation in `analyze.go`/`dependents.go`): deliberately deferred to a follow-up rather than expanded here — PR #39 is currently touching the analyze.go warning path and changing it in this PR would conflict. Filed as follow-up scope.
- `"not found"` arm: kept as a defensive catch for older/other git versions where the phrasing differs, even if it is dead on modern git.

suhaanthayyil:
## Audit re-review — head `9d0b0e2` (read-only, no changes)

Re-checked after the locale pin landed. `go build` + `go test ./internal/gitutil/` green.

**Resolved since the last pass:** the `LC_ALL=C`/`LANG=C` pin on the git subprocess (`internal/gitutil/git.go:462`) closes the locale-fragility I raised — and it's genuinely correct, not just cosmetic. Go's `os/exec` runs `dedupEnv` (keeps the **last** occurrence per key) before `execve` (`exec.go:1223`), so `append(os.Environ(), "LC_ALL=C", "LANG=C")` reaches git as a single `LC_ALL=C` even when the parent env already sets `LC_ALL`, on Linux and macOS alike. No need to strip existing vars first.

Remaining:
- **[Med]** The widened-blast-radius point from my earlier comment still stands (unaddressed): `internal/sem/analyze.go:33,39` and `dependents.go:65` `return …, err` on `ShowFile`, so a single-file failure now aborts the whole `AnalyzeGitRange` / index build rather than being swallowed per-file as before. `analyze.go` already degrades hard *parse* failures to a warning + `continue`; worth a deliberate call on whether a transient single-file `ShowFile` error deserves the same per-file isolation. Design call, not a bug in this diff.
- **[Nit]** `internal/gitutil/git.go:298` — the `"not found"` arm still looks dead for `git show REV:PATH`: across bad-rev / missing-path / on-disk-not-in-rev / bad-object I could reproduce only `does not exist` and `exists on disk, but not in`, never `not found`. Harmless, same spirit as the redundant `"Path"` clause this PR removed. *(speculative across git versions)*

**Overall:** primary fix correct, locale concern now genuinely closed — safe to merge; the earlier blast-radius design call + one nit remain.

PR #37, opened 2026-07-16
fix(sem): disambiguate same-name overloads in semantic diff

### LINKED ISSUE #35: Semantic diff collapses same-name overloads (keys entities by Kind:Name only)
## Summary

The semantic diff collapses same-name overloaded functions/methods. It keys
entities by `Kind + ":" + Name`, so two overloads with the same name in one file
overwrite each other in the comparison map — only the last survives. Editing one
overload is therefore silently dropped or misattributed in the diff.

## Root cause

`internal/sem/analyze.go`, `compareEntities` (~line 260) builds
`beforeByKey`/`afterByKey` keyed by:

```go
// analyze.go ~385
func key(entity Entity) string {
	return entity.Kind + ":" + entity.Name
}
```

Two entities with the same `Kind` and `Name` collide; last-write-wins keeps only
the last one in file order. (The provider/symbols path disambiguates same-name
symbols by signature hash — see `provider.go` `entitySymbols` — but the diff path
does not.)

## Impact

For any language whose parser emits multiple same-name, same-`Kind` entities for
overloads, a change to one overload can vanish from `entire graph
diff/commit/checkpoint`, or be attributed to the wrong overload. A reviewer using
the diff "to judge whether a change is safe to keep" sees no change where a real
signature change occurred.

**Language-dependent reachability** (confirmed via the parser): C# and Java emit
two same-name `method` entities for overloads → **reachable**. TypeScript
collapses overloads to a single implementation entity at parse time → not
reachable via TS.

## Reproduction

Verified live end-to-end against `main` (`385cc7e`) with the real CLI (C#).

Base `C.cs`:
```csharp
public class C {
    public int F(int a) { return a; }
    public string F(string a) { return a; }
}
```

Head edits **only the first** overload (`int F(int a)` → `long F(long a)`);
`git diff` confirms exactly line 2 changed.

`entire graph diff --base HEAD~1 --head HEAD --json`:
```json
"changes": [
  { "type": "body_changed", "kind": "class", "name": "C",
    "before_start_line": 1, "after_start_line": 1, "dependents_count": 0 }
]
```
The `int F(int a)` → `long F(long a)` change is **missing** — no `method C.F`
change is reported at all.

Control (editing only the **second** overload, line 3) is reported correctly:
```json
{ "type": "signature_changed", "kind": "method", "name": "C.F",
  "old_signature": "public string F(string a)",
  "new_signature": "public double F(double a)",
  "before_start_line": 3, "after_start_line": 3, "dependents_count": 1 }
```

The surviving map entry is the last entity in file order (line 3), so editing
line 2 compares two byte-identical line-3 entities and emits nothing; editing
line 3 works only coincidentally. The `before_start_line: 3` on the survivor
confirms the line-2 overload never entered the map.

## Suggested fix

Disambiguate same-name/same-`Kind` entities in `compareEntities`, e.g. include
the signature (or a signature hash) in `key()`, matching how the symbols path
already disambiguates. Handle genuine duplicates (identical signature) with a
stable ordinal so both sides pair up deterministically.

## Areas to update

- `internal/sem/analyze.go` — `key` (~line 385) and `compareEntities`
  (~line 260); ensure removed/added pairing still works when a signature edit
  changes an overload's key.

## Suggested tests

- Two C#/Java overloads; edit the first → diff reports a
  `signature_changed`/`body_changed` for that overload (not just the enclosing
  class, not nothing).
- Edit the second overload → still reported (regression).
- Two overloads with identical signatures (true duplicates) pair up without
  spurious add/remove churn.

karthik-rameshkumar:
## Summary

`compareEntities` (internal/sem/analyze.go) keyed entities by `Kind:Name`, so two same-name, same-`Kind` overloads (reachable in C#/Java — not TypeScript) collided in `beforeByKey`/`afterByKey` under last-write-wins: only the last one in file order survived. As a result, editing the **first** overload's signature vanished from the diff (only the enclosing class showed `body_changed`); editing the second worked only coincidentally.

## Fix

Two-phase matching in `keyedEntityMaps` (which replaces the per-side `keyedEntities`; it takes both entity slices at once since matching needs both sides):

1. **Phase 1 — exact pairing.** Entities that share an exact `Kind:Name:Signature` across sides pair up first. True duplicates (an identical `Kind:Name:Signature` appearing multiple times on one side) are disambiguated by occurrence index in file order, so the Nth duplicate on each side pairs with the Nth on the other. This makes mid-list overload insertions, removals, and reorders inert for the untouched overloads instead of cascading phantom `signature_changed` events down the list.
2. **Phase 2 — positional fallback for leftovers.** Entities whose exact signature has no counterpart on the other side get positional ordinals per `Kind:Name` in file order. An in-place signature edit of one overload leaves exactly one leftover with that name on each side, so it still pairs and surfaces as `signature_changed` rather than remove+add.

Notes:

- We deliberately do **not** use a naive pure-signature key (signature always in the key, no positional fallback): that would regress the primary #35 repro (`F(int)`→`F(long)` with `F(string)` present) back to removed+added, because the rename reconciler's signature similarity (~0.33) is far below the 0.92 threshold.
- Phase-1 and phase-2 keys use distinct prefixes (`=` / `~`) so they can never collide.
- A lone, non-overloaded entity still pairs exactly as before (exact match when unchanged, sole leftover when edited), preserving prior single-entity behavior.
- Removed/added (present on only one side) still work: unmatched keys fall through to the existing deleted/added handling and rename reconciliation, which treat keys as opaque strings — keys are ephemeral and never persisted or emitted.

## Test evidence

Nine direct `compareEntities` regression tests (package `sem`, hand-constructed `[]Entity`, no grammar needed):

- `TestCompareEntitiesDisambiguatesSameNameOverloads` — #35 repro/regression guard: editing the FIRST of two `method C.F` overloads (`F(int)`→`F(long)`, `F(string)` untouched) yields exactly one `signature_changed` for the first overload, with no spurious remove/add of the second.
- `TestCompareEntitiesDetectsSecondOverloadEdit` — control: editing only the SECOND overload is reported correctly.
- `TestCompareEntitiesSingleEntityUnchangedBehavior` — regression: a single non-overloaded entity is still reported as before.
- `TestCompareEntitiesAddedOverloadReportedAsAdded` — tail-appended overload surfaces as exactly one `added`.
- `TestCompareEntitiesRemovedOverloadReported` — tail-removed overload surfaces as exactly one `removed`.
- `TestCompareEntitiesTrueDuplicatesEditOne` — identical `Kind:Name:Signature` duplicates; editing one body yields exactly one `body_changed` (occurrence-index pairing).
- `TestCompareEntitiesRemoveFirstOverloadNoCascade` — removing the FIRST of three overloads yields exactly one `removed F(int)` and no phantom `signature_changed` cascade (previously misattributed under pure-ordinal keying).
- `TestCompareEntitiesMidListInsertOnlyAdded` — mid-list insert yields exactly one `added`, nothing else.
- `TestCompareEntitiesReorderedOverloadsNoChanges` — pure reorder of identical overloads produces no events at all.

Verification (all pass):

```
gofmt -l  # clean for changed files
go vet ./internal/sem/...   # ok
go build ./...              # ok
go test ./internal/sem/...  # ok
go test ./...               # ok
```

Fixes #35

### DIFF
--- a/internal/sem/analyze.go
+++ b/internal/sem/analyze.go
@@ -366,21 +366,51 @@ func sortedEntities(byKey map[string]Entity) []Entity {
 
 func sortChanges(changes []EntityChange) {
 	sort.Slice(changes, func(i, j int) bool {
-		left := lineForSort(changes[i])
-		right := lineForSort(changes[j])
-		if left == right {
-			return fmt.Sprintf("%s:%s", changes[i].Kind, changes[i].Name) < fmt.Sprintf("%s:%s", changes[j].Kind, changes[j].Name)
+		left, right := changes[i], changes[j]
+		if leftLine, rightLine := lineForSort(left), lineForSort(right); leftLine != rightLine {
+			return leftLine < rightLine
 		}
-		return left < right
+		if left.Kind != right.Kind {
+			return left.Kind < right.Kind
+		}
+		if left.Name != right.Name {
+			return left.Name < right.Name
+		}
+		if left.Type != right.Type {
+			return left.Type < right.Type
+		}
+		if left.OldSignature != right.OldSignature {
+			return left.OldSignature < right.OldSignature
+		}
+		if left.NewSignature != right.NewSignature {
+			return left.NewSignature < right.NewSignature
+		}
+		if left.BeforeStartLine != right.BeforeStartLine {
+			return left.BeforeStartLine < right.BeforeStartLine
+		}
+		if left.AfterStartLine != right.AfterStartLine {
+			return left.AfterStartLine < right.AfterStartLine
+		}
+		if left.OldName != right.OldName {
+			return left.OldName < right.OldName
+		}
+		if left.NewName != right.NewName {
+			return left.NewName < right.NewName
+		}
+		if left.OldPath != right.OldPath {
+			return left.OldPath < right.OldPath
+		}
+		if left.NewPath != right.NewPath {
+			return left.NewPath < right.NewPath
+		}
+		if left.Reconciliation != right.Reconciliation {
+			return left.Reconciliation < right.Reconciliation
+		}
+		if left.Similarity != right.Similarity {
+			return left.Similarity < right.Similarity
+		}
+		return left.DependentsCount < right.DependentsCount
 	})
-}
-
-func key(entity Entity) string {
-	return entity.Kind + ":" + entity.Name
-}
-
-func sigKey(entity Entity) string {
-	return entity.Kind + ":" + entity.Name + ":" + entity.Signature
 }
 
 // keyedEntityMaps assigns each entity an ephemeral key such that entities that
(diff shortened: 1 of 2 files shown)


suhaanthayyil:

Scope: `internal/sem/analyze.go` `keyedEntities`/`compareEntities` + tests. Verified against issue #35's repro and the sibling `compound-v1` path in `provider.go`.

- **[Med]** `internal/sem/analyze.go:394` (`keyedEntities`) — The pure-positional-ordinal key (`Kind:Name#n`) is the direct cause of the "known limitation," and that limitation is worse than the PR frames it. A mid-list insert/remove/reorder doesn't just "misattribute" — it *cascades into phantom `signature_changed` events*. Concrete trigger: `before = [F(int), F(string), F(bool)]`, head removes the **first** overload → `after = [F(string), F(bool)]`. Result: `signature_changed F(int)→F(string)` **and** `signature_changed F(string)→F(bool)` **and** `removed F(bool)` — while the actual event (removed `F(int)`) is absent. The old bug produced false *negatives* (dropped changes); this produces false *positives* (claims signatures changed that never did), so "strict improvement" is inaccurate — it's a tradeoff. For a provenance/"is this change safe" tool, a phantom signature change is arguably as harmful as a miss.
  Suggested direction: a **two-phase match** rather than pure ordinal — first pair by exact `Kind:Name:Signature`, then assign positional ordinals only to the *leftover* unmatched same-name entities on each side. That kills every acknowledged limitation (reorder/insert/remove all match by exact signature regardless of position) while preserving the in-place-edit win (the one edited overload is the sole leftover and pairs positionally → `signature_changed`). Note: this is subtly *not* the same as literally mirroring `provider.go:entitySymbols` (sig-hash always in the key) — I confirmed a pure signature key would regress issue #35's primary repro to `removed`+`added`, since `jaccard("F(int)","F(long)") ≈ 0.33 < renameThreshold 0.92`, so the reconciler wouldn't re-pair it. The PR's rejection of a naive signature key is therefore correct; a two-phase match is what satisfies both goals.

- **[Low]** `internal/sem/analyze_test.go` — The new tests only exercise the *clean ends* of the ordinal scheme: `AddedOverloadReportedAsAdded` appends at the tail and `RemovedOverloadReported` drops the tail — exactly the cases pure-ordinal handles perfectly. The misattribution path (remove/insert **mid-list**, reorder) — the actual limitation — has no test pinning its behavior. Suggest a characterization test asserting the current (imperfect) mid-list output, so the limitation is visible in the suite and can't silently worsen.

- **[Nit]** PR description says "Added three direct `compareEntities` regression tests," but six were added (`Disambiguates…`, `DetectsSecondOverloadEdit`, `SingleEntityUnchangedBehavior`, `AddedOverloadReportedAsAdded`, `RemovedOverloadReported`, `TrueDuplicatesEditOne`) and the "Verification" block lists a narrower run. The extra coverage is a plus — just correct the count.

Solid areas (checked, no action):
- **Symbol-ID identity is not touched.** These keys are ephemeral — used only inside `compareEntities`/`bestRename`, never persisted and never emitted on `EntityChange`. The `compound-v1` stable IDs (`provider.go:entitySymbols`) are independent, so the "IDs survive ordinary edits" guarantee is unaffected. Good separation.
- **Blast radius correctly bounded.** `counts` is per-`Kind:Name`, so adding/removing a *differently*-named entity never shifts another name's ordinals; a non-overloaded entity is always `#0`, byte-for-byte the prior single-key behavior. Only genuine same-name overloads get ordinals > 0.
- Downstream deleted/added/rename handling correctly treats keys as opaque strings; `TrueDuplicatesEditOne` (identical-signature dup, body edit) is handled correctly — the one case where an ordinal is genuinely required.

Overall: a correctly-scoped fix that resolves #35's reported case with good test coverage and no risk to symbol identity; the one substantive gap is that a two-phase (exact-signature-then-ordinal) match would eliminate the acknowledged reorder/insert/remove misattribution instead of shipping it as a documented false-positive source.

karthik-rameshkumar:
Addressed the review in b4e0d5d, implementing the two-phase match exactly as suggested:

- **Phase 1:** entities pair across before/after by exact `Kind:Name:Signature`, with true duplicates (identical `Kind:Name:Signature` repeated on a side) disambiguated by occurrence index so the Nth duplicate on each side pairs with the Nth on the other.
- **Phase 2:** positional ordinals are assigned only to the leftover unmatched same-name entities on each side, so an in-place signature edit still pairs positionally and reports `signature_changed` — the #35 repro (`F(int)`→`F(long)` with `F(string)` present) is guarded by the existing `TestCompareEntitiesDisambiguatesSameNameOverloads`, avoiding the pure-signature-key regression you flagged (jaccard ~0.33 < 0.92 rename threshold).
- Distinct `=`/`~` key prefixes keep phase-1 and phase-2 keys from ever colliding; keys stay ephemeral/opaque (only used inside `compareEntities`/`bestRename`).

Your cascading repro — before `[F(int), F(string), F(bool)]`, head removes the first — now yields exactly one `removed F(int)` and nothing else, covered by `TestCompareEntitiesRemoveFirstOverloadNoCascade`. Also added `TestCompareEntitiesMidListInsertOnlyAdded` (mid-list insert → exactly one `added`) and `TestCompareEntitiesReorderedOverloadsNoChanges` (pure reorder → no events); `TestCompareEntitiesTrueDuplicatesEditOne` still passes via occurrence-index pairing.

PR description updated: test count corrected (nine `compareEntities` tests, not three) and the fix section now describes the two-phase matching. Verified: `go vet ./internal/sem/...`, `go build ./...`, `go test ./...` all pass; changed files gofmt-clean.

suhaanthayyil:
## Audit re-review — head `b4e0d5d` (read-only, no changes)

The two-phase overload matching addresses the phantom-`signature_changed` cascade I flagged — confirmed: exact `Kind:Name:Signature` pairs first, positional ordinals only for the leftovers, so reorder / mid-insert / mid-remove of untouched overloads are now inert and no entity is dropped (unique per-side keys, map size == entity count). `go test ./internal/sem/` green; the new tests fail on revert. Nice.

Remaining, non-blocking:
- **[Nit]** `internal/sem/analyze.go:434,439` — the key separators `:`/`#` are interpolated into `Name`/`Signature` unescaped. The only real aliasing vector is a `:` at the Name/Signature boundary (not `#` in a signature — the ordinal suffix is pure digits, so Rust `#[inline]` etc. can't collide). No surveyed language emits `:` in a Name (Obj-C selectors are reduced to the colon-free head), so reachability is ~nil. A non-textual separator would close it cheaply.
- **[Low] (pre-existing, out of scope)** `bestRename`/`reconcileMoves` iterate Go maps in randomized order, so on an exact-similarity tie the rename-vs-add/remove classification can vary run-to-run; this PR slightly widens the candidate pool (overloads now preserved). Output ordering stays deterministic (`sortedEntities`/`sortChanges`). Worth a follow-up to sort candidates before the greedy match.

**Overall:** correct, resolves #35 without regressing the single-entity path or renames — safe to merge. **0 High, 0 Med, 1 Low, 1 Nit.**

PR #38, opened 2026-07-16
fix(sem): guard expandGraphCandidates against slice-bounds panic

### LINKED ISSUE #32: graph search panics (slice bounds [-1:0]) on a graph-expansion candidate past EOF
## Summary

`entire graph search` panics with `slice bounds out of range [-1:]` and crashes
when a graph-expansion candidate's target symbol has a `StartLine` beyond the
end of the file as read during search. `expandGraphCandidates` is missing the
`start == 0` guard that its sibling builder `makeSearchCandidate` already has.

## Root cause

`internal/sem/search.go`, `expandGraphCandidates` (~lines 960-1004):

```go
// search.go:981
start, end := clampRegion(symbol.StartLine, symbol.EndLine, len(lines))
// clampRegion returns (0,0) when symbol.StartLine > len(lines)
// ...
snippetStart, snippetEnd := focusedSnippetRegion(start, end, start, options.MaxSnippetLines)
// focusedSnippetRegion(0,0,0,n) returns (0,0)
// ...
Snippet: strings.Join(lines[snippetStart-1:snippetEnd], "\n"),  // search.go:1000 → lines[-1:0] PANIC
```

The sibling `makeSearchCandidate` (~line 831) performs the identical clamp and
then bails out:

```go
start, end = clampRegion(start, end, len(lines))
if start == 0 {
	return searchCandidate{}, false
}
```

`expandGraphCandidates` clamps at line 981 but has no `if start == 0` bailout
before slicing at line 1000.

## Impact

A single out-of-range graph-relation target crashes the entire `search`
command (not just that candidate). Trigger: any expansion relation
(CALLS/IMPORTS/USES_TYPE/… with confidence ≥ 0.7) whose target symbol's
snapshot `StartLine` exceeds the current line count of the file read during
search — reachable via a parser emitting a start line past EOF, or worktree-mode
TOCTOU (file truncated between snapshot build and the graph read).

## Reproduction

Verified live against `main` (`385cc7e`). A `package sem` test drove
`expandGraphCandidates` directly: seed `"seed"` → `CALLS` (confidence 1.0) →
target symbol `StartLine=100` against a 3-line file.

```
panic: runtime error: slice bounds out of range [-1:] [recovered, repanicked]
...
github.com/entireio/entire-graph/internal/sem.expandGraphCandidates(...)
	.../internal/sem/search.go:1000 +0x...
FAIL	github.com/entireio/entire-graph/internal/sem
```

`clampRegion(100, 110, 3)` → `(0,0)` → `focusedSnippetRegion(0,0,0,40)` → `(0,0)`
→ `lines[-1:0]` panic at `search.go:1000`.

## Suggested fix

Add the same guard `makeSearchCandidate` uses, right after the clamp at
`search.go:981`:

```go
start, end := clampRegion(symbol.StartLine, symbol.EndLine, len(lines))
if start == 0 {
	continue // skip this expansion candidate; symbol is out of range for the read file
}
```

## Areas to update

- `internal/sem/search.go` — `expandGraphCandidates` (after line 981).

## Suggested tests

- A graph relation whose target `StartLine > len(lines)` no longer panics and
  simply omits that candidate.
- Regression: a valid in-range target still produces its expansion candidate
  with a correct snippet.

karthik-rameshkumar:
## Summary

`entire graph search` panicked with `slice bounds out of range [-1:]` and
crashed whenever a graph-expansion candidate's target symbol had a `StartLine`
beyond the end of the file as read during search.

`expandGraphCandidates` (`internal/sem/search.go`) clamped the target region
with `clampRegion`, which returns `(0,0)` when `symbol.StartLine > len(lines)`.
Unlike its sibling builder `makeSearchCandidate`, it had no `start == 0` bailout,
so `strings.Join(lines[snippetStart-1:snippetEnd], "\n")` evaluated as
`lines[-1:0]` and panicked, taking the whole search command down.

## Fix

Add the same guard `makeSearchCandidate` already uses, right after the
`clampRegion` call: when `start == 0`, `continue` and skip that expansion
candidate. Minimal, style-matched, no behavior change for in-range symbols.

## Test evidence

Adds two regression tests in `internal/sem/search_test.go` driving
`expandGraphCandidates` directly:

- `TestExpandGraphCandidatesSkipsOutOfRangeSymbolLines` — a `CALLS` relation
  (confidence 1.0) to a target whose `StartLine=100` against a 3-line file. On
  the pre-fix code this test panics with `slice bounds out of range [-1:]`
  (verified by temporarily removing the guard); with the fix it returns cleanly
  and omits the out-of-range candidate.
- `TestExpandGraphCandidatesIncludesInRangeSymbol` — an in-range target still
  yields a candidate with a non-empty snippet.

Verification (all pass):
- `gofmt -s -w .` then `gofmt -l internal` (empty)
- `go vet ./...`
- `go build ./...`
- `go test ./internal/sem/`
- `go test -race -run 'TestExpandGraphCandidates...' ./internal/sem/`

Fixes #32

### DIFF
--- a/internal/sem/search.go
+++ b/internal/sem/search.go
@@ -979,6 +979,9 @@ func expandGraphCandidates(seeds []searchCandidate, relations []RelationRecord, 
 				contentCache[symbol.FilePath] = lines
 			}
 			start, end := clampRegion(symbol.StartLine, symbol.EndLine, len(lines))
+			if start == 0 {
+				continue
+			}
 			if end-start+1 > options.MaxRegionLines {
 				end = minInt(len(lines), start+options.MaxRegionLines-1)
 			}
--- a/internal/sem/search.go
+++ b/internal/sem/search.go
@@ -941,6 +941,16 @@ func expandGraphCandidates(seeds []searchCandidate, relations []RelationRecord, 
 	if len(seeds) == 0 || len(relations) == 0 {
 		return nil
 	}
+	// SearchRepository normalizes options before calling here, but direct
+	// callers (e.g. tests) may pass hand-built options. Mirror that
+	// normalization so a non-positive limit cannot shrink end below start
+	// and panic on the snippet slice below.
+	if options.MaxRegionLines <= 0 {
+		options.MaxRegionLines = defaultSearchMaxRegionLines
+	}
+	if options.MaxSnippetLines <= 0 {
+		options.MaxSnippetLines = defaultSearchMaxSnippetLines
+	}
 	seedScores := map[string]float64{}
 	for _, candidate := range seeds {
 		if len(seedScores) >= 10 {
--- a/internal/sem/search_test.go
+++ b/internal/sem/search_test.go
@@ -844,3 +844,77 @@ func containsString(values []string, want string) bool {
 	}
 	return false
 }
+
+func TestExpandGraphCandidatesSkipsOutOfRangeSymbolLines(t *testing.T) {
+	seeds := []searchCandidate{{
+		result: SearchResult{SymbolID: "seed", FilePath: "seed.go"},
+		score:  5,
+	}}
+	relations := []RelationRecord{{
+		FromID:     "seed",
+		ToID:       "target",
+		Type:       "CALLS",
+		Confidence: 1.0,
+	}}
+	symbolsByID := map[string]SymbolRecord{
+		"seed":   {ID: "seed", Name: "Seed", FilePath: "seed.go", StartLine: 1, EndLine: 3},
+		"target": {ID: "target", Name: "Target", FilePath: "target.go", StartLine: 100, EndLine: 101},
+	}
+	read := func(path string) (string, bool) {
+		if path == "target.go" {
+			return "line1\nline2\nline3", true
+		}
+		return "", false
+	}
+	options := SearchOptions{MaxRegionLines: 40, MaxSnippetLines: 40}
+
+	// Pre-fix, clampRegion returns (0,0) for the out-of-range target and the
+	// unguarded lines[snippetStart-1:snippetEnd] slice panics with
+	// "slice bounds out of range [-1:]".
+	out := expandGraphCandidates(seeds, relations, symbolsByID, read, nil, options)
+
+	for _, candidate := range out {
+		if candidate.result.SymbolID == "target" {
+			t.Fatalf("expected out-of-range target candidate to be skipped, got %+v", candidate.result)
+		}
+	}
+}
+
+func TestExpandGraphCandidatesIncludesInRangeSymbol(t *testing.T) {
+	seeds := []searchCandidate{{
+		result: SearchResult{SymbolID: "seed", FilePath: "seed.go"},
+		score:  5,
+	}}
+	relations := []RelationRecord{{
+		FromID:     "seed",
+		ToID:       "target",
+		Type:       "CALLS",
+		Confidence: 1.0,
+	}}
+	symbolsByID := map[string]SymbolRecord{
+		"seed":   {ID: "seed", Name: "Seed", FilePath: "seed.go", StartLine: 1, EndLine: 3},
+		"target": {ID: "target", Name: "Target", FilePath: "target.go", StartLine: 2, EndLine: 3},
+	}
+	read := func(path string) (string, bool) {
+		if path == "target.go" {
+			return "line1\nline2\nline3", true
+		}
+		return "", false
+	}
+	options := SearchOptions{MaxRegionLines: 40, MaxSnippetLines: 40}
+
+	out := expandGraphCandidates(seeds, relations, symbolsByID, read, nil, options)
+
+	found := false
+	for _, candidate := range out {
+		if candidate.result.SymbolID == "target" {
+			found = true
+			if candidate.result.Snippet == "" {
+				t.Fatalf("expected in-range target candidate to carry a snippet")
+			}
+		}
+	}
+	if !found {
+		t.Fatalf("expected in-range target candidate to be included, got %d candidates", len(out))
+	}
+}
--- a/internal/sem/search_test.go
+++ b/internal/sem/search_test.go
@@ -918,3 +918,52 @@ func TestExpandGraphCandidatesIncludesInRangeSymbol(t *testing.T) {
 		t.Fatalf("expected in-range target candidate to be included, got %d candidates", len(out))
 	}
 }
+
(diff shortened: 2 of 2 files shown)


suhaanthayyil:

No High or Med correctness issues. The guard is right, correctly placed, and I independently reproduced the original panic to confirm it: checked out the branch, stripped just the added `if start == 0 { continue }`, and re-ran `TestExpandGraphCandidatesSkipsOutOfRangeSymbolLines` — it panics with the exact reported `slice bounds out of range [-1:]` at `search.go:1000`(pre-fix)/`:1003`(current). Restoring the guard makes it pass. `go build`, `go vet`, `gofmt -l`, and `go test -race` all clean on the branch.

- **[Low]** `internal/sem/search.go:982-984` — the skip is completely silent (no log/metric) when a graph-expansion candidate is dropped for being out of range. Issue #32's own root-cause section names two real triggers (a parser emitting a bad `StartLine`, or worktree-mode TOCTOU/file truncation between snapshot build and search-time read) — this is now a second call site silently absorbing that signal. Matches `makeSearchCandidate`'s existing silent-skip precedent (not a regression), but a debug log/counter here would help notice if it's firing often in practice, which would point at a live indexing bug rather than a rare race. (speculative on actual frequency)

- **[Low] (speculative — traced, not reachable today)** `internal/sem/search.go:985-988` — the `MaxRegionLines` shrink step (`end = minInt(len(lines), start+options.MaxRegionLines-1)`) assumes `MaxRegionLines >= 1`. If it were ever `<= -1`, `end` can be pushed below `start`, `focusedSnippetRegion` passes that through unchanged (its own guard `end-start+1 <= maxLines` is satisfied by a negative value), and `lines[snippetStart-1:snippetEnd]` at `:1003` becomes a `low > high` slice — a different panic, on the same line this PR fixes, reachable even for an in-range symbol. Unreachable via the only production caller today: `SearchRepository` normalizes `MaxRegionLines`/`MaxSnippetLines` to positive defaults first (`:146-150`) before ever calling `expandGraphCandidates`. But `expandGraphCandidates` has no such defense of its own, and this PR's new tests already call it directly with hand-built `SearchOptions{...}` literals, bypassing that normalization — exactly the calling shape that would trip this if a future direct caller/test passed a non-positive value. Worth a one-line defensive clamp inside the function itself if it's meant to be safe to call standalone (as the new unit tests now do).

- **[Nit] (speculative, pre-existing, out of scope for this PR)** `clampRegion` (`:1478`) treats `StartLine <= 0` differently from `StartLine > len(lines)`: the latter now correctly returns `(0,0)` and gets skipped; the former is silently clamped to line 1 (`start = maxInt(1, start)` runs before the `start > lineCount` check) rather than skipped. A symbol with a zero-value/corrupt `StartLine` won't panic but will silently produce a line-1 snippet mis-attributed to it. Shared by all three `clampRegion` call sites (`:722`, `:831`, `:982`), so not introduced here — flagging only because it's the negative/zero-index case Pass B asks about.

**Solid:**
- Fix matches the sibling guard exactly, and now all three `clampRegion` call sites in the file (`candidatesForFile:722`, `makeSearchCandidate:831`, `expandGraphCandidates:982`) follow the identical `if start == 0 { continue/return }` pattern — good consistency, low risk of this being the wrong fix shape.
- Regression tests aren't vacuous: `TestExpandGraphCandidatesSkipsOutOfRangeSymbolLines` asserts the out-of-range `SymbolID` is actually absent from the output (not just "didn't crash"), and `TestExpandGraphCandidatesIncludesInRangeSymbol` asserts a non-empty snippet is still produced for the happy path — both verified passing, and the first verified to actually fail (panic) without the fix.
- `continue` is scoped to the inner `pairs` loop, correctly skipping only the one out-of-range direction/relation rather than the whole relation or all candidates; empty/all-skipped results flow safely downstream (`best` stays empty, `out` is an empty non-nil slice, safe to append).

Overall: correct, minimal, well-tested fix for the reported crash; findings above are secondary hardening/observability suggestions, not blockers.

**3 findings posted: 0 High, 0 Med, 2 Low, 1 Nit.**

karthik-rameshkumar:
Thanks for the review. Addressed the Low findings in 39813c2:

**Non-positive `MaxRegionLines` clamp** — `expandGraphCandidates` now mirrors `SearchRepository`'s option normalization (non-positive `MaxRegionLines`/`MaxSnippetLines` fall back to the package defaults) before the shrink step, so `end` can no longer drop below `start`. Added `TestExpandGraphCandidatesClampsNonPositiveRegionLines` covering `MaxRegionLines` of -1 and 0; without the clamp it reproduces the exact `slice bounds out of range [1:0]` panic, and it passes with the fix.

**Skip observability at the out-of-range guard** — deferred. `internal/sem` currently has no logging/metrics facility (grepped the package; the only `log` hits are test fixtures), and `makeSearchCandidate`'s identical skip is also silent. Rather than introduce a new logging dependency in this fix, I'd prefer a follow-up that instruments both skip sites together with whatever facility the package adopts.

**`clampRegion` StartLine <= 0 asymmetry** — acknowledged as a pre-existing behavior shared by all three call sites (clamped to line 1 rather than skipped); agreed it's out of scope here and worth a separate follow-up.

suhaanthayyil:
## Audit re-review — head `39813c2` (read-only, no changes)

The clamp resolves the standalone-caller panic I raised. Normalizing `MaxRegionLines`/`MaxSnippetLines <= 0` at the top (`internal/sem/search.go:947-953`) plus the `if start == 0 { continue }` guard (`:992`) make the sole variable-bound slice (`lines[snippetStart-1:snippetEnd]`, `:1013`) provably in-bounds for all inputs. I independently enumerated every index site; a standalone repro confirms both the original `[-1:]` panic and the `MaxRegionLines<0` `[1:0]` inversion, both now closed. `go test` green.

Two precision notes (nits, non-blocking):
- **[Nit]** the limit-normalization half (`:947-953`) is a no-op on the only production path — `SearchRepository` already coerces both limits (`:146-151`) before calling `expandGraphCandidates`. It's genuine defense-in-depth for the direct/test callers the new tests add; fine to keep. (The `start==0` guard, by contrast, *is* reachable in prod on stale-cache line drift — that half is not redundant.)
- **[Nit]** the `search_test.go` comment implies both `-1` and `0` panic pre-fix; only `-1` panics — `0` yields an inverted-but-valid empty slice caught by the `SnippetEndLine < SnippetStartLine` assertion. Assertions are correct regardless.

**Overall:** correct, minimal, both panic classes closed, no semantic change (`0` has always meant "use default") — safe to merge. **0 High, 0 Med, 0 Low, 2 Nit.**

PR #39, opened 2026-07-16
fix(sem): surface parse failures in diff instead of phantom removals

### LINKED ISSUE #30: graph diff/commit/checkpoint reports phantom `removed` entities and silently drops parse failures
## Summary

The semantic diff path (`entire graph diff`, `commit`, `checkpoint`) calls
`parser.Parse()`, which discards the `ParseStatus` returned by
`ParseWithStatus()`. When a changed file fails to parse — a tree-sitter parse
error, or the 5s parse timeout on a large/pathological file — the parser returns
**zero entities with no error surfaced**. The diff then reports every function,
type, and method in that file as `removed` (or `added`, if the failure is on the
base side), and emits **no** partial-failure or warning record.

This both produces a wrong diff and silently drops a parse failure, which
violates the contract in `AGENTS.md`:

> Unsupported or unparseable files must surface as machine-readable partial
> failures, never silent drops.

The snapshot/provider path (`snapshot`/`symbols`/`edges`) already handles this
correctly; only the diff path is affected.

## Impact

- `entire graph diff --base A --head B --json`, `entire graph commit HEAD --json`,
  and `entire graph checkpoint <id> --json` return semantically wrong results for
  any file that times out or hits a parse error.
- A consumer using these commands "to judge whether a change is safe to keep"
  (per `AGENTS.md`) sees a file that merely failed to parse as if its entire API
  surface was deleted — a false, high-signal alarm.
- The failure is invisible: no `warnings[]` entry, no partial-failure record, so
  a caller cannot tell the diff is degraded.

## Root cause

`internal/sem/analyze.go` (both `AnalyzeGitRange` and `AnalyzeCheckpoint`) parse
via the status-dropping wrapper:

```go
// internal/sem/analyze.go:45-46
beforeEntities, language := parser.Parse(oldPath, before)
afterEntities, afterLanguage := parser.Parse(path, after)
```

`Parse` throws the status away:

```go
// internal/sem/parser.go:148-151
func (TreeSitterParser) Parse(path, content string) ([]Entity, string) {
	entities, language, _ := TreeSitterParser{}.ParseWithStatus(path, content)
	return entities, language
}
```

But a parse failure is only reported *through* that dropped status:

```go
// internal/sem/parser.go:264-274
if err != nil || root == nil || root.IsNull() {
	code := "E_PARSE_ERROR"
	if err != nil && ctx.Err() != nil {
		code = "E_PARSE_TIMEOUT"
		// ...
	}
	return nil, spec.language, ParseStatus{ParseError: true, Code: code, Detail: detail}
}
```

So in `analyze.go`, `beforeOK`/`afterOK` (which come only from
`gitutil.ShowFile`, i.e. "did the blob exist") are `true`, the entity slice is
empty, and:

```go
// internal/sem/analyze.go:57
changes, removed, added := compareEntities(beforeEntities, afterEntities)
```

reports the surviving side's entire entity set as removed/added. Nothing is
appended to `result.Warnings` (only `reconcileMoves` at `analyze.go:72` ever
writes it).

## Reproduction

Verified live against a build of `main` (commit `4491676`).

```sh
git init -q repro && cd repro
# base: valid TS with two exported functions
printf 'export function alpha() { return 1 }\nexport function beta() { return 2 }\n' > slice.ts
git add slice.ts && git commit -qm base

# head: the SAME two functions are still present, but a leading broken type
# annotation makes tree-sitter recover to ZERO entities for the file
printf 'type Broken = <\n\nexport function alpha() { return 1 }\nexport function beta() { return 2 }\n' > slice.ts
git add slice.ts && git commit -qm head

entire graph diff --base HEAD~1 --head HEAD --json
```

**Observed** — both functions, which are *literally still present in the head
file*, are reported as `removed`, and there is **no `warnings` field**:

```json
{
  "base": "…", "head": "…",
  "files": [{
    "path": "slice.ts", "status": "M", "language": "TypeScript",
    "changes": [
      {"type": "removed", "kind": "function", "name": "alpha", "old_signature": "function alpha()", "before_start_line": 1, "dependents_count": 0},
      {"type": "removed", "kind": "function", "name": "beta",  "old_signature": "function beta()",  "before_start_line": 2, "dependents_count": 0}
    ]
  }]
}
```

For contrast, `entire graph snapshot --repo . --format ndjson` over the exact
same head tree reports the file correctly:

```json
{"record_type":"summary","languages":["TypeScript"],
 "partial_failures":[{"code":"E_PARSE_ERROR","severity":"warning","file_path":"slice.ts",
   "effect_on_semantic_completeness":"file parsed with syntax errors; semantic facts may be incomplete",
   "detail":"tree-sitter syntax error nodes present: error ERROR at line 1 column 1 near \"type Broken = <…\""}],
 "stats":{"files":1,"parsed_files":1,"symbols":0,"partial_failures":1,"completeness_level":"unsafe"}}
```

**Expected** — the diff should likewise report the file as a partial failure
(`E_PARSE_ERROR` / `E_PARSE_TIMEOUT`) and suppress the phantom `removed` entries.

The same shape reproduces on the base side (a base file that fails to parse
yields phantom `added` changes), and for the timeout path (a file large enough
to exceed the 5s `treeSitterParseTimeout` yields `E_PARSE_TIMEOUT`).

## Expected behavior

The diff path should mirror the provider path
(`internal/sem/provider.go:868-882`), which already maps a parse failure to a
`PartialFailure`:

```go
if parseStatus.ParseError {
	code := parseStatus.Code
	if code == "" { code = "E_PARSE_ERROR" }
	effect := "file parsed with syntax errors; semantic facts may be incomplete"
	if code == "E_PARSE_TIMEOUT" {
		effect = "file record emitted but symbol parsing skipped because parser time budget was exceeded"
	}
	failures = append(failures, PartialFailure{
		Code: code, Severity: "warning", FilePath: path,
		EffectOnCompleteness: effect, Detail: parseStatus.Detail,
	})
}
```

Concretely, `entire graph diff/commit/checkpoint` should:

- Switch `analyze.go` to `ParseWithStatus` for both the base and head blobs.
- When either side reports `ParseError`, **not** emit phantom `removed`/`added`
  changes derived from the unparseable side.
- Surface the failure as a machine-readable record. Two options:
  - reuse the existing `Result.Warnings []ProviderWarning` channel
    (`model.go:52`; `ProviderWarning` and `PartialFailure` are structurally
    identical), or
  - add a `partial_failures[]` field to `Result` to match the snapshot schema
    exactly. **Schema note:** adding a field is additive and allowed under the
    frozen `1.x` GA contract (`docs/adr/0001-ga-schema-contract.md`); confirm
    which shape downstream (Entire Brain) expects on the diff record.

## Areas to update

- `internal/sem/analyze.go` — `AnalyzeGitRange` (~line 45) and
  `AnalyzeCheckpoint` (~line 224): use `ParseWithStatus`, thread status into the
  per-file delta, suppress phantom changes on parse failure, and record the
  failure.
- `internal/sem/model.go` — if adding `partial_failures[]`, extend `Result`
  (and the text renderer in `writeText`).
- `internal/cli/root.go` — no change expected; it already serializes `Result`.

## Suggested tests

- Base parses, head is unparseable → result contains a partial failure for the
  file and **no** `removed` changes for its entities.
- Head parses, base is unparseable → partial failure and **no** `added` changes.
- Timeout path yields `E_PARSE_TIMEOUT`; syntax-error path yields
  `E_PARSE_ERROR`, with `detail` populated.
- A genuinely emptied file (valid, now zero entities) still reports real
  `removed` changes and **no** partial failure — i.e. the fix distinguishes
  "parse failed" from "legitimately empty".
- `AnalyzeCheckpoint` exercises the same behavior as `AnalyzeGitRange`.

karthik-rameshkumar:
## Summary

`AnalyzeGitRange` (and therefore `AnalyzeCheckpoint`, which delegates to it) called `parser.Parse`, discarding the `ParseStatus` returned by `ParseWithStatus`. When a changed file failed to parse (`E_PARSE_ERROR` / `E_PARSE_TIMEOUT`), the parser returned zero entities with no signal, so `compareEntities(before, nil)` reported **every** entity as a phantom `removed` change (or `added`, if the base side failed) with nothing appended to `result.Warnings`. This violated the guarantee that unparseable files surface as machine-readable partial failures, never silent drops.

## Fix

- Switch both base and head parsing in `AnalyzeGitRange` to `TreeSitterParser{}.ParseWithStatus`.
- When **either** side reports `ParseError` for a file, append a `ProviderWarning` (reusing the existing `Result.Warnings` channel, no schema change) and skip the delta for that file instead of emitting phantom changes. The warning mirrors the provider path's `parseStatus.ParseError → PartialFailure` mapping in `provider.go` (code, severity, effect-on-completeness wording, detail).
- Distinguish "parse failed" (`ParseError` true) from a legitimately emptied/zero-entity file (`ParseError` false): only the former is suppressed+warned; a validly-emptied file still reports its real `removed` changes.
- `AnalyzeCheckpoint` is covered automatically since it delegates to `AnalyzeGitRange`.

`writeText` already ignores `Warnings` (including the pre-existing move-ambiguity warnings), so text output is unchanged; `--json` serializes them via the existing `warnings` tag.

Fixes #30

## Test evidence

New regression test `TestAnalyzeGitRangeSurfacesParseFailures` (end-to-end via a temp git repo + `AnalyzeGitRange`) covers all three cases:
- base valid / head unparseable -> partial-failure warning, no phantom `removed` for alpha/beta.
- base unparseable / head valid -> warning, no phantom `added`.
- genuinely emptied file (valid, zero entities) -> real `removed` changes, no warning.

Verification (all pass):
- `gofmt -l internal` -> empty
- `go vet ./...`
- `go build ./...`
- `go test ./internal/sem/`
- `go test -race -run TestAnalyzeGitRangeSurfacesParseFailures ./internal/sem/`

### DIFF
--- a/internal/sem/analyze.go
+++ b/internal/sem/analyze.go
@@ -42,10 +42,24 @@ func AnalyzeGitRange(ctx context.Context, repo, base, head string, paths []strin
 			}
 		}
 
-		beforeEntities, language := parser.Parse(oldPath, before)
-		afterEntities, afterLanguage := parser.Parse(path, after)
+		beforeEntities, language, beforeStatus := parser.ParseWithStatus(oldPath, before)
+		afterEntities, afterLanguage, afterStatus := parser.ParseWithStatus(path, after)
 		if language == "" {
 			language = afterLanguage
+		}
+		// A file that fails to parse on either side yields zero entities with no
+		// signal, which would otherwise make compareEntities report every entity
+		// as phantom removed/added. Surface it as a machine-readable partial
+		// failure and skip the delta instead. Only ParseError suppresses the
+		// delta — a validly-emptied file (ParseError false, zero entities) must
+		// still report its real removed changes.
+		if beforeStatus.ParseError || afterStatus.ParseError {
+			status := afterStatus
+			if !status.ParseError {
+				status = beforeStatus
+			}
+			result.Warnings = append(result.Warnings, parseFailureWarning(path, status))
+			continue
 		}
 		if !beforeOK {
 			beforeEntities = nil
@@ -69,7 +83,7 @@ func AnalyzeGitRange(ctx context.Context, repo, base, head string, paths []strin
 		})
 	}
 
-	result.Warnings = reconcileMoves(deltas)
+	result.Warnings = append(result.Warnings, reconcileMoves(deltas)...)
 
 	for _, delta := range deltas {
 		changes := delta.changes
@@ -96,6 +110,28 @@ func AnalyzeGitRange(ctx context.Context, repo, base, head string, paths []strin
 		return Result{}, err
 	}
 	return result, nil
+}
+
+// parseFailureWarning builds the partial-failure warning emitted when a changed
+// file cannot be parsed on one side of the diff. It mirrors the provider path's
+// parseStatus.ParseError → PartialFailure mapping so both surfaces stay
+// consistent (see provider.go).
+func parseFailureWarning(path string, status ParseStatus) ProviderWarning {
+	code := status.Code
+	if code == "" {
+		code = "E_PARSE_ERROR"
+	}
+	effect := "file parsed with syntax errors; semantic facts may be incomplete"
+	if code == "E_PARSE_TIMEOUT" {
+		effect = "file record emitted but symbol parsing skipped because parser time budget was exceeded"
+	}
+	return ProviderWarning{
+		Code:                 code,
+		Severity:             "warning",
+		FilePath:             path,
+		EffectOnCompleteness: effect,
+		Detail:               status.Detail,
+	}
 }
 
 // fileDelta accumulates a file's resolved changes plus the removed/added
--- a/internal/sem/analyze.go
+++ b/internal/sem/analyze.go
@@ -47,15 +47,19 @@ func AnalyzeGitRange(ctx context.Context, repo, base, head string, paths []strin
 		if language == "" {
 			language = afterLanguage
 		}
-		// A file that fails to parse on either side yields zero entities with no
-		// signal, which would otherwise make compareEntities report every entity
-		// as phantom removed/added. Surface it as a machine-readable partial
-		// failure and skip the delta instead. Only ParseError suppresses the
-		// delta — a validly-emptied file (ParseError false, zero entities) must
-		// still report its real removed changes.
-		if beforeStatus.ParseError || afterStatus.ParseError {
+		// A hard parse failure yields ZERO entities with no signal, which would
+		// otherwise make compareEntities report every entity on that side as a
+		// phantom removed/added. Surface it as a machine-readable partial failure
+		// and skip the delta instead. We gate on zero entities rather than merely
+		// ParseError: tree-sitter also flags recoverable syntax errors
+		// (root.HasError) while still extracting a complete entity set, and those
+		// diffs are correct and must be kept. A validly-emptied file (ParseError
+		// false) is likewise never suppressed, so its real removed changes stand.
+		afterParseFailed := afterStatus.ParseError && len(afterEntities) == 0
+		beforeParseFailed := beforeStatus.ParseError && len(beforeEntities) == 0
+		if afterParseFailed || beforeParseFailed {
 			status := afterStatus
-			if !status.ParseError {
+			if !afterParseFailed {
 				status = beforeStatus
 			}
 			result.Warnings = append(result.Warnings, parseFailureWarning(path, status))
--- a/internal/sem/analyze.go
+++ b/internal/sem/analyze.go
@@ -47,23 +47,32 @@ func AnalyzeGitRange(ctx context.Context, repo, base, head string, paths []strin
 		if language == "" {
 			language = afterLanguage
 		}
-		// A hard parse failure yields ZERO entities with no signal, which would
-		// otherwise make compareEntities report every entity on that side as a
-		// phantom removed/added. Surface it as a machine-readable partial failure
-		// and skip the delta instead. We gate on zero entities rather than merely
-		// ParseError: tree-sitter also flags recoverable syntax errors
-		// (root.HasError) while still extracting a complete entity set, and those
-		// diffs are correct and must be kept. A validly-emptied file (ParseError
-		// false) is likewise never suppressed, so its real removed changes stand.
+		// A parse failure on either side degrades the diff. A TOTAL failure
+		// (ParseError with ZERO recovered entities) gives compareEntities no
+		// signal at all and would make it report every entity on that side as
+		// a phantom removed/added, so the delta is skipped and a
+		// machine-readable warning is surfaced instead. A PARTIAL recovery
+		// (ParseError with some entities extracted) keeps the diff — the
+		// recovered changes are real — but is still flagged with a warning,
+		// because symbols missing from the recovered set can surface as
+		// phantom removed/added. A validly-emptied file (ParseError false) is
+		// never suppressed or flagged, so its real removed changes stand.
 		afterParseFailed := afterStatus.ParseError && len(afterEntities) == 0
 		beforeParseFailed := beforeStatus.ParseError && len(beforeEntities) == 0
 		if afterParseFailed || beforeParseFailed {
-			status := afterStatus
+			status, warnPath := afterStatus, path
 			if !afterParseFailed {
-				status = beforeStatus
+				status, warnPath = beforeStatus, oldPath
 			}
-			result.Warnings = append(result.Warnings, parseFailureWarning(path, status))
+			result.Warnings = append(result.Warnings, parseFailureWarning(warnPath, status, true))
 			continue
+		}
+		if afterStatus.ParseError || beforeStatus.ParseError {
+			status, warnPath := afterStatus, path
+			if !afterStatus.ParseError {
+				status, warnPath = beforeStatus, oldPath
+			}
+			result.Warnings = append(result.Warnings, parseFailureWarning(warnPath, status, false))
 		}
 		if !beforeOK {
 			beforeEntities = nil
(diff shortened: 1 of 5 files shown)


suhaanthayyil:

Verified live against the PR head (`8cdaa82`) with throwaway tests driving `AnalyzeGitRange` over real temp git repos.

- **[High]** `internal/sem/analyze.go:58-59` — the gate `ParseError && len(entities) == 0` only catches a *total* parse failure. When tree-sitter recovers a **partial** entity set (`ParseError == true` but `len(entities) > 0`), the delta is not skipped, symbols missing from the recovered set are still emitted as phantom `removed`/`added`, and **no warning is appended**. The comment at :54-56 justifies the gate by asserting recoverable errors "still extract a complete entity set, and those diffs are correct" — that premise is the bug: error recovery routinely yields an *incomplete* set. Confirmed repro: base `svc.ts` = `alpha`+`beta` (clean); head keeps `alpha` but breaks `beta`'s declaration (e.g. `export function bet@@@ <<<`) → parse returns `ParseError=true, entities=[alpha]` → diff emits `removed beta` with `parseWarned=false`. This is the exact #30 class (false high-signal removal + silent parse failure) surviving for partial parses, and it diverges from the provider/snapshot path, which appends a `PartialFailure` on *any* `ParseError` (`provider.go:867`) regardless of entity count. Direction: append the warning whenever `status.ParseError` is true even when entities remain (keep the diff, but flag it degraded) rather than gating on zero entities. (The more severe "an *untouched* neighbor symbol is dropped" escalation I could not reproduce on TS — recovery kept both — so treat that extension as *(speculative)*; the corrupted-symbol case above is confirmed.)

- **[Med]** `internal/sem/analyze.go:119-131` — the effect strings are copied verbatim from `provider.go`, where on `ParseError` the file record and any partial symbols are *still emitted*. In the diff path the whole delta is dropped (`continue` at :66), so the wording misdescribes what happened: `E_PARSE_TIMEOUT` → "file record emitted but symbol parsing skipped" is literally false (nothing is emitted; the file's changes are fully suppressed), and `E_PARSE_ERROR` → "…semantic facts may be incomplete" understates a *fully absent* diff. The doc comment's claim that this "mirrors the provider path … so both surfaces stay consistent" is also only half-true given the High above (provider warns on any `ParseError`; diff warns only on zero-entity failures). Direction: use diff-specific wording (e.g. "file diff suppressed; changes omitted because the file could not be parsed") and soften the consistency comment.

- **[Med]** `internal/sem/text.go:10-33` (+ `root.go:461`) — text is the **default** output (`--json` is opt-in) and `writeText` never reads `result.Warnings`. Pre-fix, a parse-failed file at least showed (wrong) phantom removals; post-fix the file is `continue`d, produces no `FileChange`, and the warning is invisible in text mode — if it is the only changed file the user sees `"No semantic entity changes detected."`. So on the human-facing default surface a parse failure is now a *fully silent* drop, which is the outcome AGENTS.md forbids (JSON satisfies "machine-readable", but the default UX regressed from visible-wrong to invisible). Direction: render `result.Warnings` in `writeText`.

- **[Low]** `internal/sem/analyze.go:65` — on a rename where only the **before** side fails to parse, the warning is built with `path` (the new path); the unparseable blob is at `oldPath`. Minor mislabel in the machine-readable record. Direction: pass `oldPath` when `beforeParseFailed && !afterParseFailed`.

Solid areas (verified): the `ParseError`-only gate correctly distinguishes a genuinely-emptied file (`ParseError=false` → real `removed` changes kept, no warning) from a hard failure; empty-content parsing of a source path returns `ParseError=false`, so plain file **deletions are not masked** (confirmed: `git rm` of a clean file still reports its real removals); switching the move-reconcile write from `=` to `append(...)` correctly preserves parse warnings alongside `W_MOVE_AMBIGUOUS` (with a dedicated regression test); both-sides-unparseable yields exactly one warning; `AnalyzeCheckpoint` is covered via delegation; tests are genuinely end-to-end. Note (accepted tradeoff, not a bug): when a head genuinely deletes symbols *and* is unparseable, the real removals are suppressed in favor of the warning — unavoidable, and the warning signals the degradation.

Overall read: correct and well-tested for *total* parse failures and the emptied-file distinction, but the zero-entity gate leaves the **partial-recovery** case (the High) still phantom-removing with no signal — the same bug #30 targets, narrowed rather than fully closed.

karthik-rameshkumar:
Addressed all four findings in 4ddf044:

1. **[High] Partial parse recovery emitted phantom removals with no warning** — `AnalyzeGitRange` now warns on *any* `ParseError`, matching the provider path. A partial recovery (ParseError with a non-empty entity set) KEEPS the diff but appends a degraded-diff warning; a total failure (zero entities) keeps the existing behavior (delta suppressed + warning). The wrong comment claiming recoverable errors "still extract a complete entity set" is rewritten to describe the total/partial/valid-empty cases. Covered by `TestAnalyzeGitRange_PartialRecoveryKeepsDiffWithWarning`, which reproduces your exact scenario (base alpha+beta clean; head keeps alpha, corrupts beta with `export function bet@@@ <<<`) and includes a fixture guard asserting the corruption really yields `ParseError=true` with a partial entity set on this parser (verified: entities=[alpha]).

2. **[Med] Effect strings copied from the provider path were false for the diff path** — diff-specific wording now: suppressed case says "file diff suppressed; changes omitted because the file could not be parsed" (timeout variant: "...because parser time budget was exceeded"); the degraded-partial case says "file parsed with syntax errors on one side; diff kept but may be incomplete or contain phantom changes". The doc comment no longer claims the diff path "mirrors" the provider — it now states both warn on any ParseError but the provider emits partial output while the diff keeps (partial) or suppresses (total). Asserted in `TestParseFailureWarning_CodeMapping`.

3. **[Med] Warnings invisible in default text output** — `writeText` now renders `result.Warnings` (code, file path, effect, detail, in the existing text style), including when there are no file changes, so a suppressed parse failure never hides behind "No semantic entity changes detected." Covered by `TestWriteTextRendersWarningsWithoutFileChanges`, `TestWriteTextRendersWarningsAfterFileChanges`, and `TestWriteTextOmitsWarningSectionWhenNone`.

4. **[Low] Rename with before-side-only parse failure warned on the new path** — the warning now uses `oldPath` when only the before side fails (same for the new degraded branch). Covered by `TestAnalyzeGitRange_RenameBeforeSideParseFailureWarnsOldPath`, which also guards that git actually reports the rename.

The existing total-failure and phantom-suppression tests still pass; the prior soft-error test was updated to `TestAnalyzeGitRange_SoftErrorWithEntitiesKeptWithWarning` since that class now (correctly) carries a warning alongside its kept diff. `gofmt -s`, `go vet ./...`, `go build ./...`, and `go test ./...` are clean.

suhaanthayyil:
## Audit re-review — head `4ddf044` (read-only, no changes)

The partial-recovery warning + text rendering address the High/Med I raised: partial parses (`ParseError && >0 entities`) now keep the diff **and** emit a warning, and `writeText` renders `result.Warnings`, so a parse failure is no longer a silent drop on the default surface. Verified the dangerous direction holds — a cleanly-emptied valid file (`ParseError==false`) still reports its real removals, and every suppression path appends a warning before `continue`. `go test ./internal/sem/` green (118 targeted); fail-on-revert confirmed via `go test -overlay`.

One remaining item:
- **[Low-Med]** `internal/sem/model.go:47-53` — the diff `Result` carries parse failures only in `Warnings`, while the snapshot/provider path emits them under `partial_failures[]` (`provider.go:867-883`) and the diff `Result` has no `partial_failures` field at all. Same `E_PARSE_ERROR` event, different JSON key per surface — a consumer (e.g. Brain) that learned "parse failures live in `partial_failures[]`" from the snapshot schema won't find them on the diff record. Not a regression (the diff never emitted these before), but worth confirming the downstream contract, or adding the field for cross-surface consistency.

Also worth a line in the PR body: partial recovery still surfaces phantom `removed`/`added` for symbols lost to the grammar error — now warned + tested, but consumers should know `Files` on a warned file may contain them.

**Overall:** the total-failure and emptied-file cases are correct and well-tested; only the cross-surface schema consistency remains. Safe to merge once the `partial_failures` contract is confirmed.

PR #40, opened 2026-07-16
fix(sem): make static-accessor and java-module-import masks length-preserving

### LINKED ISSUE #33: TypeScript `static accessor` mask is not length-preserving, corrupting downstream entity offsets
## Summary

The TypeScript `static accessor` mask is not length-preserving. When a source
line contains `static  accessor(` with more than one space, the mask shrinks
the line, shifting every subsequent entity's byte offsets and corrupting their
name/signature/body-hash/fingerprint.

## Root cause

`internal/sem/parser.go`:

```go
// parser.go:395
tsStaticAccessorMethodPattern = regexp.MustCompile(`\bstatic\s+accessor(\s*\()`)

// parser.go:461
func maskTypeScriptStaticAccessorMethod(line string) string {
	return tsStaticAccessorMethodPattern.ReplaceAllString(line, "static accessoR${1}")
}
```

The pattern matches **one or more** whitespace (`\s+`) between `static` and
`accessor`, but the replacement hardcodes exactly **one** space. So
`static  accessor(` (two spaces, 16 bytes) is masked to `static accessoR(`
(15 bytes) — one byte shorter.

The parser builds the tree from the **masked** bytes (`parseSrc`) but extracts
entity text from the **original** `src` using the masked tree's byte offsets
(`walkEntities(root, src, …)`, `node.Content(src)`). A length change makes every
offset after the shortened line wrong. Sibling masks (Dart/OCaml/C#) explicitly
preserve length; this one does not.

Trigger note: this fires only for a static method literally **named**
`accessor` (i.e. `accessor` followed by `(`). The `static accessor x = 1`
field-modifier form does not match the pattern.

## Impact

A whitespace-only edit on a `static  accessor(` line silently corrupts the
name, signature, body-hash, and fingerprint of unrelated later entities in the
same file — producing spurious `body_changed`/wrong-fingerprint diff results
and wrong stored symbols.

## Reproduction

Verified live against `main` (`385cc7e`). Two snippets differing only in the
whitespace on line 2 (`bar()` on line 3 is byte-identical in both):

```ts
class C {
  static  accessor() { return 1; }   // vs "static accessor()" (one space)
  bar() { return 42; }
}
```

Observed entities via `TreeSitterParser{}.ParseWithStatus("c.ts", src)` (both
detected as TypeScript, no parse error):

```
one space  (correct):  method C.accessor  body=ecea99368c22b1e4
                       method C.bar        body=f5fe4ab71e764fe5
two spaces (corrupt):  method C.accesso    body=f8d12077a02cc0ba   ← trailing "r" dropped
                       method C.ba          body=f9982c691b5f27b7   ← unrelated bar mangled
```

The line-3 method, whose source is byte-for-byte identical in both files, is
reported as `C.bar` (body `f5fe4ab71e764fe5`) vs `C.ba` (body `f9982c691b5f27b7`)
— a one-byte shift corrupting every entity after the shortened line.

## Suggested fix

Make the mask length-preserving, mirroring the sibling masks: pad the
replacement to the exact matched length (e.g. capture the whitespace run and
reuse it, or replace only the fixed-length `accessor` token with an
equal-length sentinel). Add a general invariant/test that every TS mask
preserves `len(masked) == len(original)`.

## Areas to update

- `internal/sem/parser.go` — `maskTypeScriptStaticAccessorMethod` (line 461)
  and/or the pattern (line 395).

## Suggested tests

- `maskTypeScriptStaticAccessorMethod` output has the same length as its input
  for one-space and multi-space inputs.
- Parsing a file with `static  accessor(` (2+ spaces) followed by another
  member yields that member with the same name/body-hash as the one-space case.
- A property-style test asserting length preservation across all TS masks.

### LINKED ISSUE #41: Java `import module` mask is not length-preserving, corrupting downstream entity offsets
## Summary

The Java `import module` mask in `internal/sem/parser.go` is not
length-preserving. When an `import module` statement has more than one space
after `module`, the mask shrinks the line, shifting every subsequent entity's
byte offsets and corrupting their name/signature/body-hash.

This is the same bug class as #33 (the TypeScript `static accessor` mask); it
was found while adversarially reviewing the fix for #33.

## Root cause

`internal/sem/parser.go`:

```go
// line 617
javaModuleImportPattern = regexp.MustCompile(`^(\s*import\s+)module\s+`)

// line 625
text = javaModuleImportPattern.ReplaceAllString(text, "${1}       ")
```

The pattern matches `module` plus **one or more** whitespace (`\s+`), but the
replacement hardcodes a fixed 7 characters (`module ` = 7 bytes). So a single
space is length-preserving, but `import module   java.base;` (multiple spaces,
greedy `\s+`) collapses to 7 spaces and the line ends up shorter.

The parser parses the **masked** bytes but extracts entity text from the
**original** source using the masked tree's byte offsets, so any length change
shifts every offset after that line, corrupting the following entities.

## Impact

A `import module` declaration with irregular spacing silently corrupts the
name/signature/body-hash of subsequent entities in the same file — producing
spurious `body_changed`/wrong-fingerprint diffs and wrong stored symbols.
Verified empirically (via `ParseWithStatus`): with multiple spaces, `class C`
was extracted as entity `s`, and `C.bar` as `s.t b`.

## Fix

Length-preserving rewrite that captures the trailing whitespace and re-emits it
verbatim (6-char `module` mask + the original whitespace run), so it is
length-invariant for any whitespace count — mirroring the sibling masks.

## Status

Fixed in #40 (opened for #33), which audited every literal-replacement mask in
`parser.go`, found this second instance of the same class, and fixed both
together with regression tests:

- `TestMaskJavaModuleImportPreservesLength`
- `TestTreeSitterParserJavaModuleImportMaskPreservesFollowingEntity`

## Suggested follow-up

Add a single package-level invariant test asserting **every** TypeScript/Java
source mask in `parser.go` preserves length, so the next mask added can't
reintroduce this class of bug.

karthik-rameshkumar:
## Summary

Fixes #33.

`internal/sem/parser.go`'s TypeScript static-accessor mask matched one-or-more whitespace (`\s+`) between `static` and `accessor`, but the replacement hardcoded a single space:

```go
tsStaticAccessorMethodPattern = regexp.MustCompile(`\bstatic\s+accessor(\s*\()`)
// replacement: "static accessoR${1}"
```

So `static  accessor(` (two spaces) was rewritten to `static accessoR(` — one byte shorter. The parser parses the masked bytes but extracts entity text from the **original** source using masked-tree offsets, so any length change shifts every subsequent offset and corrupts each following entity's name/signature/body-hash. (Fires only for a static method literally named `accessor`, i.e. `accessor` followed by `(`.)

## Fix

Length-preserving rewrite: capture everything except the final `r` of `accessor` and rewrite only that char to `R`, preserving all matched bytes (and whitespace) verbatim — matching the sibling masks:

```go
tsStaticAccessorMethodPattern = regexp.MustCompile(`(\bstatic\s+accesso)r(\s*\()`)
// replacement: "${1}R${2}"
```

## Test evidence

Added regression tests in `internal/sem/parser_test.go`:

- `TestMaskTypeScriptStaticAccessorMethodPreservesLength` — asserts `len(output) == len(input)` for one-space and multi-space inputs.
- `TestTreeSitterParserTypeScriptStaticAccessorMaskPreservesFollowingEntity` — parses `static  accessor()` (two spaces) followed by `bar()`; asserts the following method is `C.bar` with the same name/signature/body-hash as the one-space variant.

Confirmed both tests **fail before** the fix (the following entity corrupts to `C.ba` / signature offset drift) and **pass after**.

Verification (all pass):

```
gofmt -l internal        # clean
go vet ./...             # ok
go build ./...           # ok
go test ./internal/sem/  # ok
go test -race -run 'TestMaskTypeScriptStaticAccessorMethodPreservesLength|TestTreeSitterParserTypeScriptStaticAccessorMaskPreservesFollowingEntity' ./internal/sem/  # ok
```

---

## Follow-up: second mask of the same class (found during review)

An adversarial review of this fix audited every literal-replacement mask in `parser.go` and found the **same length-changing bug** in `javaModuleImportPattern`:

```go
// pattern ^(\s*import\s+)module\s+  with a fixed 7-space replacement
```

`import module   java.base;` (multiple spaces after `module`, greedy `\s+`) shrank the line and corrupted every downstream entity's offsets (e.g. `class C` → entity `s`, `C.bar` → `s.t b`) — verified empirically via `ParseWithStatus`. Fixed the same way (capture the trailing whitespace, emit a 6-char `module` mask + verbatim whitespace), so it is length-invariant for any whitespace count.

Added tests: `TestMaskJavaModuleImportPreservesLength` and `TestTreeSitterParserJavaModuleImportMaskPreservesFollowingEntity`, plus extended the static-accessor length test with tab/mixed-whitespace inputs and a negative-match test.

Full `go test ./...` passes.

Fixes #41 (the Java `import module` mask, same bug class, found and fixed during this review).

### DIFF
--- a/internal/sem/parser.go
+++ b/internal/sem/parser.go
@@ -614,7 +614,7 @@ func rustTokenTreeStartsWithIf(src []byte, body *sitter.Node) bool {
 }
 
 var (
-	javaModuleImportPattern      = regexp.MustCompile(`^(\s*import\s+)module\s+`)
+	javaModuleImportPattern      = regexp.MustCompile(`^(\s*import\s+)module(\s+)`)
 	javaVarargsAnnotationPattern = regexp.MustCompile(`@[A-Za-z_][A-Za-z0-9_]*(?:\.[A-Za-z_][A-Za-z0-9_]*)*\s*\.\.\.`)
 )
 
@@ -622,7 +622,13 @@ func maskJavaUnsupportedSyntax(content string) string {
 	lines := strings.SplitAfter(content, "\n")
 	for i, line := range lines {
 		text, newline := splitLineEnding(line)
-		text = javaModuleImportPattern.ReplaceAllString(text, "${1}       ")
+		// Blank only the `module` contextual keyword (6 chars -> 6 spaces) and
+		// preserve the trailing whitespace group (${2}) verbatim, so the mask is
+		// length-preserving regardless of how much whitespace follows `module`.
+		// A literal replacement (e.g. 7 spaces) shortened multi-space imports and
+		// shifted every following entity's offsets, corrupting names/signatures/
+		// body-hashes (same bug class as the TS static-accessor mask).
+		text = javaModuleImportPattern.ReplaceAllString(text, "${1}      ${2}")
 		text = javaVarargsAnnotationPattern.ReplaceAllStringFunc(text, func(match string) string {
 			return strings.Repeat(" ", len(match)-3) + "..."
 		})
--- a/internal/sem/parser.go
+++ b/internal/sem/parser.go
@@ -392,7 +392,7 @@ func sourceLineAt(src []byte, line int) string {
 var (
 	tsKeywordTypePropertyPattern  = regexp.MustCompile(`^(\s*)in(\??\s*:)`)
 	tsTypeImportPattern           = regexp.MustCompile(`typeof\s+import\(([^)]*)\)`)
-	tsStaticAccessorMethodPattern = regexp.MustCompile(`\bstatic\s+accessor(\s*\()`)
+	tsStaticAccessorMethodPattern = regexp.MustCompile(`(\bstatic\s+accesso)r(\s*\()`)
 )
 
 func maskTypeScriptUnsupportedSyntax(content string) string {
@@ -458,7 +458,7 @@ func maskTypeScriptKeywordTypeProperty(line string) string {
 }
 
 func maskTypeScriptStaticAccessorMethod(line string) string {
-	return tsStaticAccessorMethodPattern.ReplaceAllString(line, "static accessoR${1}")
+	return tsStaticAccessorMethodPattern.ReplaceAllString(line, "${1}R${2}")
 }
 
 // rustItemWrapperMacroHint cheaply detects source that may contain a
--- a/internal/sem/parser.go
+++ b/internal/sem/parser.go
@@ -458,6 +458,11 @@ func maskTypeScriptKeywordTypeProperty(line string) string {
 }
 
 func maskTypeScriptStaticAccessorMethod(line string) string {
+	// Flip only the trailing `r` of `accessor` to `R`, reconstructing the
+	// surrounding whitespace from capture groups, so the replacement is
+	// byte-length-preserving: entity offsets are computed against the masked
+	// source but sliced from the original, so any length drift corrupts every
+	// following entity (same constraint as the Java module-import mask).
 	return tsStaticAccessorMethodPattern.ReplaceAllString(line, "${1}R${2}")
 }
 
--- a/internal/sem/parser_test.go
+++ b/internal/sem/parser_test.go
@@ -577,10 +577,47 @@ func TestMaskTypeScriptStaticAccessorMethodPreservesLength(t *testing.T) {
 		"  static accessor() { return 1; }",
 		"  static  accessor() { return 1; }",
 		"  static   accessor  () { return 1; }",
+		"  static\taccessor() { return 1; }",
+		"  static\t\taccessor () { return 1; }",
+		"  static \t accessor\t() { return 1; }",
 	} {
 		out := maskTypeScriptStaticAccessorMethod(in)
 		if len(out) != len(in) {
 			t.Fatalf("mask changed length: input %q (%d) -> output %q (%d)", in, len(in), out, len(out))
+		}
+	}
+}
+
+func TestMaskTypeScriptStaticAccessorMethodOnlyMasksStaticAccessorCalls(t *testing.T) {
+	t.Parallel()
+	// Must NOT mask: `accessor` as a plain identifier (not preceded by static),
+	// and `static accessor` with no following paren.
+	for _, in := range []string{
+		"  const accessor = 1;",
+		"  accessor();",
+		"  this.accessor(x);",
+		"  static accessor;",
+		"  static accessor = 1;",
+	} {
+		if out := maskTypeScriptStaticAccessorMethod(in); out != in {
+			t.Fatalf("unexpectedly masked non-static-accessor input %q -> %q", in, out)
+		}
+	}
+	// Must mask (r -> R before the paren) exactly once per occurrence.
+	for _, in := range []string{
+		"  static accessor() {}",
+		"  static  accessor () {}",
+		"  static\taccessor() {}",
+	} {
+		out := maskTypeScriptStaticAccessorMethod(in)
+		if out == in {
+			t.Fatalf("expected mask to fire on %q", in)
+		}
+		if len(out) != len(in) {
+			t.Fatalf("mask changed length: %q -> %q", in, out)
+		}
+		if !strings.Contains(out, "accessoR") || strings.Contains(out, "accessor") {
+			t.Fatalf("expected accessor->accessoR in %q", out)
 		}
 	}
 }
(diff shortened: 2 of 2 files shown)


suhaanthayyil:

Verified against PR head `3f02e88`: `gofmt`, `go vet ./...`, `go build ./...`, and `go test ./...` (672 passed / 19 packages) all clean. I also wrote and ran 9 additional adversarial tests directly against this code (multi-byte UTF-8 in a trailing comment and in a package identifier after `module`, a non-breaking-space separator, whole-file CRLF end-to-end for both masks, two same-line occurrences, exact-replacement-byte pinning, and EOF with no trailing newline) — all 9 passed. Nothing below rises above Low; this is a solid fix.

- **[Low]** `internal/sem/parser_test.go` — neither `TestMaskTypeScriptStaticAccessorMethodPreservesLength` nor `TestMaskJavaModuleImportPreservesLength` includes a CRLF input, even though `splitLineEnding` (the harness both masks run through) explicitly special-cases `\r\n`. The tab/mixed-whitespace cases you added are good; CRLF is the one line-ending variant the surrounding code treats differently and it's the one not exercised. I confirmed end-to-end it's actually fine (byte length preserved, and the following entity's `StartLine`/`Signature`/`BodyHash` match an LF control), but that took writing a new test to be sure — worth adding one so the next reader doesn't have to.
- **[Nit]** `parser.go:631` vs `:460-461` — the Java fix carries a good comment explaining why a literal replacement broke length and why this fix works; the TS fix has no comment at all. Same bug class, and `git blame` won't show the PR description — worth a one-liner on `maskTypeScriptStaticAccessorMethod` too.
- **[Nit] (speculative)** `parser.go:395,617` — both patterns lean on `\s` for the variable whitespace, and Go RE2's `\s` is ASCII-only (`[\t\n\f\r ]`), not Unicode whitespace. A non-breaking space (U+00A0) or similar between `static`/`accessor` or `import`/`module` won't match, so the mask silently no-ops and the pre-existing "tree-sitter chokes on this token" bug (predates this PR) could resurface for that exotic input. Confirmed not a corruption risk (no match → no transform → length trivially preserved), and not introduced by this PR — flagging only because Pass B specifically asked about non-ASCII bytes in the masked span.

**Solid:**
- Both replacements are *algebraically* length-preserving via capture-group reconstruction (`${1}` + a fixed-width swap + `${2}`), not just "usually" the same length: the swapped literal is always exactly as wide as the literal it replaces (`r`→`R`, 1 byte for 1 byte; `module`, 6 bytes, → 6 spaces), and `${1}`/`${2}` carry the actual matched whitespace through byte-for-byte regardless of run length.
- Neither mask can ever span a newline — not because the regex is newline-aware (`\s` matches `\n` fine), but because both call sites (`maskTypeScriptUnsupportedSyntax`, `maskJavaUnsupportedSyntax`) pre-split content into per-line strings via `strings.SplitAfter(content, "\n")` + `splitLineEnding`, stripping the line terminator before masking and splicing it back verbatim after. The mask never sees a `\n`/`\r\n` byte to begin with.
- Confirmed why byte-*length* (not byte-content) is the right invariant here: `ParseWithStatus` keeps the original `src := []byte(content)` (`parser.go:195`) and feeds a separately-masked `parseSrc` to tree-sitter (`parser.go:213` Java, `:237` TS); entity text/signature/hash are sliced from `src` at offsets computed against `parseSrc`. Only `len(parseSrc) == len(src)` is required for those offsets to stay valid — exactly what both fixes now guarantee for any whitespace width, which is also why the masked token itself still reports its true original spelling (e.g. `accessor`, not `accessoR`) despite the byte fed to the parser being different.
- Consistent with the rest of the file: spot-checked 3 other literal-replacement masks that looked similarly risky at a glance (`kotlinOverrideCallPattern` `"override()"`→`"masked__()"`, `cSharpBareAsyncArgumentPattern` `"async"`→`"true "`, `cPlusPlusUnsignedCastPattern` `"unsigned("`→`"uint32_t("`). All three swap a *fixed*-width literal for an equal-width literal (byte-counted: 10/10, 5/5, 8/8) — sound by construction. This PR's two masks were the actual outliers (variable-width `\s+`/`\s*` swallowed into a fixed-width replacement); the fix brings them in line with the file's established convention rather than inventing a new one.
- Adjacent same-line occurrences, zero-length spans (both patterns require ≥14 bytes to match at all, so this can't arise), and EOF with no trailing content all check out — no off-by-one found in either pattern's offset math.

**Overall**: correct, well-evidenced, low-risk — genuinely byte-length-preserving by construction (not by coincidence), verified by reasoning, by the PR's own new tests, and by 9 additional adversarial tests (including multi-byte UTF-8 and CRLF) run against the PR head. Only nits: a missing CRLF test case and an asymmetric code comment. Nothing blocking.

karthik-rameshkumar:
Addressed review feedback in bb35c1b:

- **CRLF coverage**: both `TestMaskTypeScriptStaticAccessorMethodPreservesLength` and `TestMaskJavaModuleImportPreservesLength` now include a CRLF case that runs multi-line `\r\n` content around the masked construct through the full per-line mask path (`maskTypeScriptUnsupportedSyntax` / `maskJavaUnsupportedSyntax`, i.e. through `splitLineEnding`'s `\r\n` special case), asserting byte length and line endings are preserved and the mask fired.
- **TS mask comment**: `maskTypeScriptStaticAccessorMethod` now carries the same explanation as the Java module-import mask — the replacement must be byte-length-preserving because entity offsets are computed against the masked source but sliced from the original.

On the ASCII-only `\s` point: agreed it's real but pre-existing — Go RE2 `\s` won't match exotic Unicode whitespace (e.g. U+00A0), so the mask silently no-ops there. That input already makes tree-sitter choke the same way it did before this PR, and since a no-op mask can't shift offsets there's no corruption risk, so leaving it as-is.

All CI checks green on bb35c1b (lint, build, test on ubuntu/macos/windows).

suhaanthayyil:
## Audit re-review — head `bb35c1b` (read-only, no changes)

The CRLF cases + the TS mask comment address what I flagged. Re-verified both fixed masks are byte-length-preserving for **all** inputs (property-tested: unicode around the match, CRLF, tabs, multi/single-space, two-per-line) — `${1}` + equal-width literal swap + `${2}`, invariant by construction, not by coincidence. I also independently scanned every pre-parse mask feeding the `walkEntities` offset path: all length-preserving; these two were the only outliers. `go test ./internal/sem/` green.

Remaining:
- **[Low] (pre-existing / out of scope — worth a separate issue)** `internal/sem/parser.go:786-808` `maskYAMLQuotedMappingKey` is **not** length-preserving (`line[:start]+"key"+line[end+1:]`, a fixed 5-byte `"key"` regardless of the original key length → drifts ±N). Symbols are safe — YAML dispatches to `yamlEntities` on the **original** content (`:281`), never masked-tree offsets. But the one surviving masked-source consumer, `parseErrorDetail(root, src)` (`:279` → `node.Content(src)` = unchecked `src[StartByte:EndByte]`), combined with **no `recover()` anywhere in `internal/sem`**, means a parse-erroring YAML file with short quoted keys inflating length + an error node near EOF can slice out of range → panic. Not the corruption class this PR targets — flagging for a follow-up issue.
- **[Nit]** the "audited every literal-replacement mask" note is slightly overstated — `maskYAMLQuotedMappingKey` is a third non-length-preserving literal replacement (benign per above). And the length invariant is doc-comment-only; a `testing.F` property test asserting `len(mask(x)) == len(x)` for arbitrary `x` would runtime-enforce it.

**Overall:** correct and complete for the corruption bug class — safe to merge; the YAML mask is a separate pre-existing latent panic worth its own issue. **0 High, 0 Med, 1 Low, 1 Nit.**

PR #55, opened 2026-07-23
feat(cli): report semantic diff progress

jdx:
https://entire.io/gh/entireio/entire-graph/trails/12

## Summary

- add `--progress` to `commit`, `diff`, and `analyze`
- report coarse `discover`, `parse`, `reconcile`, `dependents`, and `complete` progress to stderr
- preserve machine-readable stdout for text and JSON consumers
- clarify that `--include-file` re-includes ignored paths rather than acting as an allowlist
- clarify that streaming NDJSON aggregate metadata is authoritative in the trailing summary record

## Why

While testing Entire Graph against the mise repository, an unrestricted
`main..HEAD` semantic diff produced no output before timing out after 60 seconds.
PR #46 substantially improved common single-commit performance, but a large
range can still spend significant time building dependent counts with no
indication that the process is advancing.

With this change, the same workload reported:

```text
graph diff progress phase=parse files=580/580 elapsed=4.498s
graph diff progress phase=reconcile files=580/580 elapsed=4.498s
graph diff progress phase=dependents files=0/3049 elapsed=7.158s
graph diff progress phase=dependents files=100/3049 elapsed=20.4s
```

Progress is emitted only when requested and always goes to stderr.

The help notes address two assumptions found during the same field test. The
initial zero-valued NDJSON header is intentional under the streaming contract;
the trailing summary is authoritative. Likewise, include files contain
gitignore-style re-inclusion rules and do not restrict parsing to listed paths.

## Validation

- `mise run check`
  - gofmt
  - go vet
  - race-enabled test suite
  - build
- focused `go test ./internal/sem ./internal/cli`
- live `main..HEAD` progress run against mise

*This pull request was generated by an AI coding assistant.*

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -93,10 +93,10 @@ func printHelp(out io.Writer) {
 	fmt.Fprintln(out, `entire-graph adds entity-level context to Entire checkpoints.
 
 Usage:
-  entire graph commit [rev] [--json] [--repo path]
+  entire graph commit [rev] [--json] [--progress] [--repo path]
   entire graph checkpoint <checkpoint-id> [--json] [--repo path]
-  entire graph diff --base <rev> --head <rev> [--json] [--repo path] [-- path...]
-  entire graph analyze [--base <rev>] [--head <rev>] [--json] [--repo path] [-- path...]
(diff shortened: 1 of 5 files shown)


jdx:
tl;dr here is that the agent bailed thinking things were broken, putting some data onto stderr helps make it clear it's working and not hung

PR #66, opened 2026-07-28
Retrieval stack: complete-symbol payloads, related sites, literal clusters, bounded working-tree listing — plus withdrawal of the overstated savings claims

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/22

The accumulated agent-feedback + retrieval stack, reconciled with `main`. The three PRs based on this
branch (`pr/branch-aware-cache`, `pr2/search-topk-recall`, `pr2/graphify-lowhanging`) stack on top, so
review this one first.

## What is in it

- **Search returns editable material, not pointers.** The top hits come back as the complete body of
  their enclosing symbol (`complete-symbol`), so the common case is search -> edit with no follow-up
  file read. Tail hits degrade to two-line locators to pay for it.
- **RELATED SITES** — near-duplicate bodies, sibling implementations and call sites, one graph hop from
  the head of the ranking, funded out of the tail so it costs no extra bytes.
- **Literal clusters** — when a distinctive literal appears in more sites than the ranking has room to
  print, the un-printed ones are listed compactly instead of silently dropped.
- **Working-tree listing the way git does it, with every read bounded** (`internal/filedigest`,
  `gitutil` streaming digests) — including not warning about oversized files that were never
  dependents candidates.
- **Docs: the 54.9% / 57.7% savings claims are WITHDRAWN.** They were measured with a frugality clamp
  on the graph arm against a baseline given no working-policy instructions at all, and that same
  configuration LOST on accuracy (131/300 vs 150/300, McNemar p=0.013) while the README claimed
  parity. An adversarial harness audit then found eight further defects, every one distorting the
  comparison rather than the tool.

## The number that replaces them

50 language-stratified instances, 3 runs per arm, matched prompt discipline, billing-truth tokens,
Haiku: **31.6% fewer tokens** than a no-tool agent (CI [20.8, 41.3]), resolving 22.3 vs 20.0, in 23.9
turns against 30.2. Against codebase-memory-mcp: 35.7% fewer. cmm does not measurably beat no-tool.

**The saving is model-dependent and reverses on stronger models**, because it comes from deleting
grep-and-read turns a capable agent never spends: Haiku baseline 30.2 turns -> -31.6%; Sonnet baseline
15.6 -> +13.9% (n.s.); Fable baseline 10.1 -> +31.4% worse. Break-even is near a 20-turn baseline.
That is stated plainly in the README and AGENTS.md rather than buried.

Verified: `gofmt` clean, `go vet` clean, full suite green.

---
Entire trail: https://entire.io/gh/entireio/entire-graph/trails/22


suhaanthayyil:

PR #67, opened 2026-07-28
fix(sem): key the committed-tree caches on the resolved file cap, and the records cache on repo identity

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/23

Stacked on #66.

Two silent-wrong-answer holes, found by asking what happens to the graph cache across branches with
large drift. **Branch drift itself turned out to be safe** — the tree hash is in both keys, so no branch
can ever be served another's graph. Two other graph-shaping inputs were unkeyed.

### 1. The file cap is not keyed

`options.MaxFiles` (and `ENTIRE_GRAPH_MAX_FILES` behind it) shapes the graph: a capped build produces a
snapshot missing everything past the cap. Measured on this repo — a cap-5 build wrote **28 symbols**,
and the next **uncapped** search was served those 28 instead of rebuilding to **5740**. 99.5% of the
graph silently absent, and one capped ingest poisons every later query on that tree.

### 2. The records cache ignores repo identity

Repo identity prefixes every symbol ID the cache stores. Reproduced by re-pointing a remote: the warm
run still reported `gh/entireio/entire-graph` after the checkout had become a fork.
`searchSnapshotKey` already folded identity in — **upstream reached that independently, at
`search-snapshot-v6`** — but `providerRecordsKey` did not, and it is the cache that is **on by
default**.

### Change

- `searchSnapshotKey`: fold in the resolved cap; `search-snapshot-v6` -> `v7`.
- `providerRecordsKey`: fold in **both** identity and the cap; `provider-records-v1` -> `v2`.
- `Load/StoreProviderRecords` take a `context.Context` so the key can use `repoKey(ctx, repo)` the way
  the search cache does. Two production callers, both `internal/cli/root.go`.

Only ever turns hits into misses — no stale-serve path is introduced.

**Mutation-checked:** removing the key terms makes the new tests fail with
`records stored under gh/ownerone/probe would be served to gh/ownertwo/probe` and
`capped and uncapped search snapshots share a key`; restoring them goes green. Full suite green,
`gofmt`/`vet` clean.

Explicitly NOT done (recorded in the ADR): cross-worktree sharing (blocked by `repoRoot` being
overwritten from the cached header) and per-file content-hash incrementality, for which a total key is
the prerequisite.

---
Entire trail: https://entire.io/gh/entireio/entire-graph/trails/23

### DIFF
--- a/docs/adr/0002-committed-tree-cache-key.md
+++ b/docs/adr/0002-committed-tree-cache-key.md
@@ -1,0 +1,176 @@
+# ADR 0002 — Committed-tree cache key: total over graph-shaping inputs
+
+Status: Accepted
+Date: 2026-07-28
+
+## Context
+
+`entire-graph` has two committed-tree caches of the same shape:
+
+| cache | key | consumers | on by default? |
+|---|---|---|---|
+| search snapshot (`internal/sem/search_cache.go`, `searchSnapshotKey`) | `<cacheDir>/search/…/<sha256>.json.gz` | `search`, `neighbors`, `impact`, `index` | no — `searchFlags{Worktree: true}`, so only `--head` |
+| provider records (`internal/sem/records_cache.go`, `providerRecordsKey`) | `<cacheDir>/records/…/<sha256>.json.gz` | whole-graph `snapshot`, `symbols`, `edges` | yes — those commands default to committed-tree |
+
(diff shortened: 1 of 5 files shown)


suhaanthayyil:

PR #68, opened 2026-07-28
fix(sem): search returned zero results in every git worktree; raise top-k to 20; stop recorded fixtures outranking source

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/24

Stacked on #66.

### The worktree bug is a release blocker

In a git worktree `.git` is a **file** (a gitdir pointer), not a directory. Reading
`<repo>/.git/info/exclude` therefore fails, and the failure aborts the whole search:

```
read ignore file ".../.git/info/exclude": ... not a directory
```

**Live repro on `main`: zero results, every query, in any worktree.** The fixed binary returns hits on
the identical query in the identical worktree. Agents and harnesses that isolate work in worktrees get
a tool that silently answers nothing.

### top-k 10 -> 20

Breadth and depth are separate knobs. The old 20/40 default was expensive because of its **40-line
snippets**, not its 20 results; lowering top-k to 10 alongside the snippet fix cost recall for no real
byte saving.

Re-measured on 53 SWE-bench Multilingual instances (21 repos, 9 languages) **disjoint** from the suite
the agent configs were tuned on, using the query style real agents actually send (155 live search
calls: mean 43 chars):

| | top-k 10 | top-k 20 | delta |
|---|---|---|---|
| distilled queries | 69.2% | 75.0% | **+5.8 pt** (+1.9% bytes) |

Three instances flip, all toward HIT; none regress. top-k 30 adds nothing over 20.

Not fixed by breadth: Rust 33%, JavaScript 50% under distilled queries — those are ranking and
type-resolution misses, not truncation.

### Fixture-class prior

Snapshot / golden / recorded-expectation trees were outranking the source that produces them. They are
classified now, so a recorded expectation stops presenting as a fix site.

### Also here

- The literal-cluster test's breadth is **pinned** rather than inherited: its fixture writes 15 naming
  files, so at a default of 20 every site prints, nothing is left over, and an empty cluster is the
  correct answer. The test now creates the condition it is checking instead of depending on a tuning
  constant.
- Docs: widening the preselection pool **reduces** recall (measured, disjoint set: 96 -> 76.9%,
  256 -> 73.1%, 1024 -> 69.2%). Recorded so nobody tries it again hoping for recall.
- Agent guide: rule 2a, **two searches maximum, then switch tools**. Measured — runaway sessions ran
  8.4 searches against 2.7 for normal ones (worst case 23 near-identical rephrasings) and cost 2.2x
  what a no-tool agent spent on the same task. Search is how you start, not how you recover.

Full suite green, `gofmt`/`vet` clean.

---
Entire trail: https://entire.io/gh/entireio/entire-graph/trails/24

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -462,6 +462,24 @@ grep-and-read turns that a capable agent never spends: Haiku baseline 30.2 turns
 −31.6%; Sonnet baseline 15.6, graph ADDS 0.6, +13.9% (not significant); Fable baseline 10.1, graph
 ADDS 0.9, +31.4% (CI excludes zero). Break-even is near a 20-turn baseline.
 
+**Correction (measured later, 3 runs on 50 language-stratified instances, Haiku).** The block
+above says "do NOT re-search or grep to 'confirm'". That is right for the common case and wrong for
+the tail, and the tail is expensive. Of 135 instance-run pairs, **8 (6%)** were sessions where the
+no-tool baseline finished comfortably and the graph-assisted agent hit the 50-turn cap at **2.2x the
+baseline's tokens**. Those 8 alone cost **4.8 points** of the headline token saving (-33.5% with
+them, -38.3% without). Cause, from the wrapper call logs: they ran a mean of **8.4 searches** against
+2.7 for normal sessions (worst case **23**), each a near-identical rephrasing — e.g. "zero padding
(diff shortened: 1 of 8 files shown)


suhaanthayyil:

PR #69, opened 2026-07-28
fix(sem): honor nested .gitignore in the vendored-directory heuristic; feat(cli): default cache dir + GRAPH_REPORT.md

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/25

Stacked on #66. The low-hanging items from the graphify comparison — the ones that belong in the graph
rather than in the Brain.

- **Nested `.gitignore` is honored in the vendored-directory heuristic.** It previously consulted only
  the root ignore file, so a dependency tree ignored by a nested `.gitignore` was still walked and
  indexed as if it were first-party source.
- **A default cache directory.** The committed-tree index previously cached only when
  `ENTIRE_PLUGIN_DATA_DIR` happened to be set, so the common invocation silently re-parsed every time.
  It now uses the platform per-user cache dir unless `--cache-dir` / `ENTIRE_PLUGIN_DATA_DIR` redirect
  it. The working tree is still never cached.
- **`index --report GRAPH_REPORT.md`** — a human-readable summary of what got indexed (languages, file
  and symbol counts, partial failures), so "did it actually parse my repo" is answerable without
  piping NDJSON through jq.
- **`docs/BRAIN-VS-GRAPH.md`** — where the boundary is: the graph is deterministic, local, per-commit
  structure; the Brain is durable cross-session knowledge built on top. Written so the next feature
  request lands on the right side of the line.

Full suite green, `gofmt`/`vet` clean.

---
Entire trail: https://entire.io/gh/entireio/entire-graph/trails/25

### DIFF
--- a/docs/BRAIN-VS-GRAPH.md
+++ b/docs/BRAIN-VS-GRAPH.md
@@ -1,0 +1,69 @@
+# Brain vs. graph — what entire-graph deliberately does not do
+
+entire-graph is a **provider**: it turns a git tree into a deterministic, typed, no-egress
+snapshot and answers ranked queries over it. Entire Brain is the **consumer**: it ingests that
+snapshot alongside durable facts, sessions and checkpoints, and is where knowledge that is not
+derivable from one tree lives.
+
+The line is not organizational, it is a property of the artifact. Anything that is a pure function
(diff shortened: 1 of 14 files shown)


suhaanthayyil:

PR #70, opened 2026-07-30
search: separate the complete-body head from the locator head, and pad rank 1

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/26

Two new opt-in `SearchOptions` (`0` keeps today's behaviour byte-for-byte, so no shipped payload changes unless a caller asks):

| option | flag | effect |
|---|---|---|
| `BodyHeadRanks` | `--body-head-ranks` | caps how deep the **complete-body** upgrade reaches, independently of the locator head |
| `EnclosureContextLines` | `--enclosure-context-lines` | pads the **rank-1** body with N source lines each side |

## Why — measured, not chosen by taste

**Rank utility** (sonnet, 20 SWE-bench Multilingual instances, matched prompts across arms). Of the files a payload showed, how often the agent later touched them:

| rank | files shown | later Read | later **Edited** |
|---|---|---|---|
| 1 | 24 | **62%** | **54%** |
| 2 | 15 | 27% | 20% |
| 3 | 9 | **0%** | 11% |
| 4 | 15 | 7% | 7% |
| 5 | 13 | **0%** | **0%** |

`searchEnclosureHeadRanks = 5` buys **21 complete bodies at ranks 3–5 that are read 0–7% of the time**, and those bytes are replayed into the model on every later turn.

**Why the two heads must move separately.** The gold file sits at **rank 6–8 on 17% of instances** (5/30, disjoint set), so shortening the *ranking* loses fix sites — cutting `--top-k` to 5 was measured and rejected for exactly that. A tail locator is the only mention of its file and costs ~70 B; a body costs ~1 kB. **Bodies are the expense; locators are the insurance.** `planSearchEnclosures` now also skips ranks the allocator could never upgrade, removing the file read it used to do for them.

**Why pad rank 1.** Over 65 sessions, of the reads that re-open a file whose complete body the payload *already printed*, only **3.8%** ask for lines inside that body — **80% ask for lines outside it** (file head 15.6%, within 40 lines 41.4%), median gap **42 lines**, median requested window **50 lines**. The body is the right unit to *edit* and the wrong unit to *understand*, so the unpadded body provokes the very read it was meant to remove.

The margin is rank-1 only, is capped by `maxLines` like any other body (a padded body can never exceed what an unpadded one was allowed), clamps at file edges, and never renames the symbol.

## Screen — 30 disjoint instances, no model calls, $0

| config | mean bytes | vs as-run | bodies | gold recall | gold has body | top-1 body |
|---|---|---|---|---|---|---|
| as-run (head 5) | 4,310 | — | 3.10 | 22/30 | 14/30 | 25/30 |
| `--body-head-ranks 2` | 2,856 | **−33.7%** | 1.43 | 22/30 | 13/30 | 25/30 |
| ` + --enclosure-context-lines 40` | 4,414 | +2.4% | 1.43 | 22/30 | 13/30 | 25/30 |
| `--body-head-ranks 1 ... 40` | 3,815 | −11.5% | 0.77 | 22/30 | **7/30** | 23/30 |

Dropping the bodies at ranks 3–5 **pays for a 40-line margin on rank 1 almost exactly**: same gold recall, same gold-body coverage, same top-1 coverage, byte-neutral. `--body-head-ranks 1` is **not viable** — it loses 6 gold bodies including at ranks 1 and 2.

Gold recall is identical across every config by construction: results are never dropped, only bodies are.

## Scope and status

This is the mechanism, not a session-level claim. It has **not** yet been run end-to-end against a baseline — the intended next step is a 30-instance sonnet cell, where the falsifier is explicit: if `Read` calls with `offset ≤ 20` and reads adjacent to the rank-1 body do not fall, the margin was not what those reads wanted and should be reverted.

Defaults are unchanged, so nothing here alters current behaviour on its own.

`gofmt`, `go vet ./internal/...` and `go test ./...` are clean. Tests pin: narrowing the body head must not drop results; `0` means the built-in depth; the margin applies to rank 1 only; it is refused rather than truncated when it would exceed the line cap; it clamps at file edges.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,7 +47,7 @@ material. Symbols too large to return whole (>160 lines) keep their focused wind
 **Three blocks each replace a tool call you would otherwise spend a turn on.** They are the default
 payload's reason to exist, and they are instructions rather than background reading.
 
-- **SAME-CONCEPT LITERALS** (`literal_cluster` in JSON, `stats.literal_cluster_bytes`) is `grep`
+- **SAME-CONCEPT LITERAL** (`literal_cluster` in JSON, `stats.literal_cluster_bytes`) is `grep`
   folded into `search`. When the top hit contains one distinctive literal that names the queried
   concept — an enum constant's value, an option string, a compound identifier — the block lists every
   occurrence of it in the repository, each annotated with its enclosing symbol and one of three
@@ -176,7 +176,7 @@ prove it, sweep the concept, check the contract, check the names, check the neig
 
 ```text
 LOW CONFIDENCE  ->  CLOSED SET  ->  [CONTAINER MAP]  ->  candidate fix sites  ->  COVERING TEST
-                ->  VERIFY  ->  SAME-CONCEPT LITERALS  ->  [TYPES IN THIS SIGNATURE]
+                ->  VERIFY  ->  SAME-CONCEPT LITERAL  ->  [TYPES IN THIS SIGNATURE]
                 ->  [DECLARATIONS]  ->  RELATED SITES  ->  DOCS & FIXTURES
 ```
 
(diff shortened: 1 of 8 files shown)


suhaanthayyil:

PR #79, opened 2026-08-05
fix(sem): extract TypeScript bodyless declarations (overloads, ambients)

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/33

## The bug

`entityFromNode` gated both `function_signature` and `method_signature` to Dart, with the comment that TypeScript uses those node types for interface members. That is only half true.

tree-sitter-typescript emits `function_signature` for **every bodyless function declaration** — overload signatures, `declare function`, ambients, `.d.ts` — and `method_signature` for class-body overload declarations as well as interface members. All of them were silently dropped from the graph.

Concretely, vue's `renderList.ts` has five overload signatures plus one implementation. It exposed **1 symbol**, covering only the implementation's lines. Lines 10–53 — the entire typed surface — were unindexed, so no query phrased in terms of the declared parameter types could retrieve the file. After the fix: **6 symbols**.

## The fix

Additive, `+54` lines, no deletions, narrowly gated:

- **`function_signature`** in TypeScript becomes a `function` (or a `method` when scoped). An overload set is where the parameter types and generics live; the implementation's own signature is usually the erased `(source: any, ...)` catch-all.
- **`method_signature`** becomes a `method` **only inside a `class_body`**, via the new `typeScriptClassBodySignature` helper.

Interface and type-literal members deliberately stay inventory-only. They declare a contract, not a definition, and emitting them would make every `x.m()` call name-resolvable to a bodyless interface member. That is not hypothetical — extending this case to `interface_body`/`object_type` members fabricates a CALLS edge in `TestTypeScriptNamespaceCallSkipsParameterReceiverRoots`, where a parameter named `B` shadows a namespace and `B.parse()` then binds to `Client.parse` instead of correctly resolving to nothing.

Dart behaviour is unchanged — the existing Dart paths are reached only when the language is not TypeScript.

## Verification

- Full `internal/sem` suite green (**1976 tests**).
- New regression test `parser_typescript_regression_test.go` (+124 lines) covering overload sets, `declare function`, class-body overloads, and the interface-member exclusion.
- **Attribution check:** symbol output is byte-identical to the pre-fix binary on eight non-TypeScript repositories — django, axum, gson, carbon, bat, caddy, fmt, rubocop. So the change is attributable to TypeScript alone and cannot perturb other languages.

## Scope note

Repo-wide symbol gain is modest (vuejs +1.0%, docusaurus +2.1%) — this is a correctness fix for a whole class of declarations, not a broad indexing win. Its practical effect is concentrated in TypeScript codebases that use overloads and ambient declarations heavily.

### DIFF
--- a/internal/cli/env.go
+++ b/internal/cli/env.go
@@ -13,6 +13,15 @@ const (
 	// interactive user does not have to add flags to every call. Its value is a comma-separated list
 	// of `container-map`, `signature-types`, `type-card`, or `all`. See searchReferenceBlocks.
 	envReferenceBlocks = "ENTIRE_GRAPH_REFERENCE_BLOCKS"
+	// envPresearch names a file holding the payload for this session, computed BEFORE the agent
+	// started. When it is set, `search` echoes those bytes instead of querying — see
+	// echoPresearchPayload for the measurement that motivates it.
+	//
+	// envPresearchAlias is the same knob under the prefix the benchmark harness already uses for
+	// every one of its search knobs (EG_TOPK, EG_DEEP, EG_MAXBYTES, EG_PROFILE), so the harness can
+	// set it beside them; the ENTIRE_GRAPH_ name is the one this repo documents, and it wins.
+	envPresearch      = "ENTIRE_GRAPH_PRESEARCH"
+	envPresearchAlias = "EG_PRESEARCH"
 )
 
 // cacheDirName is this provider's directory inside the platform's per-user cache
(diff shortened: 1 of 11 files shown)


suhaanthayyil:
Closing as superseded by #82, with a caveat worth recording.

**Superseded:** #82 contains this PR's four search commits by identical SHA (8d464ce, 5511d6f, afdb54e, afd5a1c) and folds the TypeScript fix into 19b0769. `git diff pr79 pr82 -- internal/sem/parser_typescript_regression_test.go` is empty and the parser.go hunks are byte-identical; #82's parser.go is a strict superset (it also carries the Go `method_elem`/`method_spec` case).

**Two reasons not to merge this one on its own**, both measured on a 2-file TS repro:

1. It carries the unhardened `EG_PRESEARCH` path — `internal/cli/search.go:254-261` reads a pre-delivered payload with no length check. `EG_PRESEARCH=<zero-byte file> entire-graph search --query x` returns **0 bytes on stdout, exit 0**. #82 fixes this (8a4ffb3, search.go:310-312).
2. Merging it first makes #84 conflict in `internal/sem/search.go` (the `mergeSameFileSearchSpans` call lands in the same `SearchRepository` stage #84 edits), adding a third source of the #82↔#84 conflict while contributing nothing #82 lacks.

The TypeScript regression tests here are genuine and verified fail-without/pass-with — they live on in #82.

Separately: the same TS change causes compound-v1 ID churn and a fast-profile CALLS loss in **both** this PR and #82. Being fixed on #82 before it merges.

PR #80, opened 2026-08-06
fix(sem,cli): TypeScript bodyless declarations, test-file demotion, bare-locator windows

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/34

Three independent retrieval/extraction fixes. Each has a regression test **proven to fail without the change and pass with it** (production file stashed, test run, restored, run again — both outcomes captured). All measured on rep30 / SWE-bench Multilingual with a **separate `--cache-dir` per binary**, because the index cache is tree-keyed rather than binary-keyed and a shared cache silently returns the first build's results.

## 1. TypeScript bodyless declarations — `parser.go` (+54)

`entityFromNode` gated `function_signature` and `method_signature` to Dart, on the grounds that TypeScript uses those nodes for interface members. That is only half true. tree-sitter-typescript emits `function_signature` for **every bodyless function declaration** — overload signatures, `declare function`, ambients, `.d.ts` — and `method_signature` for class-body overloads as well as interface members. All were silently dropped.

vue's `renderList.ts` (five overloads + one implementation) exposed **1 symbol**, covering only the implementation's lines. Lines 10–53 — the entire typed surface — were unindexed, so no query phrased in terms of the declared parameter types could retrieve it. After: **6 symbols**.

Narrowly gated. `method_signature` becomes a symbol **only inside a `class_body`** (new `typeScriptClassBodySignature`). Interface and type-literal members stay inventory-only on purpose: they declare a contract, not a definition, and emitting them makes every `x.m()` call name-resolvable to a bodyless interface member. That is not hypothetical — extending this to `interface_body`/`object_type` members fabricates a CALLS edge in `TestTypeScriptNamespaceCallSkipsParameterReceiverRoots`, where a parameter named `B` shadows a namespace and `B.parse()` then binds to `Client.parse` instead of correctly resolving to nothing.

**Attribution check:** symbol output is **byte-identical** to the pre-fix binary on eight non-TypeScript repositories (django, axum, gson, carbon, bat, caddy, fmt, rubocop), so the change cannot perturb other languages.

## 2. Test-file demotion — `search_testrank.go`, `search_confidence.go`

3 of the 10 most expensive rep30 instances had a **test file as the rank-1 hit** (bat → `tests/integration_tests.rs`, hugo-12204 → `render_hook_integration_test.go`, fmt → `test/ranges-test.cc`). When the payload points at a test instead of the code, the injected location is wrong *and* the derived VERIFY command scopes to the wrong place.

Demotion, not exclusion — tests stay retrievable, they just stop taking rank 1.

Measured n=30: rank-1-is-a-test **3 → 1**, gold rank **unchanged 13/30**, VERIFY commands changed **0/30**, file lists moved on 2/30. Also carries an independent real bug fix: the confidence gap was computed **signed** where it should be absolute.

## 3. Bare-locator windows — `search_locator_window.go`, `cli/search.go`

A hit with no resolvable symbol shipped as a naked `path:line` — no body, no name — so the agent had to spend a turn opening the file just to learn what it was. **19 of 30 payloads contained at least one; after, 1.**

Text-renderer only: ranking is provably untouched, top-5 **byte-identical on 30/30**. Cost is +1.83% payload bytes.

## Verification

- `go test ./internal/sem/ ./internal/cli/` — green (sem 52.4s, cli 4.8s)
- Each regression test verified in **both directions**
- Collateral check on non-JS/TS instances: top-5 unchanged

## Deliberately not included

- **Same-basename path prior** — held. It buys one gold@5 but **downgrades a correctly-scoped VERIFY** on preact from a narrow `npx mocha test/browser/render.test.js` to a whole-suite `npm test`, taking the correctly-scoped subset 7 → 6. Bad trade on a benchmark where VERIFY scope is the largest token lever. Worth revisiting once the `coverage_note` loss is understood.
- **Coverage-mirror route** — dropped on its own author's no-ship: VERIFY coverage 17/30 → 17/30, and 2 of its 3 new covering-test entries named the *wrong* test.

## No token claim

Three of these are payload-quality changes, and **ten such levers have already failed as causal token levers** on this benchmark. These are correctness fixes and are measured on retrieval metrics only. Any savings claim would need its own paired cell.

Supersedes #79 (TypeScript fix alone, based on an older commit).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -197,3 +197,11 @@ mise run check   # fmt + vet + race tests + build
 ```
 
 Contract rules that must not break: schema `1.x` is frozen and additive-only (`docs/adr/0001-ga-schema-contract.md`); the provider is **no-egress** (never add remote fetches, hosted API calls, telemetry, or runtime grammar downloads); `compound-v1` symbol IDs must stay stable across ordinary edits; unsupported/unparseable files must surface as machine-readable partial failures, never silent drops. All logic lives under `internal/` (`sem` = parsing/graph/search, `cli` = hand-rolled dispatch, `gitutil` = git subprocess); `cmd/entire-graph/main.go` is a thin entry point. The plugin manifest (`entire-plugin.yml`) registers the subcommand `graph`, so users type `entire graph ...`. This project was **previously named `entire-sem`** — do not reintroduce the old name. **Entire Brain** (`entire-brain`) is the separate downstream consumer of this provider's NDJSON — not an old name for this project.
+
+<!-- entire-graph:begin -->
+This repo has the entire-graph code graph installed. Before exploring code with
+grep/find/whole-file reads, read .entire/graph-agent.md — the search-first, verify-once
+doctrine for coding agents: search instead of grepping, then check the sibling sites and
+compile (or run the nearest existing test) once before you finish.
+@.entire/graph-agent.md
+<!-- entire-graph:end -->
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1 +1,9 @@
 @AGENTS.md
+
+<!-- entire-graph:begin -->
+This repo has the entire-graph code graph installed. Before exploring code with
+grep/find/whole-file reads, read .entire/graph-agent.md — the search-first, verify-once
+doctrine for coding agents: search instead of grepping, then check the sibling sites and
+compile (or run the nearest existing test) once before you finish.
+@.entire/graph-agent.md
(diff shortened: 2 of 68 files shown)


suhaanthayyil:
Closing: self-declared superseded by #82, and unmergeable.

Forked at 1d7e3eb, before #75/#76/#78 landed — currently CONFLICTING/DIRTY and 30 commits behind main. #82 is this branch's 22 commits merged onto current main plus the three silent-failure fixes. Nothing here is lost.

PR #100, opened 2026-08-14
feat(search): multi-resolution prose retrieval

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/51

## What

Ranking returns **one region per file**. That is correct for code, where a file is the unit
you want to open. It is wrong for prose, where a single markdown document can hold the answer
across several distant regions.

The extra regions **already exist**. `search_prose_parent.go` plans them and
`search_prose_passage.go` attaches them to their parent as `Passages` — but a consumer reading
`results[]` never sees them. Measured on a real prose corpus, search returned 19 results across
19 files while carrying **195 passages holding 42 KB of evidence that no `results[]` consumer
could reach**; ranks 6–19 got a one-line snippet.

When fewer distinct files match than `--top-k` asked for, this change spends the spare slots
promoting those regions to results of their own.

## Why it cannot regress code search

- Strictly additive: runs **after** selection, only fills slots the distinct-file ranking left
  empty (`len(results) >= topK` → no-op), so it can never displace a file.
- Never breaches `--max-context-bytes`.
- No code result carries passages, so code search has nothing to promote.

Verified empirically: 16 real code queries against this repo at `--top-k 10` **and** `--top-k 50`,
full JSON compared — **0 mismatches**.

`--single-resolution` restores the previous output.

## Verification

Measured on LoCoMo (n=1540 per arm), paired and same-window, **both arms on the identical
corpus with only the binary differing** — answerer and judge both `gpt-5.6-sol`, `top_k=200`.

| | previous | this change | Δ | discordant | p |
|---|---|---|---|---|---|
| all pairs (n=634 so far) | 91.91 | 94.06 | **+2.16pp** | 16 / 4 | **0.012** |

The estimate has held in the **+2.2 to +2.5pp** band across ten reads from n=118 upward. The
full n=1540 run is still completing; this PR should not merge until it lands and passes the
gates.

Mechanism check — the change fires and does what it claims:

```
shipped binary : 17.60 units/q across 17.60 distinct files  -> ratio 1.00
this change    : 40.22 units/q across 17.60 distinct files  -> ratio 2.29  (same files)
```

The control arm was verified **byte-identical to the shipped binary** across 45 queries × 3
corpora (full JSON, latency/cache stats stripped, 0 mismatches), so the measured delta is
attributable to this change alone.

It is also not a context dump: 30–40% of corpus bytes returned, against the previous 22–29%.

## Fairness

This is a **post-ranking assembly rule that never reads the query**. It cannot key off any
benchmark, because it never sees a question. No question-conditioned branching, no vocabulary
lists, no category logic.

## Tests

`mise run check` passes — fmt, vet, `go test -race ./...` across all packages, build.
6 new tests in `internal/sem/search_multi_resolution_test.go`.

### DIFF
--- a/internal/cli/help.go
+++ b/internal/cli/help.go
@@ -127,7 +127,8 @@ var commandDocs = []commandDoc{
 		usage:   []string{`entire graph search --query "issue or concept" --repo . [--top-k 10] [--format text|json|ndjson|agent] [--head] [--profile fast|full] [--deep]`},
 		long: "Ranked source regions for a plain-language description, with source and file:line inline, budgeted to drop straight into context. This is the first move for almost every locate task.\n\n" +
 			"By default search returns: ranked candidate fix sites (top hits as full function bodies), RELATED SITES, the COVERING TEST plus other tests over the same code (ALSO COVERING), SAME-CONCEPT LITERAL (every place the concept is named, tagged EDIT/CONSUMER/DOC), a VERIFY line (the narrowest test command for the file), and a CLOSED-SET WARNING when a switch over a sealed set would fail at runtime. The three reference blocks (container map, signature types, declaration card) are OFF by default because they cost turns in agent sessions; --reference-blocks all turns them on for interactive reading.\n\n" +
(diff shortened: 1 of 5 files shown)


suhaanthayyil:

PR #101, opened 2026-08-14
bench: LoCoMo memory-system comparison results and reproduction kit

suhaanthayyil:
## What this is

`bench/memory/` — the LoCoMo memory-system comparison, **self-contained**: what the results were, and how anyone outside can reproduce them. The harness previously existed only on a single VM, uncommitted, so the published table was not reproducible by a third party. This closes that.

Additive only — no existing entire-graph source is modified. Docs, Python adapters, and patch files; no Go touched.

- **[`LOCOMO-COMPARISON.md`](bench/memory/LOCOMO-COMPARISON.md)** — the results.
- **[`README.md`](bench/memory/README.md)** — the reproduction kit.
- **[`RUN-INDEX.md`](bench/memory/RUN-INDEX.md)** — every run, traceable to an artifact.

## The claim

> **entire-graph matches the leading open-source memory system on accuracy, while building its index 3x faster and using zero LLM calls to do it.**

Deliberately **not** "first on accuracy". Measured side by side against mem0 OSS the two are a statistical dead heat, and the document says so. The pre-registered decision rule (significance below p=0.05, non-significant lead to 0.10, tie at 0.10+) is followed exactly, and the same-window control is still running behind a clearly-marked placeholder that no claim rests on.

All rows: n=1540, no subsetting, answerer **and** judge both `gpt-5.6-sol`, `top_k=200`, zero drops and zero empty-context retrievals.

| # | system | run id | LoCoMo | ingest LLM calls |
|---|---|---|---|---|
| **1** | **entire-graph** (shipped + PR #100) | `mrq_mres` | **93.83** | **0** |
| 2 | mem0 OSS | `plan_g_mem0` | 93.77 | 1+ per memory |
| 3 | cognee | `field_cognee_loco2` | 92.86 | 1+ per memory |
| 4 | entire-graph (shipped default) | `mrq_base` | 91.56 | **0** |
| 5 | cmm (patched, Markdown-Section) | `full_cmm` | 91.30 | **0** |
| 6 | graphify | `full_graphify` | 87.34 | **0** |
| 7 | letta | `field_letta_loco` | 80.58 | 1+ per memory |
| 8 | supermemory | `field_sm_loco` | 77.60 | hosted |
| — | graphiti | — | — | 4,463 s/conv, 3 of 10 done, ~160h left |
| — | mem0 Pro / managed | — | — | hosted; answerer/judge cannot be pinned |

## The finding it anchors on — a bug in our own ranker

entire-graph returns **one region per file**. Correct for code; wrong for prose, where a file is a whole conversation session.

LoCoMo conversation 0 indexes as 19 session documents, so entire-graph could return at most 19 things at any `top_k`. Measured across conv 0's 152 questions:

| system | items returned / question, conv 0 |
|---|---|
| **entire-graph, before PR #100** | **mean 18.9, min 7, max 19** |
| mem0 | **200.0** — full, every question |
| cmm | 198.9 |
| graphify | 84.4 |
| **entire-graph, with PR #100** | **198.7** (full dataset mean) |

The budget was 200; we were using nine percent of it, structurally.

**The corpus does not force this** — cmm and graphify get the identical 19-file corpus and return 419 and 423 distinct units across conv 0. **And the regions already existed** — the same pre-PR-#100 run emitted **377 distinct regions** across conv 0 while never returning more than 19 on any single question. The ranker was computing at a far finer grain than it would emit, then discarding the rest unread. PR #100 spends the unused slots on exactly those regions: no ingest change, no re-indexing, no migration, no new model call.

**Paired, same window, binary the only variable: 91.56 → 93.83, +2.27pp, discordant 46–11, exact McNemar p = 3.3e-06, n = 1540.** Gains concentrate in multi-hop (+3.54) and temporal (+3.43) — the categories that need several turns from one session, which is exactly what a one-region-per-file ranker cannot deliver. Mechanism and result agree.

## Where we win, and where we lose

**Win, outright:** ingest 17m 46s vs mem0's 55m 25s and cognee's 14h 17m; **zero** LLM calls at ingest vs 1+ per memory — a cost that repeats on every update forever. Structural, no p-value needed.

**Loss, stated plainly and not buried:** search latency is **fourth** at 868 ms, against cmm 22 ms, graphify 194 ms, mem0 535 ms. The previous default was 252 ms, so multi-resolution cost ~600 ms per query to buy 2.27 accuracy points.

## Integrity

- **Three competitor bugs we found and fixed in the competitors' favour** — mem0 +6.04pp (`limit` swallowed by `**kwargs`, every search returning 20 not 200), cognee +13.77pp (empty-buffer defect), cmm from structurally zero to 91.30 (shipped v0.9.0 excludes `Section` nodes from its own BM25 results). Two of these took arms entire-graph was beating and put them level with or ahead of it. Published anyway.
- **P(correct | gold evidence retrieved) is 0.96–0.97 for every system.** Answering is solved and undifferentiated here; the arms differ only in retrieval.
- **Loss forensics:** 18/32 retrieval miss, 9/32 synthesis, 3 rank burial, 2 bad gold, **0 judge errors**. Perfect session recall (32/32 losses, 46/46 wins), turn-level miss.
- **Recall for mem0 and letta is dashed, not estimated** — they store LLM-rewritten facts with zero verbatim overlap, so substring matching scores them at exactly 0.000, an artifact rather than a measurement. No proxy substituted.
- **A fix that was built, measured, and rejected** is reported, including the earlier version of it whose apparent gain came from bounding bytes on the JSON envelope rather than on the text a consumer reads.
- **Oracle runs appear nowhere as results.** `EG_FULL_CONTEXT=1` runs carry `fair_mode: false` and are recorded in `RUN-INDEX.md` as never-publishable so their numbers cannot later be mistaken for scores.
- **Window discipline:** a measured −2.21pt drift on an identical config sets the rule — cross-window gaps under ~2pp are not orderable, same-window comparisons are. `RUN-INDEX.md` marks which is which, including that entire-graph's 94.68 at `turn+session` granularity must always be quoted beside the 92.14 its default scored in that same window.

## Verified, not transcribed

Every number in the results document was re-derived from the run artifacts: `mrq_mres` 93.8312 and `mrq_base` 91.5584 at n=1540 with `drops=0`/`zero-context=0`; the paired all-pairs statistic; the per-category deltas; and the item-count and context-size figures behind the mechanism.

**One discrepancy found and disclosed in the document.** The retrieval-latency artifact on the benchmark host (`timings/timings_entire_locomo.json`) is stamped **`valid_for_publication: 0`** — collected while another arm was ingesting on the same box — and its competitor figures disagree with the reported table (cognee 4,987 ms there vs 7,585 ms reported; supermemory 417 ms vs 8,014 ms). §5 therefore marks the latency ranking **indicative, pending an idle-host re-measurement**, while still stating plainly that entire-graph is not first and that PR #100 increased latency substantially. Accuracy and ingest results are unaffected — they come from run artifacts, not the timing files.

## Reproduction kit

Upstream is `mem0ai/memory-benchmarks` @ **`4b61c5d31b9c668a12b4f5e78064248a02c82d2b`** (Apache-2.0). The upstream harness is **not** vendored. `UPSTREAM.md` carries a file-level md5 manifest byte-verifying which files are unmodified upstream (`metrics.py`, `schema.py`, `utils.py`, `locomo/prompts.py`, both `__init__.py`) and which we changed; the changed ones ship as patches, and all four `git apply --check -p1` cleanly against a fresh checkout of that commit. That `locomo/prompts.py` is byte-identical is load-bearing: the answerer and judge prompts are upstream's, untouched, identical for every arm.

Vendored: our four adapters (`entire`, `cmm`, `graphify`, `graphify_mem_bridge`), the `runmeta` provenance capture and `FAIR_MODE` guard, the five patches, the launcher, and the fairness spec and results docs.

## Secrets

None. Credentials are referenced by env-var **name** only. Verified with `gitleaks` across all commits on the branch (1 hit, `patches/0003:175` — the code expression `api_key=args.mem0_api_key if backend == "cloud" else None`, no literal), plus targeted scans for key shapes, assignment-shaped secrets, Azure endpoint hostnames, and high-entropy strings. Internal scaffolding — agent ids, a DRAFT banner, an internal host name, first-person asides — was stripped from the vendored docs before publication; technical content is unchanged.

## Checks

No Go changed, so `mise run check` is not implicated. Ran `python3 -m py_compile` on all five vendored modules, `bash -n` on the launcher, and `git apply --check -p1` for patches 0001–0004 against a fresh upstream checkout at the pinned commit — all clean.

### DIFF
--- a/bench/memory/.gitignore
+++ b/bench/memory/.gitignore
@@ -1,0 +1,2 @@
+__pycache__/
+*.pyc
--- a/bench/memory/AUTO-SCORES.md
+++ b/bench/memory/AUTO-SCORES.md
@@ -1,0 +1,37 @@
+- [2026-08-13 17:56:54 UTC] **mem0_loco** (field_mem0_loco, locomo): 93.44155844155844% (1439/1540), drops=0, restarts=0, subset_applied=False, abandoned=False -- NOT YET folded into MEMORY-BENCH-DELIVERABLE.md
+- [2026-08-13 18:00:27 UTC] **mem0_loco** (field_mem0_loco, locomo): 93.44155844155844% (1439/1540), drops=0, restarts=0, subset_applied=False, abandoned=False -- NOT YET folded into MEMORY-BENCH-DELIVERABLE.md
+## full_graphify  (2026-08-14T16:26:53+00:00)
+```
+=== full_graphify  (locomo_results_20260814_162508.json)
+  ACC top_200: 8733.77%  (1345/1540)  avg_score=87.3377
+    multi-hop      8191.49%  (231/282)
+    open-domain    7395.83%  (71/96)
+    single-hop     9227.11%  (776/841)
+    temporal       8317.76%  (267/321)
+  GATES: drops=0 (0.00%)  zero_ctx=0 (0.00%)  empty_generated_answers=0 (0.00%)
+  units/q mean=98.0 median=94.5 min=1  chars/q mean=16342 min=164
+  zero-ctx by conv: none
+```
+
+=== full_graphify  (locomo_results_20260814_162508.json)
+  ACC top_200: 87.34%  (1345/1540)  avg_score=87.3377
+    multi-hop       81.91%  (231/282)
+    open-domain     73.96%  (71/96)
+    single-hop      92.27%  (776/841)
+    temporal        83.18%  (267/321)
+  GATES: drops=0 (0.00%)  zero_ctx=0 (0.00%)  empty_generated_answers=0 (0.00%)
+  units/q mean=98.0 median=94.5 min=1  chars/q mean=16342 min=164
+  zero-ctx by conv: none
+## full_cmm  (2026-08-14T16:37:00+00:00)
+```
+=== full_cmm  (locomo_results_20260814_163440.json)
+  ACC top_200: 91.30%  (1406/1540)  avg_score=91.2987
+    multi-hop       91.84%  (259/282)
+    open-domain     71.88%  (69/96)
+    single-hop      93.58%  (787/841)
+    temporal        90.65%  (291/321)
+  GATES: drops=0 (0.00%)  zero_ctx=0 (0.00%)  empty_generated_answers=0 (0.00%)
+  units/q mean=199.8 median=200.0 min=34  chars/q mean=37006 min=6752
+  zero-ctx by conv: none
+```
+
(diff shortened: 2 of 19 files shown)


suhaanthayyil:

PR #103, opened 2026-08-15
fix(search): rank prose sections as independent candidates

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/54

Stacked on #100. Base is `feat/multi-resolution-retrieval`, so the diff here is this change alone.

## The defect

`selectSearchCandidates` admits **one candidate per file**. For prose that is the wrong unit: a
markdown document's heading regions already exist in the graph as `section` entities
(`parser.go` `markdownEntities`) and are already scored — then collapsed to one per file.

#100 promotes the discarded regions back into empty top-k slots, but as **satellites**:
`proseResolutionResult` sets `Score: parent.Score` verbatim, promotions are ordered by a
round-robin over *parent index*, and every promotion is appended after every parent.

Artifact proof on a real prose corpus at `--top-k 200`:

```
200 results  ->  19 distinct scores
```

181 of 200 results are score-duplicates of one of 19 parents.

**The downstream consequence.** A consumer sorting by score with a stable sort gets 19 blocks —
one document's parent followed by all of its promoted regions, then the next. Truncate to the
top 50, 20 or 10 and you see only the first one or two documents. Any question needing evidence
stitched across documents structurally cannot get it.

## The change

The prose retrieval unit becomes the **section**, not the file (`proseParentKey`, keyed by the
heading's SymbolID, falling back to file path where a document has no section structure). A
section competes on its own score, as an independent file would.

Breadth is preserved: **no document gets a second region until every document that fits has
one** — that rule binds every selection pass, not just the first.

`--document-resolution` restores the previous behaviour.

The unit is decided by the parser's section entities and the file extension. Nothing reads the
query.

## Measured

152 questions on a real prose corpus, identical search arguments:

| | items/q | distinct regions/q | **distinct scores/q** | files/q | visible chars/q |
|---|---|---|---|---|---|
| before (#100) | 195.0 | 243.0 | **17.1** | 18.9 | 60,854 |
| **after** | **196.0** | **295.3** | **84.9** | **18.9** | **53,929** |
| `--document-resolution` | 195.0 | 243.0 | 17.1 | 18.9 | 60,854 |

Score diversity 17.1 → 84.9, document coverage unchanged at 18.9, and **11.4% fewer bytes** —
the previous build was computing regions and discarding them unread. The off switch is
byte-identical to the base branch on every metric.

## Code search is unaffected

**30 of 30 full-JSON payloads identical** — 15 real locate queries against this repo at
`--top-k 10` *and* `--top-k 50`, comparing the entire payload (results element by element plus
`stats`, `commit`, `tree`, `completeness`, `diagnostics`), latency telemetry excluded.

The prose branch sits behind `proseParentCorpus`, which requires ≥4 units and ≥80%
doc-extension paths. A source repo cannot reach it — verified rather than assumed: three
deliberately documentation-flavoured queries against this Go repo returned **zero** results
carrying `retrieval_mode=prose-parent` in either build.

## Benchmark result

LoCoMo, all 1540 questions, answerer and judge both `gpt-5.6-sol`, `top_k=200`, measured against
a mem0 OSS control **running concurrently in the same window**:

| | overall | single-hop | multi-hop | temporal | open-domain |
|---|---|---|---|---|---|
| **this branch** | **94.74** | **96.67** | **93.97** | **95.33** | 78.12 |
| mem0 OSS | 93.83 | 95.12 | 93.62 | 94.70 | **80.21** |

**+0.91pp overall, discordant 43–29, McNemar p = 0.125 — statistically tied, point estimate
favours entire-graph.** The wording follows a decision rule registered before the run finished.

**single-hop +1.55pp, p = 0.035** is the one cell that clears significance on its own.

Multi-hop moved from **−1.44 against mem0 to +0.35** — that was the deficit this change targets,
and it is the category the 19-distinct-scores collapse was starving.

No harness flag and no ingest change: this is the shipped corpus shape.

### Stated plainly

- Run-to-run noise on this harness is **0.65pp** — two identical mem0 runs scored 93.83 and 93.57.
- entire-graph delivers ~30% more context per question than mem0, and **mem0 is more
  context-efficient below ~25k chars**. Item-count parity at `top_k` is not context parity.
- Prose search is **2.1× slower per query** (0.300s → 0.628s) because passages are planned for
  ~196 units instead of 19. Code-search latency is unchanged, since the prose path never fires.

## Tests

`mise run check` green. Three regressions were found and fixed during development, each now
pinned by a test: whole-document exclusion at small top-k, loss of same-document passages at
tight top-k, and one region being returned up to seven times.

### DIFF
--- a/internal/cli/help.go
+++ b/internal/cli/help.go
@@ -128,7 +128,7 @@ var commandDocs = []commandDoc{
 		long: "Ranked source regions for a plain-language description, with source and file:line inline, budgeted to drop straight into context. This is the first move for almost every locate task.\n\n" +
 			"By default search returns: ranked candidate fix sites (top hits as full function bodies), RELATED SITES, the COVERING TEST plus other tests over the same code (ALSO COVERING), SAME-CONCEPT LITERAL (every place the concept is named, tagged EDIT/CONSUMER/DOC), a VERIFY line (the narrowest test command for the file), and a CLOSED-SET WARNING when a switch over a sealed set would fail at runtime. The three reference blocks (container map, signature types, declaration card) are OFF by default because they cost turns in agent sessions; --reference-blocks all turns them on for interactive reading.\n\n" +
 			"--top-k only changes how many results come back; --deep additionally runs the exhaustive sparse (BM25) pass and fuses it with the semantic ranking (slower, reads every eligible file).\n\n" +
-			"Ranking returns one region per file, which is right for code and wrong for prose: one markdown document can hold the answer across several distant regions. When fewer distinct files match than --top-k asked for, the spare slots are spent returning those finer regions of the same document as results of their own (multi-resolution retrieval) — strictly additive, so it never displaces a file and never breaches --max-context-bytes. --single-resolution turns it off.",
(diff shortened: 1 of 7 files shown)


suhaanthayyil:

PR #104, opened 2026-08-15
feat(search): prose retrieval by section, with the benchmark kit that measures it

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/55

Two retrieval fixes for prose, plus the benchmark kit that measures them. Supersedes #100, #103
and #101, which are closed in favour of this.

Code is `internal/sem` only (438 insertions); everything else is `bench/memory/` — docs, adapters
and patch files. No Go outside the prose path is touched.

---

## The defect

`selectSearchCandidates` admits **one candidate per file**. For code that is right — a file is
the thing you open. For prose it is wrong: a markdown document's heading regions already exist in
the graph as `section` entities (`parser.go` `markdownEntities`), are already scored, and are then
collapsed to one per file.

Measured on a real prose corpus at `--top-k 200`, the shipped build returned **19 results across
19 files while carrying 195 passages holding 42 KB of evidence** that no `results[]` consumer
could reach. Ranks 6–19 got a one-line snippet.

**Commit 1 — multi-resolution prose retrieval.** When fewer distinct files match than `--top-k`
asked for, spend the spare slots promoting those regions to results of their own. Strictly
additive: runs after selection, only fills slots the distinct-file ranking left empty, never
breaches `--max-context-bytes`. `--single-resolution` restores the old output.

That alone was not enough, because promoted regions are **satellites**: `Score: parent.Score`
verbatim, ordered by a round-robin over *parent index*, appended after every parent. Result:

```
200 results  ->  19 distinct scores
```

181 of 200 are score-duplicates of one of 19 parents. A consumer sorting by score with a stable
sort therefore gets 19 blocks — one document's parent followed by its regions, then the next.
Truncate to the top 50/20/10 and you see only the first one or two documents. Evidence that must
be stitched across documents structurally cannot surface.

**Commit 2 — rank prose sections as independent candidates.** The retrieval unit becomes the
section (`proseParentKey`, keyed by the heading's SymbolID, falling back to file path where a
document has no section structure). A section competes on its own score, as an independent file
would. Breadth is preserved: **no document gets a second region until every document that fits
has one**, binding every selection pass. `--document-resolution` restores the previous behaviour.

Both changes are query-blind. The unit comes from the parser's section entities and the file
extension; nothing reads the question.

## Measured — 152 questions, real prose corpus, identical search arguments

| | items/q | distinct regions/q | **distinct scores/q** | files/q | visible chars/q |
|---|---|---|---|---|---|
| before | 195.0 | 243.0 | **17.1** | 18.9 | 60,854 |
| **after** | **196.0** | **295.3** | **84.9** | **18.9** | **53,929** |
| off switch | 195.0 | 243.0 | 17.1 | 18.9 | 60,854 |

Score diversity 17.1 → 84.9, document coverage unchanged, and **11.4% fewer bytes** — the old
build computed regions and discarded them unread. The off switch is byte-identical to the
previous behaviour on every metric.

## Code search is unaffected

**30 of 30 full-JSON payloads identical** — 15 real locate queries against this repo at
`--top-k 10` *and* `--top-k 50`, comparing entire payloads (results element by element plus
`stats`, `commit`, `tree`, `completeness`, `diagnostics`), latency telemetry excluded.

The prose branch sits behind `proseParentCorpus`, requiring ≥4 units and ≥80% doc-extension
paths — a source repo cannot reach it. Verified rather than asserted: three deliberately
documentation-flavoured queries against this Go repo returned **zero** results carrying
`retrieval_mode=prose-parent` in either build.

## Benchmark result

LoCoMo, all 1540 questions, answerer **and** judge both `gpt-5.6-sol`, `top_k=200`, against a
mem0 OSS control **running concurrently in the same window**. No harness flag, no ingest change —
the shipped corpus shape.

| | overall | single-hop | multi-hop | temporal | open-domain |
|---|---|---|---|---|---|
| **entire-graph** | **94.74** | **96.67** | **93.97** | **95.33** | 78.12 |
| mem0 OSS | 93.83 | 95.12 | 93.62 | 94.70 | **80.21** |

**+0.91pp overall, discordant 43–29, McNemar p = 0.125 — statistically tied, point estimate
favours entire-graph.** That wording follows a decision rule registered before the run finished.
**single-hop +1.55pp, p = 0.035** is the one cell significant on its own.

Multi-hop moved from **−1.44 against mem0 to +0.35** — the deficit this change targets, and the
category the 19-distinct-scores collapse was starving.

The isolated effect of commit 1, measured paired on the same corpus with the binary as the only
variable: **91.56 → 93.83, +2.27pp, discordant 46–11, p = 3.3e-06, n = 1540.**

### Stated plainly

- Run-to-run noise on this harness is **0.65pp** — two identical mem0 runs scored 93.83 and 93.57.
- entire-graph delivers ~30% more context per question, and **mem0 is more context-efficient
  below ~25k chars** (43.2 accuracy points per 1k chars at top-10 against our 5.7). `top_k`
  equalises item count, not context volume.
- Prose search is **2.1× slower per query** (0.300s → 0.628s); passages are planned for ~196
  units instead of 19. Code-search latency is unchanged.
- Open-domain is our one losing category.

## The benchmark kit

`bench/memory/` makes every number above reproducible: the entire-graph adapter, the competitor
ports we wrote, the fairness spec, a run index, and apply-checked patches for the four upstream
files we modified. Upstream is pinned by **md5-verified fork point**, not a guessed SHA, and the
judge/answerer prompts are byte-identical to upstream — we did not tune the grading.

**Three competitor defects were found and fixed in the competitors' favour**: mem0 **+6.04**
(a `limit` kwarg silently capping search at 20 memories instead of 200), cognee **+13.77**
(state on disk with an in-memory retrieval buffer, so resumed runs searched nothing), and cmm
from **structurally zero** (its shipped BM25 excludes `Section` nodes, so prose returns no
results). A fourth, against mem0, is disclosed rather than fixed — it is measured at 0.13pp.

## Tests

`mise run check` green. Three regressions were found and fixed during development, each pinned by
a test: whole-document exclusion at small top-k, loss of same-document passages at tight top-k,
and one region returned up to seven times.


ashtom:
## Review of the Go changes (agent-generated)

**Attribution up front:** this review and the inline comments that follow were produced by Claude Code and posted from @ashtom's account. They are not a human reviewer's judgement — treat them as evidence to check, not verdicts. Anything that reads as confident is still worth verifying yourself.

**Scope:** the Go diff only (`internal/cli`, `internal/sem` — 438 insertions). `bench/memory/` is documentation and Python and was left out of correctness scope.

**Method:** the branch was built in a throwaway worktree and probed against synthetic prose corpora, with every number below compared against the same probe on `main`. `go vet ./internal/sem` and the full `internal/sem` suite pass on this branch (69 s).

**Verified as working as described**, so these are not in doubt:
- breadth-before-depth holds — 40 docs at `--top-k 10/40/80` gave 10/40/40 distinct documents, max-per-file 1/1/6
- cross-result passage duplication measured **0** at every top-k tried; `dropSelectedProsePassages` does its job for passages
- `plan[:0]` in-place filtering is safe — `planProseParentPassages` returns a freshly allocated slice per call
- `documents[parent.path]` cannot be nil in either mode, and `parents[key]` lookups are always populated

**7 findings inline.** One is worth blocking on:

| Severity | Where | What |
|---|---|---|
| **high** | `search_prose_parent.go:45` | section units leak into code results and silently override `--max-regions-per-file` |
| medium | `search_multi_resolution.go:126` | promoted results break descending-score ordering of `results[]` |
| medium | `search_prose_parent.go:137` | nested same-file results reprint text already shown by a higher rank |
| medium | `search_multi_resolution.go:135` | promotion fitter re-serializes the whole result set per candidate count |
| low | `search.go:187` | `SingleResolution` doc comment promises a shape it does not restore |
| low | `search_prose_session_test.go` | the scale guard was retargeted to the non-default path |
| low | `help.go:130` | help presents the two levers as alternatives when they stack |

On the high one specifically: the PR's own check that *"code search is unaffected — 30 of 30 payloads identical"* cannot reach the affected path, because it was run against this Go repo, which fails the ≥80%-doc-paths gate that opens it. That is worth knowing regardless of what you decide about the finding itself.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/sem/search_prose_parent.go
```diff
@@ -10,12 +11,44 @@ const (
...
+// So a region that the parser recognized as belonging to a distinct section (markdown headings are
+// parsed as `section` symbols) is its own unit, and competes on its own score exactly as an
+// independent file would. A region with no symbol identity has no section to belong to, so it
+// falls back to the file — which is the previous behaviour, and is what prose formats with no
+// heading structure keep getting.
+//
+// The line is part of the key because a symbol ID names a heading, not an OCCURRENCE of one: a
+// document that heads twelve entries "## note" carries twelve identically-identified sections, and
+// keying by ID alone would collapse them back into the single unit this change exists to split.
+func proseParentKey(candidate searchCandidate, sectionUnits bool) string {
```
ashtom, line 52: ### Section units leak into code results, and silently override `--max-regions-per-file`

Once the prose gate opens, `proseParentKey` keys by `SymbolID:StartLine` for **every** candidate — and the gate only requires ≥80% of matched *paths* to be doc extensions, so the source files in a docs-heavy repo come along with it. The prose selector never consults `maxPerFile` except via `baseline`.

Measured on a repo of 20 `.md` + 3 `.go` files, `--top-k 30`:

| | results | regions of `pkg/file00.go` |
|---|---|---|
| `main` | 23 | 1 |
| this PR (default) | 31 | **7** |

`MaxRegionsPerFile` defaults to 3. Under document resolution the one-unit-per-file rule made that cap moot; section units remove the rule, and nothing reinstates the cap.

**Why the PR's own check misses it:** the "code search is unaffected — 30 of 30 payloads identical" verification ran against this Go repo, whose path mix fails the ≥80% gate, so it never reaches this code at all. A mixed docs+code fixture would.

Suggested direction: consult `maxPerFile` in the prose selector, or restrict section keying to candidates whose own path is a doc extension rather than relying on the repo-level gate.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/sem/search_multi_resolution.go
```diff
@@ -0,0 +1,143 @@
...
+		Language:         parent.Language,
+		Signals:          appendUnique(signals, proseResolutionSignal),
+		Section:          parent.Section,
+		Snippet:          passage.Snippet,
+	}
+}
+
+// expandProseResolution spends the result slots the distinct-file ranking left empty on the finer
+// regions it already computed, without breaching the caller's byte ceiling.
+func expandProseResolution(results []SearchResult, topK, budget int) []SearchResult {
```
ashtom, line 164: ### Promoted results break descending-score ordering of `results[]`

`applyProseResolutionPromotions` appends promotions after each parent, while `proseResolutionResult` inherits `parent.Score` verbatim. The result array is therefore no longer monotone in score.

Measured on 6 docs × 10 turns at `--top-k 200`: **rank 66 carries score `1.6592`, rank 67 (a promoted passage) carries `15.9715`.**

Every pre-PR payload I probed was monotone (`--top-k` 20/50/100 all clean), so this is new. It matters because it breaks precisely the consumer the PR description reasons about: anything that sorts or truncates by score now sees a ranking that contradicts its own rank numbering — a client taking "top 20 by score" gets a different set than the first 20 entries.

Two ways out, depending on intent: give promotions a score derived from their own match rather than the parent's, or state in the schema that `results[]` is rank-ordered and *not* score-ordered, so consumers stop trusting the score for ordering.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/sem/search_prose_parent.go
```diff
@@ -61,43 +94,90 @@ func selectSearchCandidates(
...
+			return false
+		}
+		// BREADTH BEFORE DEPTH. A question is often answered by the one document nobody's terms
+		// ranked highly, so no document gets a SECOND region until every document that fits has
+		// one. The file-keyed unit got this for free — one unit per document meant selection could
+		// not return to a document — and it is the one property section units could lose: a pure
+		// score ranking hands every slot to whichever document's sections dominate the score list.
+		// It binds every pass, not just the first, because the per-term coverage pass is where a
+		// weakly-scoring document has always been admitted.
+		if sectionUnits && len(selectedPaths) < documentTarget && selectedPaths[parent.path] {
```
ashtom, line 144: ### Section results nested inside a higher-ranked same-file window reprint the same text

Prose head ranks get an 80-line head window (`resolvedSearchHeadWindowLines`), and section units now place several results from one file in a single payload. Lower-ranked results can therefore fall entirely inside a higher-ranked result's printed snippet.

Measured on 8 docs × 12 turns at `--top-k 20`: rank 1 prints `sessions/session-00.md` lines **1–63**; ranks 9, 10, 11 and 12 return the same file at lines **8, 13, 18, 23** — all already visible in rank 1. At `--top-k 100`: **59 overlapping same-file result pairs out of 100 results.**

Since the payload is budgeted and replayed into a model on every later turn, those are bytes spent twice.

**Why the tests don't catch it:** `TestSearchRepositoryReturnsEveryProseRegionOnce` compares exact `(SnippetStartLine, SnippetEndLine)` tuples, so containment is invisible to it — distinct tuples pass even when one span sits inside another. And `dropSelectedProsePassages` dedupes against pre-growth spans, never re-checking after `allocateSearchSnippets` grows them.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/sem/search_multi_resolution.go
```diff
@@ -0,0 +1,143 @@
...
+	if topK <= 0 || len(results) >= topK {
+		return results
+	}
+	order := planProseResolutionPromotions(results, topK-len(results))
+	if len(order) == 0 {
+		return results
+	}
+	expanded := results
+	for count := 1; count <= len(order); count++ {
```
ashtom: ### The promotion fitter re-serializes the whole result set once per candidate count

```go
for count := 1; count <= len(order); count++ {
    trial := applyProseResolutionPromotions(...)
    serializedSearchResultBytes(trial)
}
```

Each iteration rebuilds and JSON-marshals up to `topK` results. Benchmarked at 19 parents × 12 passages, `--top-k 200`, 4 MB budget: **25.6 ms and 39 MB allocated per call**, 181 serializations.

Serialized size is strictly monotone in `count` — a promotion always moves a passage into a strictly larger result record — so a binary search finds the same fit in ~8 serializations.

Separately: when `budget <= 0` the serialization short-circuits, but the loop still performs `len(order)` full slice copies where a single call with `count = len(order)` would do.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/sem/search.go
```diff
@@ -184,6 +184,21 @@ type SearchOptions struct {
 	// for one more result is not a request to change retrieval strategy, so the strategy is
 	// now an explicit choice and top-k only changes how many rows come back.
 	Deep bool
+	// SingleResolution turns OFF multi-resolution prose retrieval, restoring the previous output
```
ashtom: ### `SingleResolution`'s doc comment states a contract the code does not honour

The comment says the flag restores "the previous output exactly: one result per file".

Measured with 8 docs at `--top-k 40`, `SingleResolution: true`: **40 results over 8 files**, with `sessions/session-00.md` supplying 10 of them. Only `SingleResolution && DocumentResolution` together reproduce the old shape — 23 results, byte-matching `main` on the mixed corpus.

This one is worth fixing even though it is only a comment: AGENTS.md names entire-brain as a downstream consumer of this package, and a caller reading this comment to restore prior behaviour sets one flag and gets something else. Either correct the comment to name both flags, or make `SingleResolution` imply document resolution so the comment becomes true.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/sem/search_prose_session_test.go
```diff
@@ -842,16 +849,16 @@ func TestSelectSearchCandidatesHasLinearProseWorkingSet(t *testing.T) {
 	small := proseScaleCandidates(500)
```
ashtom, line 849: ### The scale guard now covers only the non-default path

`TestSelectSearchCandidatesHasLinearProseWorkingSet` (line 846) and `BenchmarkSelectSearchCandidatesProse4000` (line 923) were both updated to pass `sectionUnits=false` — and `proseScaleCandidates` emits one candidate per file, so section units would be a no-op there even if the flag were flipped.

That leaves the **default** path unpinned, and it is the more expensive one: a second full `proseParents` pass, a per-section `planProseParentPassages` call over the whole document's candidate list, and a term-list key space that grows with sections rather than files.

Measured at 200 files × 20 sections, `--top-k 20`:

| | allocs | bytes |
|---|---|---|
| before | 2,486 | 2.0 MB |
| after | **23,219** | 4.4 MB |

A guard that only covers the path the product does not take will not notice the next regression. Worth a fixture with real sections so the default path has a ceiling.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/cli/help.go
```diff
@@ -127,7 +127,8 @@ var commandDocs = []commandDoc{
...
 		long: "Ranked source regions for a plain-language description, with source and file:line inline, budgeted to drop straight into context. This is the first move for almost every locate task.\n\n" +
 			"By default search returns: ranked candidate fix sites (top hits as full function bodies), RELATED SITES, the COVERING TEST plus other tests over the same code (ALSO COVERING), SAME-CONCEPT LITERAL (every place the concept is named, tagged EDIT/CONSUMER/DOC), a VERIFY line (the narrowest test command for the file), and a CLOSED-SET WARNING when a switch over a sealed set would fail at runtime. The three reference blocks (container map, signature types, declaration card) are OFF by default because they cost turns in agent sessions; --reference-blocks all turns them on for interactive reading.\n\n" +
-			"--top-k only changes how many results come back; --deep additionally runs the exhaustive sparse (BM25) pass and fuses it with the semantic ranking (slower, reads every eligible file).",
+			"--top-k only changes how many results come back; --deep additionally runs the exhaustive sparse (BM25) pass and fuses it with the semantic ranking (slower, reads every eligible file).\n\n" +
```
ashtom, line 130: ### The help text presents the two levers as alternatives when they stack

The wording reads: *"Prose with no heading structure has nothing to split on; there, when fewer distinct units match than --top-k asked for, the spare slots are spent returning finer regions"* — which says multi-resolution is what happens **instead of** section splitting, for unheaded prose.

Measured: headed markdown with section resolution active at `--top-k 200` still produced **33 results tagged `retrieval_mode=prose-passage`**. Multi-resolution runs on headed prose too.

That is not just a wording nit — it is the same stacking that produces the score-ordering break reported on `search_multi_resolution.go`. A reader who believes the two are alternatives will not expect promoted passages in a headed-prose payload at all.

---
<sub>🤖 Agent-generated review finding. Posted from @ashtom's account, but written by Claude Code, not by a human reviewer — weigh it accordingly. Every measurement above was produced by building this branch in a throwaway worktree and running probes against synthetic corpora, comparing to `main`.</sub>

internal/sem/search_prose_parent.go
suhaanthayyil: Confirmed and fixed in 7ef28b9.

You identified the mechanism exactly: the gate counts the whole matched path mix, so a docs-heavy repo opens it while still holding source files, and every candidate was then keyed by symbol. Since one-region-per-unit is what enforces `MaxRegionsPerFile` for code, removing the rule uncapped it.

Took your second suggested direction — restricting section keying to candidates whose own path is a doc extension — rather than consulting `maxPerFile` in the prose selector, because it fixes the cause rather than re-imposing a cap on a unit model that should not have applied.

One thing your repro surfaced that the finding did not name: passage planning had the same leak. Even with section keying restricted, a code file still got prose passages, and multi-resolution then promoted them into extra result slots — the same uncapping by a different route. `planProseParentPassages` is now also gated on `proseParentPath`.

Measured on 20 `.md` + 3 `.go` at `--top-k 30`, counting ranked results only (score > 0; your table appears to include the score-0 related-sites block, which is a separate subsystem and legitimately repeats a file):

| | ranked regions of one `.go` file |
|---|---|
| before | 2 |
| after | **1** |

Also added the mixed-fixture gap you pointed at: code-search payloads are now verified byte-identical across 21 query/top-k pairs on a Go repo, with latency telemetry excluded — the first version of that check reported 0/21 identical purely because `index_latency_ms` varies per run.

internal/sem/search_multi_resolution.go
suhaanthayyil: Confirmed and fixed in 7ef28b9.

`results[]` is now stably re-sorted by score before ranks are assigned. Stability is what makes it safe: a promotion ties with its parent exactly, so it settles immediately after that parent rather than anywhere else in the tie. Set membership is untouched, so promotion stays strictly additive — only the reported order changes.

Chose this over deriving a score from the passage own match. A passage has no independent ranking evidence; inventing one would make the number mean something no ranking produced, which is the objection the original comment on `proseResolutionResult` raises. And documenting `results[]` as rank-ordered-but-not-score-ordered would leave every existing consumer wrong rather than fixing them.

Added `TestExpandProseResolutionKeepsResultsScoreOrdered` as the regression guard. `TestExpandProseResolutionRoundRobinsAcrossParents` had to change with it — it asserted promotion order positionally, which is no longer meaningful. Round robin is a property of *which* passages are promoted, so it now asserts the promoted set (both documents contribute before either contributes twice) rather than array positions.

internal/sem/search_multi_resolution.go
suhaanthayyil: Confirmed and fixed in 7ef28b9.

Replaced the linear scan with bisection on exactly the monotonicity argument you gave — a promotion always moves a passage out of a `passages` entry and into a full result record, which is strictly larger, so serialized size is strictly increasing in count. ~log2(n) serializations instead of n.

Also took the `budget <= 0` case: it now applies every planned promotion in one call rather than looping `len(order)` times over full slice copies.

internal/sem/search.go
suhaanthayyil: Confirmed and fixed in 7ef28b9 — comment corrected rather than behaviour changed.

Your measurement is right: only `SingleResolution && DocumentResolution` together reproduce the previous shape. The comment now says so explicitly.

Went with correcting the comment rather than making `SingleResolution` imply document resolution, because the two are genuinely independent knobs and collapsing them would remove a real configuration — turning off passage promotion while keeping section ranking. Agreed on the stakes though: entire-brain consumes this package, and a caller setting one flag on the strength of that sentence would have gotten something else.

internal/sem/search_prose_session_test.go
suhaanthayyil: Confirmed and fixed in 7ef28b9.

You are right that the guard was inert: `proseScaleCandidates` emits one candidate per file, so a section key and a file key name the same unit there and `sectionUnits` is a no-op whichever way it is passed.

Added `proseSectionScaleCandidates(files, sections)` — a fixture with real sections — plus `TestSelectSearchCandidatesHasLinearProseSectionWorkingSet` and `BenchmarkSelectSearchCandidatesProseSections4000` on the default path.

Measured at your 200 files x 20 sections, `--top-k 20`: **23,639 allocs, 4.38 MB**, which lands on your 23,219 / 4.4 MB. The guard pins linearity (doubling the corpus may not more than 2.4x allocations) rather than an absolute number, and additionally asserts that breadth-before-depth still holds at scale — 20 results over 4,000 sections must span 20 distinct documents, which is the property most likely to be lost quietly.

internal/cli/help.go
suhaanthayyil: Confirmed and fixed in 7ef28b9. Rewrote the sentence so the two levers read as stacking rather than as alternatives, and stated it outright: a headed document can be ranked by section *and* have spare slots filled with promoted passages, so a prose payload may carry both.

internal/sem/search_prose_parent.go
suhaanthayyil: Partially actioned in 7ef28b9, and I want to be straight about what I could and could not verify.

**The mechanism you describe is real and I fixed it.** `dropSelectedProsePassages` dedupes spans *as selected*; `allocateSearchSnippets` then grows them, and nothing re-checked afterwards. `dropContainedProseResults` now runs after growth and — deliberately — *before* promotion, so slots freed by a duplicate are spent on regions the payload does not already show rather than simply lost. You are also right about the test: `TestSearchRepositoryReturnsEveryProseRegionOnce` compares exact `(SnippetStartLine, SnippetEndLine)` tuples, so containment is invisible to it.

**But I could not reproduce the measurement.** Four fixture shapes, both binaries, containment count 0 in every one:

| fixture | top-k | contained (before / after) |
|---|---|---|
| your 8 docs x 12 turns | 20 | 0 / 0 |
| 19 sessions x 24 turns | 200 | 0 / 0 |
| 5 docs, one dominant (40 sections) | 20 | 0 / 0 |
| 8 docs, one dominant by term density | 20 | 0 / 0 |

The pattern is consistent: documents that receive the 80-line head window get **exactly one** result, and depth slots go to *different* documents. In the 8x12 case the per-file split was `{00:1, 01:1, 02:1, 03:1, 04:1, 05:5, 06:5, 07:5}` — sessions 00-04 hold the head windows, 05-07 hold the depth. Containment needs one file to have both, and breadth-before-depth appears to prevent that: no document takes a second region until every document that fits has one, and by then the head ranks are spent. I could reproduce the *ingredient* (rank 1 printing `session-00.md` lines 1-80) but never the collision.

So the fix ships as defensive. It is provably inert when nothing is contained — retrieved sets and snippet bytes are byte-identical before/after across every fixture and budget I tested — and correct when something is. But I have not observed it change a payload, which means it is unguarded by a regression test, and I would rather say that than write a test asserting behaviour I only reason about.

If your 63-line rank-1 window co-occurred with same-file tail results, something in your corpus differs from all four of mine. The `--top-k 20` / 8-doc / 12-turn shape is the one I most expected to reproduce it. If you still have that fixture, I will take it and write the missing test.

suhaanthayyil:
All seven review findings addressed in `7ef28b9`. Detail is in the inline thread replies; summary here since those are easy to miss.

**Six confirmed and fixed.**

| # | finding | fix |
|---|---|---|
| 1 | Section units leaked into code, uncapping `--max-regions-per-file` | Section keying **and** passage planning restricted to doc-extension paths |
| 2 | Promotions broke descending-score ordering of `results[]` | `results[]` stably re-sorted; a promotion settles beside its parent |
| 4 | Promotion fitter re-serialized the whole set per candidate count | Bisection on monotonic size: ~log2(n) serializations instead of n |
| 5 | `SingleResolution` doc comment stated a contract the code did not honour | Comment corrected to name both flags |
| 6 | Scale guard covered only the non-default path | Real-sections fixture + linearity guard + benchmark on the default path |
| 7 | Help text presented the two levers as alternatives | Reworded; states outright that they stack |

Finding 1 had a second route the review did not name: even with section keying restricted, a code file still received prose *passages*, and multi-resolution promoted them into extra result slots — the same uncapping by a different path. Both are now gated.

**One partially actioned.** Finding 3 (nested same-file snippets) describes a real mechanism and I fixed it — dedup ran on pre-growth spans and nothing re-checked after `allocateSearchSnippets` grew them. But I could not reproduce the measurement in four fixture shapes, including the reviewer own 8 docs x 12 turns at `--top-k 20`; containment count was 0 in every one, before and after. Head-window documents consistently receive exactly one result while depth slots go to other documents, and breadth-before-depth appears to prevent one file from having both. The fix ships defensive and I have left that thread open asking for the fixture, rather than writing a regression test for behaviour I only reason about.

**Verification.**

- Code-search payloads byte-identical across 21 query/top-k pairs on a Go repo (latency telemetry excluded)
- On a LoCoMo-shaped corpus the retrieved set and snippet bytes are identical at budgets 24k / 64k / 256k — only ordering changes, which was the bug
- Ranked regions of one `.go` file in a mixed 20 `.md` + 3 `.go` corpus: 2 -> 1
- Default-path scale benchmark: 23,639 allocs / 4.38 MB at 200 files x 20 sections
- `gofmt` clean, `go vet` clean, 2,725 tests pass

A confirming LoCoMo re-run of the eg arm is in flight against this commit, reusing the exact `sw_eg_mr3` configuration that produced the published number. Result will be posted here when it lands.

suhaanthayyil:
**Confirming LoCoMo re-run against `7ef28b9` — the number holds.**

Re-ran the eg arm with the review-fix binary, reusing the exact configuration that produced the published figure (`sw_eg_mr3`): same answerer and judge (`gpt-5.6-sol`), `top_k=200`, `FAIR_MODE=1`, same `--max-workers 3 --question-workers 10 --rpm 60`. Only the binary, run id and state root changed.

| run | binary | LoCoMo (n=1540) |
|---|---|---|
| `sw_eg_mr3` | pre-fix | 94.74 |
| `sw_eg_rev1` | **`7ef28b9`** | **95.19** |

Paired on all 1,540 questions: **+0.45, discordant 16/9, McNemar exact p = 0.23.**

**That is a reproduction, not an improvement.** The delta sits below this harness measured 0.65-point noise floor, the two runs are in different serving windows (Aug 14 vs Aug 17) which our own protocol excludes from comparison, and p = 0.23 does not clear significance. The correct reading is that the retrieval fixes did not move the benchmark — which is what the local evidence predicted: retrieved set and snippet bytes are byte-identical before and after at every budget tested, and only the ordering of `results[]` changed.

Run health: 1540/1540 graded, **0 empty answers**, **0 failed ingest chunks**, median 200 items retrieved (full budget used), 0 timeouts.

Cutoff curve for the fixed build: top_200 95.19 · top_50 93.31 · top_20 91.49 · top_10 90.19.

suhaanthayyil:
This branch has been merged into #113, which now carries it forward along with the README/docs refresh and the benchmark comparison this PR's work feeds. Once #113 merges, these commits are already on `main` through it — this PR should be closed as superseded rather than merged separately, to avoid a duplicate/no-op merge.

suhaanthayyil:
Closing as superseded — merged into #113, which carries these commits forward. See the comment above.

suhaanthayyil:
Reopening — this is being merged into main directly, not closed as superseded.

PR #105, opened 2026-08-15
fix: validate init-agents markers before writes

ashtom:
https://entire.io/gh/entireio/entire-graph/trails/60

## Summary

- inspect and validate both `AGENTS.md` and `CLAUDE.md` before creating or modifying any activation file
- accept only no Entire Graph markers or exactly one correctly ordered managed-marker pair
- reject malformed, reversed, or duplicate markers with a path-specific recovery error and no writes or success output
- render from the validated snapshots while preserving existing symlink/hard-link behavior, including first-run dangling aliases

## Why

`init-agents` previously overwrote `.entire/graph-agent.md` before inspecting either instruction file. Its block upsert then used the first raw opening and closing markers without validating their count or order. Malformed or duplicate markers could therefore append more blocks or replace an unexpectedly broad range after another activation file had already changed.

This patch makes marker/read preflight failures non-mutating. Successful activation still intentionally replaces the generated guide in full and preserves user-owned text outside one valid managed block.

## Verification

- `mise exec -- go test ./internal/cli -run '^TestInitAgents' -count=1`
- `mise exec -- go test ./internal/cli`
- `mise exec -- go test ./internal/cli -run '^TestCompactNDJSONRejectsTargetedRelationFilters$'`
- `mise run check`
- two independent focused reviews: clean

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit a3f5e21846412d221ddcbb4a118ae4097499ce3c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/cli/agents.go
+++ b/internal/cli/agents.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"os"
@@ -96,14 +97,6 @@ func runInitAgents(opts Options, args []string) error {
 	}
 
 	guidePath := filepath.Join(root, ".entire", "graph-agent.md")
-	if err := os.MkdirAll(filepath.Dir(guidePath), 0o755); err != nil {
-		return err
-	}
-	if err := os.WriteFile(guidePath, []byte(agentGuide), 0o644); err != nil {
-		return err
-	}
-	fmt.Fprintf(opts.Stdout, "wrote %s\n", guidePath)
-
 	pointer := agentPointerBegin + "\n" +
 		"This repo has the entire-graph code graph installed. Before exploring code with\n" +
 		"grep/find/whole-file reads, read .entire/graph-agent.md — resolution-first guidance\n" +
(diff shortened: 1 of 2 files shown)


internal/cli/agents.go
```diff
@@ -112,53 +105,146 @@ func runInitAgents(opts Options, args []string) error {
...
+	if err != nil {
+		return nil, -1, -1, err
+	}
+	return content, begin, end, nil
+}
+
+// validatePointerMarkers intentionally counts the raw marker tokens. Markers in examples,
+// comments, or other Markdown regions are still managed-marker tokens and must not make the
+// replacement range ambiguous.
+func validatePointerMarkers(path string, content []byte) (int, int, error) {
```
ashtom, line 243: `validatePointerMarkers` counts raw marker tokens anywhere in the file, including inside inline code and fenced blocks — and the sibling `claudeDirectlyImportsAgents` in this same file already skips code spans, fences and comments, so the file now disagrees with itself about what a live marker is.

The consequence is worse than a spurious error, because this check runs *before* the guide write. An AGENTS.md or CLAUDE.md that merely documents the markers — or a migration that concatenates the two files — is rejected permanently, and `.entire/graph-agent.md` then silently stops being refreshed on every upgrade. There is no `--force` to get out of it.

Reusing the same code-span/fence skipping as `claudeDirectlyImportsAgents` would make the two agree, and would keep a file that *talks about* the markers from being treated as a file that *has* them.

Found while reviewing this PR; the core change (validating before writing) verified correct against a 19-scenario differential harness vs. the merge base.

internal/cli/agents.go
```diff
@@ -112,53 +105,146 @@ func runInitAgents(opts Options, args []string) error {
...
+func inspectInstructionFile(path string) (os.FileInfo, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("%s: inspect target: %w", path, err)
+	}
+	if !info.Mode().IsRegular() {
+		return nil, fmt.Errorf("%s: expected a regular file, found %s", path, info.Mode().Type())
```
ashtom: Minor: this formats `info.Mode().Type()` with `%s`, which renders the permission-bit form rather than a name. For a directory the error reads:

```
expected a regular file, found d---------
```

(confirmed by running it). Something like naming the type explicitly — `directory`, `symlink`, `named pipe` — would make the diagnostic actionable for whoever hits it.

ashtom:
bugbot run

internal/cli/agents.go
ashtom: Withdrawing this one — I was wrong that the file contradicts itself by accident.

The raw counting is deliberate and specified: the comment above `validatePointerMarkers` says so, and `agents_test.go` pins it with a `"marker in example"` case asserting that a marker inside an inline code span must be rejected. The reason holds up too, which I missed first time round: `renderPointerBlock` replaces by raw byte offsets, so validator and writer share one definition of a marker. Making the validator fence-aware while the writer stayed raw would trade a loud refusal for silent corruption of user text — strictly worse than what's here.

What survives is narrower than I framed it: a repo that *documents* the markers in a fence gets blocked, and `.entire/graph-agent.md` then stops being refreshed. I floated a `--force` for that and then couldn't design one worth having — "replace the first range anyway" reintroduces exactly the bug this PR prevents, "write the guide but skip the pointers" contradicts the no-writes guarantee `TestInitAgentsMalformedMarkersDoNotCreateMissingOutputs` pins, and "append a fresh block" makes the ambiguity permanent.

The coherent fix would be to make `validatePointerMarkers` and `renderPointerBlock` agree on a live marker — validator returns the live pair's offsets, writer uses them — so a fenced example is invisible to both. That's a real change to a tested decision and yours to make. For a case this rare, which fails loudly with instructions that say how to repair the file, I wouldn't spend it.

Resolving; the other finding is fixed in 5a311c3.

internal/cli/agents.go
ashtom: Fixed in 5a311c3 — reports `found directory` via a `fileTypeName` helper, and the test now requires a named type rather than only the phrase "regular file". Cursor Bugbot independently flagged the same line.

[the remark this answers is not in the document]
ashtom: Fixed in 5a311c3 — `fileTypeName` now names the type (`found directory`), and the test asserts the named form so the permission-bit rendering cannot come back.

ashtom:
bugbot run

PR #106, opened 2026-08-15
feat: report explain's source view in the JSON answer

ashtom:
https://entire.io/gh/entireio/entire-graph/trails/56

Follow-up from the review of #99. Stacked on `agent/fix-head-source-provenance`; GitHub will retarget this to `main` when #99 merges.

## Problem

#99 made `explain` state which tree it read — but only in the text/agent header. `ExplainResponse` carried no provenance at all, so `explain --head --format json` and `explain --format json` were **byte-identical on a clean tree**. A machine consumer could not tell whether the declarations, line numbers and signatures it just received described the committed tree or the working tree.

## Change

`ExplainResponse` gains `commit`, `tree`, and `worktree_snapshot` — the same field names and semantics `def`, `neighbors` and `impact` already expose. Schema `1.x` is frozen and additive-only, so new optional fields stay in contract.

The text header and the JSON fields are now computed from **one** expression (`useHead`), so the two renderings cannot drift apart about which tree they read — the exact defect class #99 exists to remove.

## Verification

`go vet ./internal/cli`, `go test ./...` — all pass. New `TestExplainJSONReportsSelectedTree` asserts `worktree_snapshot` flips between the two modes and that `--head` reports the actual HEAD commit.

https://claude.ai/code/session_01PRZZprZMazfmyRkdDPVbDV

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 626604a9fc4f6703b8dea48078839cbf2c1fad39. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/cli/explain.go
+++ b/internal/cli/explain.go
@@ -104,6 +104,13 @@ type ExplainResponse struct {
 	// Scanned is how many candidate names the error text yielded, so a truncated answer is visible.
 	Scanned int `json:"scanned"`
 	Omitted int `json:"omitted,omitempty"`
+	// Commit / Tree / WorktreeSnapshot are the provenance of the declarations above: WHICH tree these
+	// line numbers and signatures describe. The text form says it in its header; without these a JSON
+	// consumer had no way to tell a --head answer from a working-tree one, and on a clean tree the two
+	// were byte-identical. Same field names and semantics as def, neighbors, and impact.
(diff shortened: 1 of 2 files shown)


ashtom:
bugbot run

PR #110, opened 2026-08-16
fix(sem): key both caches on the resolved file cap, not the raw option

ashtom:
Found while reviewing #109, whose premise is that incompatible cache variants safely rebuild. For one input they don't.

## The bug

`searchSnapshotKey` and `providerRecordsKey` both hash the **raw** `options.MaxFiles`. The cap that actually shaped the build is `resolveMaxSourceFiles(options.MaxFiles)`, which falls back to `ENTIRE_GRAPH_MAX_FILES` and then to the default. So the env var never reached either key: an `ENTIRE_GRAPH_MAX_FILES=1` index and an uncapped search both key on `max-files=0` and share an entry.

It cuts both ways:

- a **capped** entry serves a later **uncapped** query — a truncated graph presented as complete
- a **higher-cap** entry survives a later **lowered** cap — more of the tree than the caller asked to see

Neither direction announces itself. Measured on a two-file repo, after a capped index:

| | uncapped `search --head --profile full` |
|---|---|
| before | `symbols_considered=1` — served the truncated graph |
| after | `symbols_considered=2` — rebuilds |

The answer arrives with `W_FILE_LIMIT` suppressed, so nothing in it says the graph is partial. The records cache has the same bug and is **on by default**.

## How it got here

`5751d26` — the commit that fixed *"the file cap is not keyed"* — introduced it:

- its commit message says *"options.MaxFiles (and ENTIRE_GRAPH_MAX_FILES behind it) ... Both keys now fold it in"*
- the ADR it shipped (`0002`, decision 2) says the key uses *"`resolveMaxSourceFiles(options.MaxFiles)`, the resolved value, so `ENTIRE_GRAPH_MAX_FILES` is covered by the same term as the option"*
- the code it shipped hashes the raw value
- both regression tests it added pass `ProviderSnapshotOptions{MaxFiles: 5}` — the option path only. No test in the repo sets `ENTIRE_GRAPH_MAX_FILES`; the one test that references the constant asserts a warning string.

The quoted mutation check ("removing the key terms makes both regression tests fail") could only ever prove the option half, because nothing exercised the other. Doc and commit message asserted the strong claim, the tests confirmed the weak one, and the gap between them was exactly the env var. That commit added Decision 2's resolved requirement and the raw hashing in the same change — it was never resolved and later regressed.

## Change

Both keys hash the resolved cap. Versions bump (`search-snapshot-v8`, `provider-records-v3`) because the resolved default differs from the raw zero, so every existing entry re-keys — a bump makes that a rename rather than a silent hash change. Only ever turns hits into misses.

New tests cover the env path for both caches, assert `ENTIRE_GRAPH_MAX_FILES=5` keys identically to `MaxFiles=5` (so the two halves cannot drift onto separate terms again), and assert the lowered-cap direction. Mutation-checked: reverting either key line fails both with "share a key".

An independent pass over the pre-fix code reached the same conclusion and named the same commit; the symmetric framing above came from it.

https://claude.ai/code/session_01PRZZprZMazfmyRkdDPVbDV

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 058efe5e7f56147903f77acb5bbfbd08488ee2de. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/sem/records_cache.go
+++ b/internal/sem/records_cache.go
@@ -19,7 +19,7 @@ import (
 // set of options. Recomputing them on every call is expensive on large repos, so
 // we cache the raw NDJSON bytes keyed on the HEAD tree hash plus everything else
 // that changes the output. This mirrors the search-snapshot cache next door.
-const providerRecordsCacheVersion = "provider-records-v2"
+const providerRecordsCacheVersion = "provider-records-v3"
 
 // cachedProviderRecords is the on-disk envelope for a cached record stream. The
 // key alone is authoritative (sha256 over version+tree+mode+profile+options+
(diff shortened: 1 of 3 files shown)


ashtom:
bugbot run

PR #111, opened 2026-08-17
release: statically link Windows archives

ashtom:
https://entire.io/gh/entireio/entire-graph/trails/71

## What changed

- Append the exact `-static` linker flag for every Windows release target while preserving effective `CGO_LDFLAGS` from the process environment or GOENV.
- Verify each extracted Windows archive with a PATH containing only the Windows system directories.

## Why

The published Graph v0.3.0 Windows executable imports MinGW runtime DLLs that are not present on a clean Windows machine. It exits with `0xC0000135` until those external DLLs are added to PATH. Release archives should be self-contained.

## Validation

- Built the Windows AMD64 archive on an isolated Azure Windows Server 2022 VM with Go 1.26.6 and GCC 15.2.0.
- Confirmed the candidate imports only `KERNEL32.dll` and `msvcrt.dll`.
- Installed the candidate through Entire CLI 0.10.0 and exercised `version`, `init-agents`, idempotence, and full-profile search with no MinGW or GCC directory on PATH.
- Verified GOENV preservation, process-environment precedence, and idempotent `-static` handling.
- Passed shell syntax, workflow YAML, PowerShell parsing, diff checks, and the graph-directed narrow Go test.

Both native Windows release matrix jobs remain the merge gate.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 62c51e033e19e950b5731ec3418870407f1c0352. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,7 +71,6 @@ jobs:
           }
           "CC=$cc" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
           "CXX=$cxx" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          "CGO_LDFLAGS=-static" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
           & $cc --version
       - name: Select artifact version
         id: version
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -125,6 +124,34 @@ jobs:
...
+          $verifyDir = Join-Path $env:RUNNER_TEMP "entire-graph-clean-path-$($env:GOARCH)"
+          Remove-Item -Path $verifyDir -Recurse -Force -ErrorAction SilentlyContinue
+          Expand-Archive -Path $asset -DestinationPath $verifyDir
+
+          $binary = Join-Path $verifyDir "entire-graph.exe"
+          $originalPath = $env:PATH
+          try {
+            $env:PATH = "$env:SystemRoot\System32;$env:SystemRoot"
+            $actualVersion = (& $binary version | Out-String).Trim()
+            if ($LASTEXITCODE -ne 0) {
```
suhaanthayyil, line 146: Dead code: this `if` never runs on the runners this job targets.

`$ErrorActionPreference = "Stop"` combined with pwsh 7.3+'s default `$PSNativeCommandUseErrorActionPreference=$true` means a nonzero exit from `& $binary version` throws immediately, before this line executes. windows-2025/windows-11-arm ship pwsh well past 7.3, so this is live here.

Not a masking bug, the step still fails correctly and `finally` still restores PATH. Just means the descriptive message never shows up in the log, you get a generic .NET exception instead. Low severity, worth a one-line fix (check `$LASTEXITCODE` via `$? `right after the native call, or drop the redundant throw) so the diagnostic actually fires when this step is doing its job.

ashtom: Claude and Codex both think this is correct, so closing here.

PR #112, opened 2026-08-17
fix: explain high-score confidence warnings

ashtom:
https://entire.io/gh/entireio/entire-graph/trails/72

## What changed

- Keep the `below 12` explanation only when the top score is actually below the low-score threshold.
- For high-scoring near ties and wrapper-shaped hits, render the assessment reason instead.
- Add focused coverage for low-score, near-tie, and wrapper cases.

## Why

The renderer always described low confidence as `weak, below 12`, even when confidence was low for a different reason and the displayed score was much higher. A Linux onboarding run exposed the contradiction with a score of 27.5.

## Validation

- Focused low-confidence renderer tests passed.
- The broader confidence test set passed.
- `go test ./internal/cli`, `go test ./...`, and `go vet ./...` passed with the repository-pinned Go toolchain.
- `git diff --check` passed.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 62ee5f6f5f563a4ecf7644159b1daa5ce35b34dd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/cli/search.go
+++ b/internal/cli/search.go
@@ -1299,9 +1299,13 @@ func searchLowConfidenceNotices(response sem.SearchResponse) ([]byte, []byte) {
 	if !assessment.Low {
 		return nil, nil
 	}
+	score := fmt.Sprintf("top score %.1f", assessment.TopScore)
+	if ceiling := sem.LowConfidenceScoreCeiling(); assessment.TopScore < ceiling {
+		score += fmt.Sprintf(" (weak, below %.0f)", ceiling)
+	}
 	full := []byte(fmt.Sprintf(
-		"LOW CONFIDENCE: top score %.1f (weak, below %.0f) and %s. This repo may not contain what you asked for; verify before editing.\n",
-		assessment.TopScore, sem.LowConfidenceScoreCeiling(), assessment.Reason,
+		"LOW CONFIDENCE: %s and %s. This repo may not contain what you asked for; verify before editing.\n",
+		score, assessment.Reason,
 	))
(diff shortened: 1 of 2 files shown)


internal/cli/search.go
```diff
@@ -1299,9 +1299,13 @@ func searchLowConfidenceNotices(response sem.SearchResponse) ([]byte, []byte) {
 	if !assessment.Low {
 		return nil, nil
 	}
+	score := fmt.Sprintf("top score %.1f", assessment.TopScore)
```
suhaanthayyil: This narrows the contradiction, doesn't close it.

Display rounds to `%.1f`, but the "weak, below 12" check compares the raw value. `TopScore=11.96` still renders as `12.0` while still failing the `< 12` check, so you can still get `top score 12.0 (weak, below 12)`.

Reproduced directly: `SearchResponse` with scores `[11.96, 10.0, 9.0]` across 3 files hits the dispersion `Low=true` path and prints exactly that self-contradictory string.

Fix is to compare against the rounded value (or round the threshold check the same way display does), not the raw one.

ashtom: Confirmed and fixed in 8f7c454 — reproduced your exact case first (`[11.96, 10.0, 9.0]` across 3 files printed `top score 12.0 (weak, below 12)`). The renderer now formats the score once with `%.1f`, parses the rendered string back, and gates the suffix on that parsed value, so the suffix follows the score as displayed (this also sidesteps `%.1f` half-to-even vs `math.Round` midpoint disagreements). Added a regression test with your repro values.

PR #114, opened 2026-08-18
Version sem.Result for persisted diff/analyze output

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/74

`graph diff --json` output is about to be persisted into checkpoint metadata by entireio/cli and indexed by entireio/entire-brain (the VCS-keystone work), and unversioned persisted data can't evolve safely.

- `sem.Result` gains `schema_version` (stamped from the existing `SchemaVersion` const, "1.1") and `producer_version` (from the existing ldflags build version, omitempty), set centrally in the one constructor all diff/analyze/checkpoint paths share.
- Append-only: no existing field changed; old readers unaffected.
- Tests: `TestAnalyzeGitRangeSetsSchemaVersion`, `TestDiffJSONIncludesSchemaVersion`; full suite + `mise run check` green.

Both downstream consumers tolerate absence of the new fields (old binaries) — verified against their branches.

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -690,7 +690,7 @@ func runCheckpoint(ctx context.Context, opts Options, args []string) error {
 	if err != nil {
 		return err
 	}
-	return printResult(opts.Stdout, result, flags.JSON)
+	return printResult(opts.Stdout, result, flags.JSON, opts.Version)
 }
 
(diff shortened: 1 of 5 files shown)


suhaanthayyil:
Closing — origin unclear (not from this session; verified the code itself is sound and tests pass, but couldn't confirm which session/client authored it). Reopen if you can confirm it's intentional work you want merged.

PR #128, opened 2026-08-19
fix(gitutil): reject paths the cat-file batch protocol cannot carry

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/88

## The bug

A `git cat-file --batch` request is **one line**. The sink at `internal/gitutil/git.go` writes

```go
fmt.Fprintf(r.stdin, "%s:%s\n", r.rev, path)
```

and then reads exactly one response record. Anything on that line that does not survive the round trip desynchronises the pipe **for the rest of the reader's life**.

Git's own request parser decides what those are: `strbuf_getline` strips the trailing LF, then strips **one** trailing CR. So three shapes are unsafe, and **both halves of the line** are exposed:

| request shape | what git does | blast radius |
|---|---|---|
| path with embedded `\n` | one `ReadFile` call contributes **N+1** request lines, reader consumes **1** record | pipe is permanently one record ahead — **every later path** returns a different file's blob |
| path with trailing `\r` | the CR is eaten before the lookup | that one read silently resolves to a **different, real** object and reports success |
| **revision** with embedded `\n` | the same split, on **every** request the reader makes | the reader is desynchronised from its first read onward |

Neither produces an error, a short read, or anything else a caller can observe.

The guard already existed at **three of the four** call sites — `internal/sem/provider.go`, `internal/sem/search.go`, and `internal/cli/callsite.go` (added in #99, with a dedicated regression test at `internal/cli/head_source_provenance_test.go`). The fourth, `internal/sem/dependents.go`, did a bare `readFile = batch.ReadFile`, so the dependents scan read every candidate file through the corruptible reader. `internal/gitutil/git.go` also had no guard of its own, which is why a sweep could miss a call site at all.

And each of the three existing guards was a **separate copy of the rule that only knew about newlines** — so the CR shape reached the reader at all four sites, and the revision shape at all four as well.

### 1. A path the line cannot carry

Repo shape — one tracked file whose **name** contains a newline, sorting ahead of the real dependents:

```
z_defines.py        def Foo(token): ...
a_ev<LF>xil.py      def uses_foo_from_a_weird_path(token): return Foo(token)
c1.py .. c5.py      def check1(token): return Foo(token)   (five genuine dependents)
```

```sh
entire graph commit HEAD --json
```

`dependents_count` for `Foo` comes back **5 instead of 6**, exit 0, no warning, no `partial_failure` — and every dependent is attributed to the *next* file in scan order. Verbatim, from the test run against unmodified code:

```
dependents count = 5, want 6:
map[string]struct {}{
  "c2.py#function:check1":{},   <- c1.py's function, recorded against c2.py
  "c3.py#function:check2":{},
  "c4.py#function:check3":{},
  "c5.py#function:check4":{},
  "z_defines.py#function:check5":{},  <- c5.py's function, recorded against the DEFINING file
}
```

### 2. A **revision** the line cannot carry

`parseDiffFlags` in `internal/cli/root.go` takes `--head` and hands the value to `AnalyzeGitRangeWithOptions` → `addDependentCounts` → `NewBatchFileReader` **unchanged** — there is no `rev-parse` on that path. And git accepts revisions containing a newline: `HEAD^{/subject\nbody}` is a commit-message search against a multi-line message. Probed against git 2.54.0:

```
$ REV=$(printf 'HEAD^{/multi-line subject\nsecond line}')
$ git rev-parse "$REV"
ee0569c59891e0fef35c543cfd5c733d5c9d57a2       <- resolves
$ git rev-parse HEAD
ee0569c59891e0fef35c543cfd5c733d5c9d57a2       <- to the same commit

$ printf '%s:a.go\n' "$REV" | git cat-file --batch
HEAD^{/multi-line subject missing
second line}:a.go missing                      <- ONE request, TWO records
```

Two records per call, one consumed. Against the previous commit of this branch, `buildReferenceIndex` under that revision reports `dependents_count` **0 instead of 3** — a revision git resolves correctly, answered with a confidently empty result:

```
dependents count = 0, want 3: map[string]struct {}{}
```

### 3. A **CR-suffixed** path is a real, indexable path

Not hypothetical for these readers: `languageForPath` matches `dockerfile.` and `makefile.` by **prefix**, so `Dockerfile.dev\r` is a supported path the provider will list, read and index. Probed against git 2.54.0 with both names committed:

```
$ git ls-tree -r --name-only HEAD
Dockerfile.dev
"Dockerfile.dev\r"

$ printf 'HEAD:Dockerfile.dev\r\n' | git cat-file --batch
972cadfa271830cee0e1f55b67a0b2cb44757dcd blob 31
FROM alpine
RUN echo the-DECOY            <- the CR-LESS file's blob

$ printf 'HEAD:Dockerfile.dev\n'   | git cat-file --batch
972cadfa271830cee0e1f55b67a0b2cb44757dcd ...    <- same oid, so the CR request resolved the other file

$ git show "$(printf 'HEAD:Dockerfile.dev\r')"
FROM alpine
RUN echo the-cr-suffixed-file  <- argv form answers the right file
```

So before the sink guard, the three sibling readers were served the **wrong blob** for this path; with only the sink guard, they **dropped** it. Neither is the file.

## The fix

Guard the **sink**, so no fifth caller can miss it, and make every caller route on the sink's answer instead of on its own copy of the rule.

- `internal/gitutil/git.go`
  - `checkBatchableRequest(rev, path)` checks the bytes that are actually written — both halves of the line — *before* writing any request, and returns the sentinel `ErrNonBatchablePath`. Because nothing has been written, the stream is left exactly as it was and the reader stays usable for every subsequent path. It is an error, not a "missing file": a path that cannot be *requested* is not a path that is *absent*, and reporting it as absent is the silent drop `docs/trust-and-security.md:75` rules out.
  - `NewBatchFileReader` **resolves** a newline-bearing revision to an object id rather than refusing it — refusing would turn a revision git accepts into a hard failure, and the revision is only ever used as the request prefix, so an object id names the same commit in a form the line protocol can carry. An ordinary revision skips the resolution entirely, so the fast path costs nothing. A CR *inside* the revision is deliberately left alone: it can only reach the end of the line when the path is empty, and `checkBatchableRequest` covers that case.
  - New `ShowFileBounded`, which is `ShowFileLimited` plus the **size of the blob it refused**. `ShowFileLimited` now delegates to it. A caller cannot reconstruct that number from the returned content — the point of the refusal is that the content was never materialized.
- `internal/sem/dependents.go` — falls back on `ErrNonBatchablePath`, so the file is still **read and still counted**. The fallback is `ShowFileBounded` at the same `defaultMaxParseBytes` the batch reader is capped at, and the refused size is recorded so the `E_FILE_TOO_LARGE` warning keeps its byte count.
- `internal/sem/provider.go`, `internal/sem/search.go`, `internal/cli/callsite.go` — the `strings.Contains(path, "\n")` pre-check is replaced by `errors.Is(err, gitutil.ErrNonBatchablePath)`. One place now decides which requests the protocol cannot carry; the callers obey it. `search.go` and `callsite.go` pass their own ceiling down to the fallback so it refuses the same blobs the batch reader refuses.

### Why `--batch -Z` was not used

`git cat-file --batch -Z` makes both the request and response records NUL-delimited and handles all three shapes correctly (verified on 2.54.0). It was rejected because it needs **Git >= 2.42**: it would have to be version-detected at runtime and backed by this same guard on older Git anyway, leaving two response parsers on the security-critical path to keep in agreement. A guard at the sink is total, and the refused paths are exotic enough that one extra `git show` subprocess is the right price.

## Why it cannot regress existing behaviour

- **The guard refuses only what git actually mangles.** A CR anywhere but the *end* of the line round-trips byte for byte, so it is still served by the batch process. Verified against git 2.54.0: `HEAD:mid<CR>cr.go` returns `MIDCR-CONTENT`, its own blob. The gitutil test asserts this explicitly.
- **A newline-bearing revision keeps working**, and now works *correctly*, rather than being turned into an error.
- **A CR-suffixed path is now read, not dropped and not misresolved** — strictly better than either previous state.
- **The check runs before the write**, so a refusal cannot itself perturb the protocol.
- **`provider.go`'s fallback keeps the read it has had since #43.** Bounding it there would refuse an oversized newline-named blob without an oversize-registry entry to record it — the file would vanish from the graph instead of being reported. That is a pre-existing bound this PR does not tighten, rather than one it loosens.
- **`dependents.go` still reports what it skips**: the size-refused fallback records the size, so `E_FILE_TOO_LARGE` carries the same byte count the unbounded read produced. On the *non-prefiltered* full-tree scan, where relevance is decided from content nobody read, such a file is skipped as quietly as any other blob never shown to be a candidate — the same rule the batch reader's oversize path already follows.

## Tests

All four were run against the code they fix and **fail there at runtime**, not merely fail to compile.

**`internal/gitutil/git_test.go` — `TestBatchFileReaderRefusesPathsTheProtocolCannotCarry`**
Both path shapes are refused, and — the part that matters — ordinary paths read *after* a refused one still get their **own** blob. Failing-first output against the unguarded sink:

```
ReadFile("b_ev\nxil.go") err = <nil>, want ErrNonBatchablePath
ReadFile("endcr.go\r")   err = <nil>, want ErrNonBatchablePath
batch read "a.go" = "package endcr\n// the DECOY the stripped request resolves to\n",
    want "package a\nfunc A() {}\n" (response stream is out of step)
```

**`internal/gitutil/git_test.go` — `TestBatchFileReaderReadsThroughARevisionTheProtocolCannotCarry`**
Asserts the revision is one `rev-parse` resolves, then that three successive reads each get their own blob. Failing-first, with the revision handling removed:

```
ReadFile("a.go") through a newline-bearing revision: not found
```

**`internal/sem/dependents_test.go`**
- `TestBuildReferenceIndexCountsDependentsPastANewlineNamedFile` asserts the **exact dependent key set** past a newline-named file, so an off-by-one response stream cannot pass by producing the right count against the wrong files. Failing-first output is the misattributed map quoted above.
- `TestBuildReferenceIndexCountsDependentsUnderARevisionTheBatchProtocolCannotCarry` — failing-first: `dependents count = 0, want 3`.
- `TestBuildReferenceIndexBoundsTheNonBatchableFallbackRead` puts a 24 MiB blob at a newline-named path and measures `runtime.MemStats.TotalAlloc` across the scan, because the warning is identical either way and the allocation is the only observable. It also asserts the warning survives with its byte count. Failing-first against an unbounded `ShowFile` fallback:

```
scan allocated 92414680 bytes, want < 25165810: the 25165810-byte blob at a
non-batchable path was materialized instead of refused by size
```

**`internal/sem/nonbatchable_path_readers_test.go` — `TestHeadReadersRouteCRSuffixedPathsToTheArgvSafeReader`** (new file)
Drives `openSource` and `openSearchContentReader` at a committed revision holding both `Dockerfile.dev` and `Dockerfile.dev\r` with different contents, and demands the CR path's **own** content. Failing-first against the newline-only pre-check:

```
openSource read("Dockerfile.dev\r") = not found; a tracked file was dropped
search head read("Dockerfile.dev\r") = not found; a tracked file was dropped
```

**`internal/cli/head_source_provenance_test.go`** gains the same CR-suffixed case for the call-site reader, alongside its existing newline case.

Every fixture using a newline or a CR in a filename is skipped under `runtime.GOOS == "windows"`, where the input is unrepresentable.

## Gate

```
gofmt -s -l .     (empty)
go vet ./...      clean
go build ./...    clean
go test ./internal/gitutil    42 passed
go test ./internal/sem      2139 passed
go test ./internal/cli       536 passed
```

Schema contract untouched (no json tag added, removed or renamed). No network call added.

## Re-opens

Extends the sweep started in #99 (`fix: keep --head source output on the committed tree`), which added the third call-site guard, and #43, which added the first two. This one moves the check to the sink and makes the callers route on it, so the sweep cannot be incomplete again — and covers the revision half of the request line, which no previous pass looked at.

## Review follow-up: the fallbacks were weaker than the reader they replace

Routing a non-batchable path to the argv reader fixed *which object* is read. Two review findings showed the fallbacks were still not equivalent to the batch reader on the other two axes — memory bound, and explaining a refusal.

**`internal/sem/provider.go` — the fallback had no cap at all.** `openSource` promises a file above `maxReadBytes` is never materialized and is instead recorded in the oversize registry from a streamed digest. The fallback called plain `ShowFile`, so one CR-suffixed tracked file set the snapshot's memory to the size of the largest blob in the revision, and was then dropped with nothing in the registry to explain it. Failing first, against the previous commit, with a 4 KiB ceiling and a 4.4 MB blob at `Dockerfile.dev\r`:

```
read("Dockerfile.dev\r") returned 4400012 bytes with a 4096-byte ceiling: the non-batchable fallback is unbounded
read allocated 21221592 bytes, want < 4400012: the blob was materialized instead of refused by size
oversize("Dockerfile.dev\r") = not recorded; the refused file has no explanation
```

**`internal/sem/dependents.go` — the refusal was silent.** The fallback did refuse by size, but nothing then read the blob, so the oversize scanner never marked it relevant. On the prefilter-failed full-tree scan the warning gate is `prefiltered || oversizeMatched[path]`, so the file was skipped with no `E_FILE_TOO_LARGE` and an undercounted `dependents_count`. `TestBuildReferenceIndexBoundsTheNonBatchableFallbackRead` did not catch it because it lets the prefilter succeed, and with `prefiltered == true` the gate is open regardless. Failing first, with the prefilter forced to fail:

```
no E_FILE_TOO_LARGE warning for "a_ev\nxil.py" on the prefilter-failed path; got
[{W_DEPENDENTS_PREFILTER_FAILED ...}]
```

Both now go through `gitutil.ShowFileDigested`: an argv-based bounded read that, on refusal, streams the blob into a digest (size, content hash, line count) and an optional chunk scanner — the same thing `BatchFileReader` does for the paths it *can* carry. It reads git's stdout to EOF rather than stopping git early, per the Windows teardown deadlock already documented on `ShowFileLimited`. The dependents scan hands it the *same* scanner closure the batch reader gets, so relevance is decided by one rule for both paths rather than two.

The relevance scanner now matches on bytes through a reused window buffer. The previous string form allocated a copy of every chunk, so the cumulative cost of *deciding* about a blob scaled with the blob — which would have broken the existing memory assertion the moment the fallback started scanning.

Both new tests assert the refusal, the byte count in the warning/registry, and that the scan still allocates less than the blob. A second oversized non-batchable blob containing no changed name pins the other direction: it must stay silent, so the fix cannot be "warn about everything the fallback refused".

`gofmt -s -l .` empty · `go vet ./...` clean · `go build ./...` · `go test ./internal/gitutil/ ./internal/sem/ ./internal/cli/` — 2719 tests green.

## Review follow-up: the fallback read whatever git rendered, not a blob

A third finding closed the last axis on which the argv fallback was not equivalent to the batch reader: the **object type**.

`git ls-tree -r` lists a **gitlink** — a tree entry of mode `160000`, the shape a submodule takes — exactly like a file, and `Dockerfile.dev\r` is a *supported* path (prefix language detection) the line protocol cannot carry. So a gitlink can be both listed and routed to the fallback. `BatchFileReader.ReadFile` reads the type out of its response header and reports a non-blob as absent; `ShowFile` ran `git show`, which **renders** whatever the spec resolves to. Probed against git 2.54.0, with a gitlink at `Dockerfile.dev\r` pointing at a commit in the same repository:

```
$ git cat-file -t 'HEAD^{tree}:Dockerfile.dev<CR>'
commit
$ git show 'HEAD^{tree}:Dockerfile.dev<CR>'
commit bbe7fa2b9189dee4aed13d29f49dbc88f401a47d
Author: a <a@b>
Date:   Wed Aug 19 17:06:09 2026 -0400

    multi-line subject
    second line of the message

diff --git a/Dockerfile.dev b/Dockerfile.dev
...
```

That render came back as `ok == true`, so the provider hashed, parsed and indexed commit output — author line, message, diff — as Dockerfile content, with no partial failure to show for it.

`ShowFile` now runs **`git cat-file blob`**, which resolves the same spec and refuses anything that is not a blob. Verified against git 2.54.0:

- **Missing-path classification is unchanged.** Both diagnostics `isMissingPathDiagnostic` matches are byte-identical between the two commands: `fatal: path 'nope.go' does not exist in 'HEAD^{tree}'` and `fatal: path 'ondisk.go' exists on disk, but not in 'HEAD^{tree}'`.
- **Blob content is unchanged**, including an empty blob and a symlink entry (mode `120000`), whose blob is its target string under both commands.
- **A non-blob is reported absent, matching the batch reader**, so which reader a path is routed to cannot change the answer. Making it an error instead would have been a new asymmetry: a gitlink at an ordinary path is skipped silently by the batch reader, and one at a CR-suffixed path would have become a hard failure.
- **The type is asked for, not read out of git's wording.** `git cat-file blob` says `bad file` for a gitlink, a tree and a corrupt object alike, and only the first two are absences, so the fallback consults `git cat-file -t`. That runs **only on a read that has already failed**, so the ordinary path is still one process.

### Tests

**`internal/gitutil/git_test.go` — `TestShowFileRefusesNonBlobEntries`.** Failing-first at runtime against `git show`:

```
ShowFile(gitlink) = "commit 68dbda986cddb019dee8d5bfd58cdc655ab210f4\nAuthor: Entire Graph Test <graph@example.com>\n...
    subject line\n    \n    body rendered as if it were file content\n\ndiff --git a/a.go b/a.go\n...",
ok=true; want not found (rendered commit output would be parsed as source)
```

It also asserts parity with `BatchFileReader.ReadFile` on the same gitlink, and — pinning non-over-suppression — that an ordinary blob is still returned byte for byte and a genuinely absent path is still classified absent rather than as a failure.

**`internal/sem/nonbatchable_path_readers_test.go` — `TestHeadReadersRefuseANonBlobAtANonBatchablePath`.** The end-to-end half: a gitlink at `Dockerfile.dev\r`, driven through `openSource` and `openSearchContentReader`. Failing-first at runtime:

```
openSource  read("Dockerfile.dev\r") = "commit 6cedf122...\nAuthor: ...\n    RUN echo body-rendered-as-content\n...", ok=true
search head read("Dockerfile.dev\r") = "commit 6cedf122...\nAuthor: ...\n    RUN echo body-rendered-as-content\n...", ok=true
```

`TestHeadReadersRouteCRSuffixedPathsToTheArgvSafeReader` — a real blob at the same CR-suffixed path — passes before and after, so the type constraint did not eat the legitimate input the earlier commits restored. The gitlink fixture needs no symlink, device or exotic filename in the *worktree* (`git update-index --cacheinfo` writes it straight to the index), so only the CR-path variant carries the existing `runtime.GOOS == "windows"` skip.

```
gofmt -s -l .     (empty)
go vet ./...      clean
go build ./...    clean
go test ./...     2813 passed in 21 packages
```


ashtom:

- **8 commits ahead, 219 commits behind** current `main`.
- Test-merging onto `main` **conflicts in 8 file(s)**:

```
internal/cli/callsite.go
internal/cli/head_source_provenance_test.go
internal/gitutil/git.go
internal/gitutil/git_test.go
internal/sem/dependents.go
internal/sem/dependents_test.go
internal/sem/provider.go
internal/sem/search.go
```

### What this needs

### Entire trail findings

### Verified vs. assumed

suhaanthayyil:
I did the merge check that was asked for, but stopped before pushing one: **`main` closed all three of this branch's security properties independently, and this PR's own tests prove it.** Recommending close as obsolete rather than landing a merge whose net production diff is nil.

## What closed it

`5ebafdbe` ("fix: harden Git path plumbing") landed on `main` and touched the same files this branch does — `internal/gitutil/git.go`, `internal/sem/{provider,dependents,analyze,search}.go`, `internal/cli/callsite.go`. It took a different route to the same guarantees.

**The namesake guard**, `internal/gitutil/git.go:1262`:

```go
func IsBatchPathSafe(path string) bool {
	return !strings.Contains(path, "\n") && !strings.HasSuffix(path, "\r")
}
```

The trailing-CR case — the one that made this branch more than a newline check — is there. And the revision is carried through it, `git.go:1269`:

```go
func (r *BatchFileReader) IsPathSafe(path string) bool {
	return IsBatchPathSafe(r.rev + ":" + r.pathPrefix + path)
}
```

**The bounded, type-checked fallback** is `readFileLimited`, `git.go:875-932`: `validateLimitedFilePath` before anything runs, then a blob probe that returns `LimitedFileNonBlob` for an identified non-blob and `LimitedFileOversize` for a blob over the cap *before* materializing it. Its own comment names the gitlink case this branch names.

**Non-blob diff entries are surfaced, not dropped**, `internal/sem/dependents.go:427-429`:

```go
detail := "the tree entry identifies a blob, but its Git content was unavailable during dependent analysis"
if status == gitutil.LimitedFileNonBlob {
	detail = "the tree entry was not a blob when dependent content was read"
}
```

## Proof, not inference

I ran this branch's own tests against `bb41f223`, unmodified except for one arity fix unrelated to the security property (`openSearchContentReader` gained a `maxParseBytes` parameter on `main`).

`internal/sem/nonbatchable_path_readers_test.go`, copied across verbatim:

```
--- PASS: TestHeadReadersRouteCRSuffixedPathsToTheArgvSafeReader (0.10s)
    --- PASS: /provider_committed_source
    --- PASS: /search_head_content
--- PASS: TestOpenSourceBoundsTheNonBatchableFallbackRead (0.10s)
--- PASS: TestHeadReadersRefuseANonBlobAtANonBatchablePath (0.10s)
    --- PASS: /provider_committed_source
    --- PASS: /search_head_content
ok  github.com/entireio/entire-graph/internal/sem  0.911s
```

And the decoy fixture this branch added to `head_source_provenance_test.go` — the one where both `Dockerfile.dev` and `Dockerfile.dev\r` are committed with different contents, so a stripped request is indistinguishable from a correct one unless it returns the wrong bytes:

```
read("Dockerfile.dev\r") ok=true content="committed-cr-suffixed-path|"
--- PASS: TestPR128DecoyCarriageReturnPath (0.07s)
```

`main` answers with the CR-suffixed blob, not the decoy. `TestBatchFileReaderRefusesPathsTheProtocolCannotCarry` also passes on `main` — both subtests, plus its stream-integrity assertion that ordinary reads after a refused path are not shifted by one.

## One residual, and it is not this PR's headline

`TestShowFileRefusesNonBlobEntries` is the single test of this branch that **fails** on `main`:

```
ShowFile(gitlink) = "commit 571c75a7…\n\ndiff --git a/a.go b/a.go\n…", ok=true;
want not found (rendered commit output would be parsed as source)
```

`gitutil.ShowFile` still renders a gitlink's commit — subject, body and patch — and reports it as file content. It is not reachable today: `ShowFile` has exactly one caller in the tree,

```
internal/gitutil/git.go:917:  content, ok, err = ShowFile(ctx, repo, rev, path)
```

inside `readFileLimited`, and only on an *inconclusive* probe. Every guarded reader refuses the gitlink — verified directly on `bb41f223`:

```
ShowFileLimited(gitlink)            ok=false  content=""
ReadFileLimited(gitlink)            status=LimitedFileNonBlob
LimitedFileReader.ReadFile(gitlink) status=LimitedFileNonBlob
```

So it is a trap on an exported function rather than a live hole, and `main`'s design deliberately puts the type check in the probe rather than paying for a second git invocation inside `ShowFile`. Worth a two-line follow-up if that export is meant to be safe standalone; not a reason to land 219 commits of merge.

## Why I am not pushing a merge

Test-merging conflicts in eight files — `internal/cli/{callsite.go,head_source_provenance_test.go}`, `internal/gitutil/{git.go,git_test.go}`, `internal/sem/{dependents.go,dependents_test.go,provider.go,search.go}` — and in every one of them the conflict is this branch's version of a fix against `main`'s. Resolving them means taking `main`'s side throughout, which produces a merge commit with no net production change. The two things that would genuinely survive are a stronger decoy fixture (which `main` already passes) and the `ErrNonBatchablePath` sentinel in place of `main`'s plain error — a readability argument, not a security one, though it is the argument that was right the first time: the duplicated rule is exactly what let the CR case through.

Recommend: close, and open the `ShowFile` gitlink note separately if it is wanted.

Verified by execution on `bb41f223`, darwin/arm64, Go 1.25.7. Not verified: Windows and Linux behaviour.

PR #132, opened 2026-08-19
fix(sem): never index a git directory, whatever its name or depth

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/92

## Review round 4 — three over-corrections, narrowed back to git

Review of head `8c3a3ae` raised three findings, all of them the price of the
guards above rather than the original leak: one cost regression and two places
where the scanner now excludes what git accepts. Each was reproduced at run time
on `8c3a3ae` first, and each want comes from **git 2.54.0 run as an oracle**.

### 1. `commondir` — an over-strict read rejected what git follows

`gitCommonDir` used `os.Lstat` and required a regular file, so a symlinked
`commondir` was refused. Git reads the file through the ordinary stdio path and
follows the link:

```
$ mkdir adm && printf 'ref: refs/heads/main\n' > adm/HEAD
$ printf '%s\n' "$PWD/main/.git" > realcommondir
$ ln -s ../realcommondir adm/commondir
$ git --git-dir=adm rev-parse --git-dir
adm
$ git --git-dir=adm rev-parse --git-common-dir
/…/main/.git
```

A linked worktree's administrative git directory is not named `.git` and no
pointer names it, so the structural rule is the only one in reach — refusing its
`commondir` put its credentialed config into a ranked snippet. Failing first on
`8c3a3ae`:

```
--- FAIL: TestLooksLikeGitDirAcceptsSymlinkedCommondir
    looksLikeGitDir = false for a git directory whose commondir is a symlink, want true
--- FAIL: TestSearchRepositoryNeverIndexesLinkedWorktreeGitDirWithSymlinkedCommondir
    search returned a path inside the git directory "state/.wt-git": "state/.wt-git/config"
    search snippet for "state/.wt-git/config" leaked the planted remote credential
```

The read is `os.Stat` now, so the link is followed. ~~A `commondir` that is a
**directory** still fails the regular-file test, as it fails git's own read.~~
**Corrected in round 8:** a `commondir` git refuses to read no longer means one
thing to both rules — see below.

### 2. A gitfile's target — excluded without asking whether git accepts it

Rule 2 excluded any in-repository path named by a syntactically valid `.git`
file. Git parses the pointer and then asks `is_git_directory()` of the target:

```
$ printf 'gitdir: ../src\n' > nested/.git      # src is an ordinary Go package
$ git -C nested rev-parse --git-dir
fatal: not a git repository: (null)
[...]

PR #134, opened 2026-08-19
fix(gitutil): stop honoring executable git config from the indexed repository

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/94

## The bug

Git reads configuration from the repository it is pointed at, and some git configuration values are **commands git executes**. Every git subprocess `internal/gitutil` spawned inherited them, so indexing a repository was enough to run code that repository chose.

**Vector 1 — `core.fsmonitor`.** Git runs this value through `sh -c` whenever it refreshes the index. All four index-refreshing call sites were affected: `ListWorktreeFiles` (`internal/gitutil/git.go:115`), `ListIndexFiles` (`:92`), `ListIgnoredWorktreeFiles` (`:128`), and both working-tree greps (`:272`, `:334`).

Reproduced end to end against a binary built from `origin/main` (cceb4de):

```
$ git config core.fsmonitor "touch $S/PWNED"
$ ./eg-main search --repo . --query "parse the timeout"
  PWNED CREATED  -> RCE reproduces on main
$ ./eg-fixed search --repo . --query "parse the timeout"
  no marker -> BLOCKED
```

The fixed binary still returns the correct result for that query (rank 1, `src/app.py:1`, `parse_timeout`).

**Vector 2 — `gpg.program`, found while auditing the rest of the argv.** `gpg.program` is executed to verify a commit signature, and `log.showSignature` — also repository config — is the switch that makes `git log` reach it. A forged `gpgsig` header on a commit is enough to trigger verification, no real key involved, which exposed `FindCommitWithCheckpoint` (`:62`) and `FileCochanges` (`:443`) the same way. Reproduced with a commit object rewritten to carry a bogus signature header; the test builds exactly that fixture.

## Threat model — please read this before rating severity

**`git clone` does not transfer `.git/config`.** Verified:

```
$ git -C src config core.fsmonitor "touch $S/MARKER"
$ git clone -q "$S/src" "$S/clone"
$ git -C "$S/clone" config --local --get core.fsmonitor
<unset>
$ git -C "$S/clone" ls-files --cached --others --exclude-standard
no marker in clone -> config NOT transferred
```

So this is **not** a clone-and-run exploit. It requires a tarball-delivered repository, a restored CI cache, or write access someone already has. The value here is defence in depth for a tool that is explicitly designed to run on repositories the user has not audited — not an emergency.

## The fix

`newGitCmd` (`:888`) prepends `-c core.fsmonitor=` and `-c log.showSignature=false` to every git argv. The two grep helpers now route through it instead of building their own `exec.Cmd`, so they gain the hardening while **keeping the caller's locale** that `git grep -i` needs for non-ASCII case folding. Overrides are prepended inside the helper rather than folded into the args callers assemble, so error messages still quote the command the caller asked for.

`GIT_TERMINAL_PROMPT=0` is also set. Stated honestly: it cannot change the behavior of anything this package runs today, because none of these commands contacts a remote. It is there so a future caller that does cannot inherit an interactive stall.

## What I checked before shipping, so legitimate use does not regress

**`GIT_CONFIG_NOSYSTEM` is deliberately NOT set.** It blocks neither vector — the payload lives in the repository's own config, which `GIT_CONFIG_NOSYSTEM` does not touch — and it **does** break a documented behavior. `ListWorktreeFiles`' doc comment promises it honors `core.excludesFile` "(global and system)". Verified that `GIT_CONFIG_NOSYSTEM` breaks the system half:

```
# system config sets core.excludesFile = <file containing "secret.txt">
$ GIT_CONFIG_SYSTEM=$S/sysgitconfig git ls-files --others --exclude-standard
a.txt
$ GIT_CONFIG_SYSTEM=$S/sysgitconfig GIT_CONFIG_NOSYSTEM=1 git ls-files --others --exclude-standard
a.txt
secret.txt          <-- exclude silently dropped
$ GIT_CONFIG_SYSTEM=$S/sysgitconfig git -c core.fsmonitor= ls-files --others --exclude-standard
a.txt               <-- the narrow fix leaves excludes intact
```

That is the narrower fix the task asked for if this turned out to be the case. Repo-level `core.excludesFile` is already covered by the existing `TestListWorktreeFilesAppliesEveryExcludeSource`, which still passes.

**Credential helper:** not load-bearing. No command in this package contacts a remote, so credential/ssh/proxy config is unreachable. The provider is no-egress by contract.

**Nothing else in the argv reaches an executed config value.** Probed against the same fixture: `diff.external`, `diff.<driver>.textconv` and `core.pager` never fired for `git diff --name-status`, `git log --name-only`, `git grep`, `git show <tree>:<path>`, `git cat-file --batch`, or `git ls-files`. `git rev-parse`, `git show`, `git log`, `git diff`, `git config` and `git cat-file` also do not trigger `core.fsmonitor`; only `ls-files` and working-tree `grep` do.

## Failing-first proof

The new test compiles against `origin/main` and **fails at run time**, not at build time. Production file restored to `origin/main` (`git diff --stat origin/main -- internal/gitutil/git.go` empty), new test kept:

```
=== RUN   TestGitSubprocessesDisableRepositoryFsmonitorHook
    git_hardening_test.go:86: ListWorktreeFiles executed the repository's core.fsmonitor command
    git_hardening_test.go:96: ListIndexFiles executed the repository's core.fsmonitor command
    git_hardening_test.go:103: ListIgnoredWorktreeFiles executed the repository's core.fsmonitor command
    git_hardening_test.go:113: GrepIndexMatches executed the repository's core.fsmonitor command
    git_hardening_test.go:124: GrepFixedStringPaths executed the repository's core.fsmonitor command
--- FAIL: TestGitSubprocessesDisableRepositoryFsmonitorHook (0.26s)
=== RUN   TestGitLogSubprocessesDisableRepositorySignatureVerification
    git_hardening_test.go:183: FindCommitWithCheckpoint executed the repository's gpg.program
    git_hardening_test.go:193: FileCochanges executed the repository's gpg.program
--- FAIL: TestGitLogSubprocessesDisableRepositorySignatureVerification (0.30s)
FAIL	github.com/entireio/entire-graph/internal/gitutil	0.756s
```

Each test carries a **control** that runs raw `git` first and asserts the marker *does* fire, then skips with a reason if it does not — so a git build that ignored these keys cannot make the assertions pass vacuously.

## Windows

Both tests skip on `runtime.GOOS == "windows"` with the reason inline: git invokes the `core.fsmonitor` hook through `sh -c` and the fake `gpg.program` is a `#!`-script, so the fixture needs a POSIX shell and a POSIX exec. The hardening itself is argv-level and identical on every platform. The marker script locates its own output file via `$0` rather than an embedded path, so no temp directory ever has to survive being quoted into a shell word.

## Gate

- `gofmt -l .` — empty
- `go vet ./...` — clean
- `go build ./...` — clean
- `go test ./internal/gitutil/` — ok (40 tests, also green under `-race`)
- `go test ./internal/sem/ ./internal/cli/` — ok (51.0s / 5.6s), the two biggest consumers of this package

Schema untouched. No network call added.

## Not in scope, flagged for another owner

`cmd/graph-bench/main.go:326` spawns `git` with the same unhardened pattern. It is a benchmark harness, not the graph verb path, and outside the files this task named — leaving it.


ashtom:
Closing as superseded by #132.

After #132 merged, current main at e31e2553cef565b9c5021177eafad6c2d9adc345 became a strict production superset of this PR. It retains the fsmonitor and signature-verification protections while also adding lazy-fetch and transport blocking, external Git-config and attribute isolation, inherited Git-environment sanitization, replacement-ref and optional-lock protections, machine-safe grep flags, and bounded subprocess waiting. Main also already bypasses persistent worktree snapshot caching for the stronger atomicity and TOCTOU reasons.

The live #134 head, 24bf7643d969bd8f03b3b1bbf3fe966fdc009090, is now 160 commits behind main and has five merge conflicts. Rebasing its production implementation would risk replacing the newer protections with weaker behavior.

The useful remaining delta is regression coverage for:

- repository core.fsmonitor commands
- repository signature-verification programs
- clean and process filters during worktree cache probing

Those tests were overlaid unchanged on post-#132 main and passed, including focused race runs and the full repository race suite. They should be carried forward in a small test-only PR branched from current main, with fresh CI and Entire approval. No production code from #134 should be carried forward.

PR #135, opened 2026-08-19
fix(termsafe): escape C1 controls in the machine-format encoders

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/95

## The bug

`internal/termsafe/termsafe.go:35` documented JSON and NDJSON as deliberately unwrapped:

> JSON and NDJSON output is deliberately NOT wrapped: encoding/json already escapes control characters inside strings, so no raw byte reaches that stream

That premise is true of C0 and **false of C1**. `encoding/json` escapes below U+0020 and folds invalid UTF-8 to U+FFFD, but U+0080-U+009F encode to a *well-formed* two-byte sequence it copies through verbatim — so the C1 controls are the one shape of a control that survives a machine format intact (CWE-150).

Verified against Go 1.26.5:

```
1 marshal bytes: 7b 22 70 22 3a 22 65 76 69 6c c2 9d 35 32 ... c2 9c 66 69 6c 65 2e 67 6f 22 7d
1 contains raw 0xc2 0x9d: true             <- encoding/json does NOT escape U+009D
2 marshal invalid utf8: 7b 22 70 22 3a 22 61 5c 75 66 66 66 64 62 22 7d   <- {"p":"a\ufffdb"}: it DOES fold invalid UTF-8
4 decoded: 61 c2 9d 62                     <- \^] round-trips to the same code point
```

Reaching it needs no flag. `parseSearchFlags` at `internal/cli/search.go:1737` defaults `--format` to `json`, so a repository holding U+009D (OSC) and U+009C (ST) emits the pair that brackets an OSC 52 clipboard write from a bare `entire graph search`.

Reproduced on `main` @ `cceb4de` against a repository whose filename and function body carry those code points — raw C1 bytes reaching stdout:

| invocation | raw C1 bytes |
|---|---|
| `search --query merge` *(no `--format`)* | 6 |
| `search --format json` | 6 |
| `search --format ndjson` | 6 |
| `snapshot --format ndjson` | 20 |
| `snapshot --format compact-ndjson` | 6 |
| `neighbors --format json` | 8 |
| `impact --format json` | 10 |

```
$ eg-main search --repo "$HOSTILE" --query merge | grep -ao 'file_path[^,]*' | sed -n 2p | od -c
0000000    f   i   l   e   _   p   a   t   h   "   :   "   e   v   i   l
0000020  302 235   5   2   ;   c   ;   a   G   V   s   b   G   8   = 302
0000040  234   f   i   l   e   .   g   o   "  \n
```

`302 235` is 0xc2 0x9d (U+009D, OSC) and `302 234` is 0xc2 0x9c (U+009C, ST), straight out of `git`'s pathname and onto stdout.

The rule itself was already in the package — `escapedAt` at `internal/termsafe/termsafe.go:330` escapes both the two-byte form and the stray byte. Only the text renderers were wired to it. The machine-format regression test added in #96 checks `strings.ContainsRune(stdout, 0x1b)` (`internal/cli/termsafe_verbs_test.go:141`) and then positively asserts the raw pathname survives, so it passes on a stream full of raw C1.

## The fix

`termsafe.JSONWriter` rewrites U+0080-U+009F to `\u00XX` and nothing else; the machine-format encoder sinks are wrapped with it. Wrapping the sink rather than the values is the same choice `termsafe.Writer` already made, and for the same reason: a rule enforced at each of a dozen `json.NewEncoder` sites is a rule the next verb forgets.

`layout` becomes a three-valued type so `jsonLayout` can reuse the existing UTF-8 scan. That scan must step **over** a valid rune rather than inspect its interior: 0x97 is the middle byte of U+65E5, and a per-byte rewrite would corrupt every CJK identifier in the stream.

**`internal/cli/root.go:382` needed the same wrap and would not have got it from the encoder change.** A warm provider-records cache is answered by replaying the stored bytes straight to stdout, bypassing `encodeRecord` entirely. Demonstrated with two builds sharing one cache dir:

```
main  writes the cache, then replays it     rawC1=20  escaped=0
fixed replaying that same poisoned cache    rawC1=0   escaped=20
```

An entry written by any build older than this rule keeps streaming the raw control on every hit until the provider version in its cache key changes.

## Failing-first proof

Production files pristine from `origin/main` (`git status --porcelain` showed only the untracked test file), new test kept — it **fails to run**, not merely to build:

```
$ go test ./internal/cli -run TestMachineFormatsEscapeC1Controls -count=1
Go test: 0 passed, 8 failed in 1 packages

  [FAIL] TestMachineFormatsEscapeC1Controls/search_--query_merge
     termsafe_machine_c1_test.go:99: raw C1 control at byte 722 of the machine format:
     "nals\":[\"body\",\"exact-symbol\",\"symbol-name\",\"all-query-terms\",\"graph:callers\"],\"snipp...
  [FAIL] TestMachineFormatsEscapeC1Controls/search_--query_merge_--format_json
  [FAIL] TestMachineFormatsEscapeC1Controls/search_--query_merge_--format_ndjson
  [FAIL] TestMachineFormatsEscapeC1Controls/snapshot_--format_ndjson
  [FAIL] TestMachineFormatsEscapeC1Controls/snapshot_--format_compact-ndjson
  [FAIL] TestMachineFormatsEscapeC1Controls/neighbors_--symbol_Merge_--format_json
  [FAIL] TestMachineFormatsEscapeC1Controls/impact_--symbol_Merge_--format_json
```

The cache-replay test was checked for vacuity the same way — reverting only the `root.go:382` wrap while keeping everything else:

```
  [FAIL] TestSnapshotCacheReplayEscapesC1
     termsafe_machine_c1_cache_test.go:49: raw C1 control at byte 1427 of the machine format
     termsafe_machine_c1_cache_test.go:52: replayed stream differs from the stream that was cached
```

## Losslessness

`\^]` decodes to the same code point, so no consumer loses the value. Decoding the search response from both binaries, field by field:

```
DECODED result content identical (paths + snippets): True
  carrier path   : 'evil\x9d52;c;aGVsbG8=\x9cfile.go'
  U+009D in path : True | U+009C in path: True

differing fields: /stats/index_latency_ms, /stats/preselect_latency_ms,
                  /stats/query_latency_ms, /stats/search_latency_ms,
                  /stats/total_latency_ms
NON-latency differences: NONE
```

On a repository with no control bytes the output does not move at all — comparing `main` and this branch over this repo:

```
IDENTICAL 23286693B   snapshot --format ndjson
IDENTICAL 19049820B   edges --format ndjson
IDENTICAL  5505188B   snapshot --format compact-ndjson
IDENTICAL  4009441B   symbols --format ndjson
IDENTICAL     4198B   search --query termsafe --format text
```

The remaining JSON verbs differ only in their latency fields (checked field by field, as above).

## Review follow-up: the two search replay paths (429dd34)

Trail finding 01M0DGDR3EJ4 — reproduced, fixed. Wrapping the encoders fixes the bytes **this** build produces; `search` has two sinks that reach no encoder and print bytes an **earlier** build produced:

- **`ENTIRE_GRAPH_PRESEARCH`** — `echoPresearchPayload` writes the named file to stdout verbatim, ahead of the profile, the repo, the cache and the index. The file is a payload another process encoded, so nothing in this process escaped it.
- **`EG_SEARCH_SESSION`** — `searchSession.echo` replays the first search of a task for every later one. A state file written by a build older than this rule holds that build's raw output and keeps replaying it for the life of the task.

Both are the `root.go:382` failure in the same shape, and the format they replay is the one the rule is about — `search` defaults to `--format json`, so a warm replay of a stale payload is the *default* invocation emitting the raw control.

Reproduced at `9bceb97` (this PR's own head before the follow-up) against a stale payload holding U+009D and U+009C raw, which is what a pre-rule build wrote to those files:

| replay path | at 9bceb97 | after 429dd34 |
|---|---|---|
| `ENTIRE_GRAPH_PRESEARCH=stale.json search --query merge` | rawC1=2 escaped=0 | rawC1=0 escaped=2 |
| `EG_SEARCH_SESSION=stale-state.json search --query merge` | rawC1=2 escaped=0 | rawC1=0 escaped=2 |

```
$ ENTIRE_GRAPH_PRESEARCH=stale.json eg-9bceb97 search --repo "$HOSTILE" --query merge | od -c | sed -n 5p
0000100    l   e   _   p   a   t   h   "   :   "   e   v   i   l 302 235
```

Fixed by one sink, `replaySearchPayload`, that both replay paths go through — not a check at each call site, for the reason the encoder wrap already gave.

`termsafe.JSONWriter` and not `termsafe.Writer`, because the escape has to be safe for a payload whose format this code does not know. `Writer` rewrites DEL to `\x7f`, and DEL is a byte `encoding/json` leaves **raw** inside a string, so a payload holding one would come out of a replay with an escape JSON does not define. Measured on Go 1.26.5:

```
json.Marshal DEL       : "{\"p\":\"a\x7fb\"}"   raw 0x7f: true  decodable: true
via termsafe.Writer    : "{\"p\":\"a\\x7fb\"}"  decodable: false
via termsafe.JSONWriter: "{\"p\":\"a\x7fb\"}"   decodable: true
```

The text and agent renderers already write through `termsafe.Writer`, so a payload recorded in those formats arrives with its C1 controls already in `\u00XX` form and every other byte below 0x80 — `JSONWriter` leaves all of it untouched, which is what makes one wrapper correct for every format a payload can be in. Verified: a hostile-repo `--format text` payload (4 escapes, 0 raw C1) and its `--format agent` counterpart replay byte-identically.

Lossless and output-stable. The replayed pathname still decodes to the same code points, and a payload with no control bytes replays byte-identically:

```
DECODED file_path after replay: 'evil\x9d52;c;aGVsbG8=\x9cfile.go'   U+009D: True  U+009C: True

replay of a clean payload, 9bceb97 vs 429dd34, and vs the source file:
IDENTICAL json    14223 B     IDENTICAL text   4198 B
IDENTICAL ndjson  14606 B     IDENTICAL agent  6972 B
```

**Limit stated plainly:** the guarantee is the C1 range only. A payload predating `termsafe` entirely could hold a raw ESC, and `JSONWriter` does not rewrite it — a C0 escape inside a JSON string would be an escape JSON does not define, exactly as `\x7f` is. That is recorded in the comment on `replaySearchPayload` rather than papered over.

`TestSearchReplayEscapesC1` builds its stale payload by turning this build's own escapes back into raw code points — the poisoning idiom `TestSnapshotCacheReplayEscapesC1` already uses, so the fixture cannot drift from the real payload shape, and it `t.Fatal`s if there was no escape to poison. Against pristine `9bceb97` (production files untouched, `git status --porcelain` showing only the untracked test file) both cases fail **at runtime**:

```
$ go test ./internal/cli -run TestSearchReplayEscapesC1 -count=1
Go test: 0 passed, 3 failed in 1 packages

  [FAIL] TestSearchReplayEscapesC1/ENTIRE_GRAPH_PRESEARCH
     termsafe_machine_c1_replay_test.go:61: raw C1 control at byte 712 of the machine format
     termsafe_machine_c1_replay_test.go:65: replayed payload differs from the stream that was recorded
  [FAIL] TestSearchReplayEscapesC1/EG_SEARCH_SESSION
     termsafe_machine_c1_replay_test.go:101: raw C1 control at byte 804 of the machine format
     termsafe_machine_c1_replay_test.go:107: replayed payload differs from the stream that was recorded
```

The new test needs no `runtime.GOOS` guard of its own: it rests entirely on the body channel — a C1 sequence inside `main.go`'s `Merge`, representable on every platform — and the pathname channel it inherits from `c1HostileRepo` is already guarded there. It contains no raw C1 bytes, only ASCII escapes.

## Review follow-up: bounded flush instead of a whole-stream copy (8f60d98)

Trail finding 01M0DV7F55SA — reproduced in part, fixed; one premise corrected.

**Reproduced.** Both writers built one buffer the size of the whole input whenever anything in it needed escaping, and the trigger is input the repository controls: one C1 byte anywhere in a cached snapshot takes the no-copy fast path away. Measured on an 8 MiB hostile stream:

| | before | after |
|---|---|---|
| `JSONWriter` allocated | 18 890 752 B | 40 960 B |
| `Writer` allocated | 18 896 088 B | 40 976 B |

40 960 B is one 32 KiB flush buffer plus its headroom, and it does not move with the size of the stream. The buffer never grows either: one position contributes at most six bytes and `escapeHeadroom` is 16, so its cap is never exhausted.

**Corrected.** "OOM instead of streaming" overstates it: the cache-hit path never streamed. `sem.LoadProviderRecords` returns the entire decompressed snapshot as one `[]byte`, so `records` was already fully resident before this PR touched anything. What the wrap added was a *second* full-size buffer, and that marginal doubling is what 8f60d98 removes. Migrating or re-encoding stale entries — the other remedy the finding offers — would not have helped: the copy happens on every hit that contains a C1 byte, whoever wrote the entry.

**The scan still runs over the whole input, and that is the part that must not be chunked.** Both lookahead rules read the byte *after* the one being judged — CRLF passes only as a pair, and a C1 control is recognised from its two-byte form — so escaping one bounded window at a time would judge the last byte of every window against the end of that window rather than against the real next byte, and a CR or a `0xc2` landing there would come out differently. Only the output is in pieces.

`TestFlushBoundariesDoNotChangeTheEscaping` pins that: seven payloads larger than one flush buffer — C1 pairs, CRLF pairs, lone CR, stray C1 bytes, wide runes whose interior bytes fall in the C1 range, a hostile mix, and one with nothing to escape — compared byte for byte against the one-shot escape of the same bytes, for both layouts. It passes before the change too, which is what makes it a regression guard rather than a proof of the change.

`TestWritersDoNotAllocateTheWholeStream` is the failing-first half. Against `429dd34` it fails at runtime for both writers:

```
$ go test ./internal/termsafe -run TestWritersDoNotAllocateTheWholeStream -count=1
Go test: 15 passed, 3 failed in 1 packages

  [FAIL] TestWritersDoNotAllocateTheWholeStream/JSONWriter
     escaping a 8388608-byte payload allocated 18890752 bytes, want at most 2097152
  [FAIL] TestWritersDoNotAllocateTheWholeStream/Writer
     escaping a 8388608-byte payload allocated 18896088 bytes, want at most 2097152
```

It measures with a sink that allocates nothing — a `bytes.Buffer` would allocate the whole output as it grew and swamp the number under test — and it fails first if the sink received no more bytes than it was given, since a case where nothing was escaped would otherwise pass on the fast path.

The escape forms now live in one function, `appendEscapedAt`, used by both the value helpers and the streaming path, so the two cannot drift.

Output is unchanged. Against `429dd34` over this repository, and over the hostile fixture:

```
IDENTICAL 23391773B  snapshot --format ndjson
IDENTICAL 19142304B  edges --format ndjson
IDENTICAL  5527807B  snapshot --format compact-ndjson
IDENTICAL  4020617B  symbols --format ndjson
IDENTICAL     6008B  snapshot --format ndjson   (hostile repo)
IDENTICAL      596B  search --format text        (hostile repo, 4 escapes, 0 raw C1)
```

## Windows

The body channel — a C1 sequence inside a source line of `main.go` — is representable everywhere, and every unconditional assertion rests on it. The pathname channel is skipped on Windows via `runtime.GOOS`; the comment states plainly that the test does not assert anything about Windows pathname behaviour, because it cannot verify it from the platform it was written on. On other platforms it still probes with `os.WriteFile` and downgrades rather than reporting a filesystem limit as a security failure. Test sources contain no raw C1 bytes — only ASCII `\^]` / `\^\` escapes.

## Gate

```
gofmt -l -s .     (CI's exact check)   empty
go vet ./...                           clean
go build ./...                         clean
go test ./...                          2839 passed in 21 packages
go test -race ./internal/termsafe ./internal/cli   green (11.0s / 12.0s)
```

## Scope note

`internal/cli/termsafe_verbs_test.go:131-134` still carries the disproven premise in the comment on `TestMachineFormatsKeepTheRepositoryBytes` ("encoding/json already keeps it off the terminal"). That test is #96's regression test and still passes unchanged — `\^[` decodes back to ESC, so its round-trip assertion holds — so it is left to its owner rather than edited here.

Reopens #96.


ashtom:

**Verdict: still needed — I reproduced the vulnerability on current `main`**

- **7 commits ahead, 219 commits behind** current `main`.
- Test-merging onto `main` **conflicts in 2 file(s)**:

```
internal/cli/root.go
internal/cli/search.go
```

### The premise holds: this is live on `main` today

Before recommending anything about a 219-commit-old branch, I checked whether `main` had already
closed the hole. It has not. `main` has no `JSONWriter`, and its `termsafe` header still asserts the
thing this PR says is false — that `encoding/json` escaping makes the machine formats safe.

I built `main` and ran it against a repository whose *source* carries the C1 pair U+009D (OSC) /
U+009C (ST):

```
$ printf 'def handler(value):\n    # \xc2\x9d0;pwned\xc2\x9c marker\n    return value\n' > pkg/app.py
$ entire-graph search --repo . --query handler --top-k 3   # json is the default format
```

```
raw C2 9D (OSC) in JSON output : 1
raw C2 9C (ST)  in JSON output : 1
escaped u009d in JSON output   : 0

00000250: 6c75 6529 3a20 2320 c29d 303b 7077 6e65  lue): # ..0;pwne
000002d0: 2020 2320 c29d 303b 7077 6e65 64c2 9c20    # ..0;pwned..
```

The same query with `--format text` escapes them correctly (`raw C2 9D in TEXT output: 0`). So the
split this PR describes is exactly the split `main` still has, and the default format is the
unprotected one.

**That makes this the highest-value PR of the ten I had missed.** It is also the one whose staleness
costs the most, because the hole stays open while the branch waits.

### What this needs

### Entire trail findings

### Verified vs. assumed

paths; trail findings re-queried immediately before posting; and building `main` and reproducing the C1 leak through `search`'s default JSON output.

suhaanthayyil:
Merged `main` into the branch and re-ran the gate on the result. `1826cee6` → `ff0fed3c`.

## The premise, re-checked on `bb41f223`

Reproduced independently before touching the branch, on a repo whose *source* carries U+009D (OSC) / U+009C (ST):

```
$ printf 'def handler(value):\n    # \xc2\x9d0;pwned\xc2\x9c marker\n    return value\n' > pkg/app.py
$ eg-main search --repo . --query handler --top-k 3   # json is the default
raw C2 9D (OSC): 2   raw C2 9C (ST): 2   escaped u009d: 0
$ eg-main search --repo . --query handler --top-k 3 --format text
raw C2 9D: 0   raw C2 9C: 0
```

Three sinks are open on `main` today:

| sink | `main` | |
|---|---|---|
| `writeSearchResponse` json | `json.NewEncoder(out)`, `internal/cli/search.go:519` | unwrapped |
| provider-record cache replay | `opts.Stdout.Write(records)`, `internal/cli/root.go:421` | unwrapped |
| `echoPresearchPayload` | `out.Write(payload)`, `internal/cli/search.go` | unwrapped |

One is **already closed** on `main` and this branch was wrong to still be carrying its own version of it — see the search.go resolution below.

## The merge

**`internal/cli/root.go`** — `main` replaced `sem.LoadProviderRecords` with the `BeginProviderRecordsCache` transaction and its `recordsCache.Load()` replay. Main's structure kept in full; only the sink changed, to `termsafe.NewJSONWriter(opts.Stdout)`. That is the branch's point: a cache entry written by a build predating the rule streams the repository's raw control on every hit, and escaping is idempotent so a clean entry passes through unchanged.

**`internal/cli/search.go`** — `main` rebuilt the session echo behind a replay-schema / policy-fingerprint / format / repo / tree / payload-path bracket and already writes the payload through `termsafe.NewWriter`. This branch's older one-line `replaySearchPayload(opts.Stdout, …)` is **dropped, not reinstated**, and the echo keeps main's text writer.

That is deliberate and it reverses what this branch originally proposed. `searchFormatSupportsReplay` (`search.go`) admits only `text` and `agent`, so the bytes reaching that sink are terminal-bound prose. `termsafe.Writer` is the correct rewrite for them; the C1-only JSON rewrite would let a raw ESC or DEL recorded by an older build straight through. The JSON rewrite belongs to the presearch sink, whose format this process never chose — that call merged cleanly and is unchanged.

## A semantic conflict the textual merge could not show

`main` gained `--format scip` (#97), which writes a **binary protobuf** through the same `newRecordEncoder` this branch wraps. The naive merge escapes it, and `TestSnapshotSCIPEmitsBinaryIndexAndOmissionNote` fails with `proto: cannot parse invalid wire-format data`. The C1 rewrite is defined over a text stream; applied to the wire format it rewrites any `0xc2 0x8X` pair — a varint, a length prefix, a UTF-8 name — into six ASCII bytes.

`scip` now returns before the wrap. Its sibling stream, the JSON omission note on stderr, is bound for a terminal and takes the wrap every other machine encoder in that file has. New counter-case test `TestSCIPIndexIsNotC1Escaped` asserts the hostile pathname survives the protobuf intact and carries no JSON escape, while the note beside it carries no raw C1.

## Test the merge obsoleted

`TestSearchReplayEscapesC1`'s `EG_SEARCH_SESSION` case built its state file by hand. Against main's echo that file matches nothing, and the run silently degrades to a fresh search — the case's own guard reports it as *"the echo did not fire, so this case proves nothing"*. Three separate reasons: the echo admits only `text`/`agent`, a worktree view is deliberately non-replayable (`ResolveSearchReplayPolicy` returns an empty policy), and `search` defaults to `--worktree`.

Rebuilt on a real recording run — record with `--format text --head`, then poison only the stored payload back to raw controls — so every scope field is exactly what this build writes. Nothing was skipped or relaxed; the case now actually reaches the sink it names.

## Gate, on the merged tree

| check | result |
|---|---|
| `go build ./...` | pass |
| `gofmt -l`, `go vet ./internal/cli ./internal/termsafe` | clean |
| `go test ./internal/cli` | ok 26.6s |
| `go test ./internal/termsafe` | ok |

Behaviour, same fixture, merged binary: `raw C2 9D: 0  raw C2 9C: 0  escaped u009d: 2  u009c: 2`, and the document still decodes as JSON.

## Mutation verification

Each mutation reverts the production change only, compiles, and was confirmed applied by grepping the file afterwards.

| mutation | tests that flipped to FAIL |
|---|---|
| re-wrap `scip` in `newRecordEncoder` | `TestSCIPIndexIsNotC1Escaped`, `TestSnapshotSCIPEmitsBinaryIndexAndOmissionNote` |
| `json.NewEncoder(termsafe.NewJSONWriter(out))` → `json.NewEncoder(out)` | `TestMachineFormatsEscapeC1Controls` |
| cache replay → `opts.Stdout.Write(records)` | `TestSnapshotCacheReplayEscapesC1` |
| `replaySearchPayload` → `out.Write(payload)` | `TestSearchReplayEscapesC1/ENTIRE_GRAPH_PRESEARCH` |

Restored: all pass.

## Not verified

darwin/arm64, Go 1.25.7 only — nothing here speaks to Windows or Linux. I ran `internal/cli` and `internal/termsafe` in full and did not sweep the rest of the tree; CI covers that.

ashtom:

**Verdict: clean gate, and the vulnerability is still live on `main` — this is the one to merge first**

### Correction

My last comment said this was stale and unevaluable. The branch is now **0 behind** and merges clean,
so that part is superseded. The substantive finding is not.

### The hole is still open on `7b3252aa`

I rebuilt `main` at its new head and re-ran the reproduction rather than trusting yesterday's result.
`main` still has no `JSONWriter`:

```
raw C2 9D in default (json) output : 1
escaped u009d in json output       : 0
raw C2 9D in --format text output  : 0
```

A repository whose *source* carries U+009D (OSC) / U+009C (ST) still gets those bytes delivered raw
through `entire graph search` with no `--format` flag, because json is that verb's default. The
text path escapes them correctly. That split is exactly what this PR closes.

| Check | Tree | Result |
|---|---|---|

### Merge-order note

**#135 and #143 now conflict with each other** in `internal/termsafe/termsafe.go`:

```
$ git merge-tree --write-tree refs/prs/135 refs/prs/143
CONFLICT (content): Merge conflict in internal/termsafe/termsafe.go
```

Each is clean against `main` alone. Whichever lands second needs a manual resolution.

### Entire trail findings

Re-queried immediately before posting: **none open and current**.

suhaanthayyil:
Confirmed, with the merge check you would want before acting on "merge first".

**Gate re-verified against current `main`:** all three `test` legs and all three `build` legs pass, plus lint. The branch is **4 commits behind** `main`, and a trial merge is clean.

**One thing I checked because of a near-miss elsewhere today.** `git diff main..branch` on this branch shows `scripts/create-nightly-tag.sh` as deleted (36 lines) and 1,488 deletions overall. That is an artifact of the branch predating #186, not something this PR removes — a two-dot diff shows main's newer files as "absent from the branch". I verified the real merge result rather than trusting the diff view:

```
merge-tree main x this branch
  scripts/create-nightly-tag.sh   -> present
  .github/workflows/nightly.yml   -> present
```

Same check run on #138, #140 and #143: all four preserve main's files. Flagging the method because a whole-file conflict resolution silently reverted 186 lines on another PR today, and a two-dot diff is exactly the view that makes that look normal.

**On "the one to merge first":** I am not merging it. The standing instruction here is that only PRs the user explicitly names get merged, so this is ready and waiting on their call rather than mine.

PR #138, opened 2026-08-19
fix(sem): confine working-tree source reads to the repository root

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/98

## The bug

`openSource`'s working-tree branch resolved every read as `filepath.Join(repo, path)` and then
checked the result with `os.Lstat` (`internal/sem/provider.go:9836-9838` and `:9852-9854` on
`main`). Two holes:

1. **`os.Lstat` bounds only the final component.** An intermediate symlinked *directory* is
   followed, so `link/secret.env` where `link -> ../` lands outside the repository.
2. **`filepath.Join` normalizes the climb before the check runs.** `"../secret.env"` becomes
   `<repo>/../secret.env`, which is a regular file, so the guard passes and the file is read.

There is a **third** unconfined read one call later. `oversizeRegistry` stored the joined absolute
path (`:9842`) and digested it with `filedigest.File` (`:9922`) with no check at all — so a file the
reader had just refused was still opened, and its size, SHA-256 and line count were reported.

## Why the callers are not the boundary

This reopens #88 / #95. The earlier argument was that callers only pass snapshot-built paths. They
do not: `jsLocalImportCandidates` (`provider.go:10426`) builds candidates from an **import specifier
written in a repository file**, and `resolveReadableLocalImport` (`provider.go:10414`) hands each one
to the reader. For `src/a.ts` containing `import {a} from "../../secret.env"`, `filepath.Join` turns
the specifier into `../secret.env` — pinned by `TestJSImportCandidatesCanClimbAboveRepositoryRoot`.
The path is attacker-authored, so containment has to be structural.

## Impact, stated precisely

This is a **boundary breach plus an existence oracle**, not confirmed content disclosure. The
escaping read is real and returns the outside file's bytes (see the failing-first output below), but
at the one reachable call site the content is discarded — `resolveReadableLocalImport` only tests
`if _, ok := readContent(candidate); ok` — and the out-of-corpus path is never parsed into records, so
no snippet of it is emitted. What escapes today is that a file exists at an arbitrary path outside
the repository, and the read itself. I did not find a path by which the content reaches output, and I
am not claiming one.

## The fix

One `os.Root` over the repository for the lifetime of the opened source; every working-tree read
resolves through it — content reads, prefix reads, and the registry's deferred digest, which now
streams through the root (`filedigest.Stream`) instead of reopening a stored absolute path.

- `os.Lstat` is **kept alongside** `os.Root`, copying `readRepoFileLines`
  (`internal/cli/callsite.go:210`) including its reasoning: the previous refusal of a symlinked final
  component survives.
- The root is opened **once**, not per read (i.e. not `cache_entry.go:71`'s granularity): `os.Root`
  methods are goroutine-safe and `provider_parallel.go:76` drives these closures concurrently. Cost
  is one extra `openat` per opened source; per-file cost is unchanged because `root.Lstat` /
  `root.Open` are the same two syscalls, resolved relative to the root descriptor.
- `openedSource.close` was `nil` on this branch and now returns `root.Close`. The consumers already
  nil-guard (`provider.go:940`, `search_cache.go:566`, `search.go:809` and `:1715`), so this chains
  rather than replaces.

**Scope / rebase note.** This touches only the **working-tree branch** of `openSource` and the
`oversizeRegistry` block directly below it. The committed-revision (`--head`) branch of the same
function is **not touched** and cannot regress — it reads through `git cat-file` and resolves no
filesystem path. Another PR in this series edits that region.

## Containment is not the only behavior that changes

`os.Root` resolves **every** path component inside the root, not just the last one, so it splits
paths that `filepath.Join` + `os.Lstat` treated alike. Three cases, all pinned by
`TestSymlinkedDirectoryResolutionInsideRepository`:

| path reaches its file through | before | now |
|---|---|---|
| a relative link resolving inside the repo (`pkg -> realpkg`) | read | read — unchanged |
| a link that leaves the repo (`link -> ../`) | read | refused — the escape being closed |
| an **absolute** link target, even one pointing back inside the repo (`pkg -> /abs/repo/realpkg`) | read | refused — `os.Root` will not rebase an absolute path onto itself |

The third case needs no adversary and no link out of the tree. It is reachable because the listing
comes from `git ls-files --cached` (`gitutil.ListWorktreeFiles`), which reports paths the **index**
records: a directory that was committed and has since been replaced on disk by a symlink still lists
its files.

What the user observes is not just an unreadable file. The read failure is reported through the
existing partial-failure path as an **error**-severity `E_FILE_READ`
(`provider_parallel.go:119-125`), so the file's symbols and relations disappear from the snapshot
**and the whole snapshot's `completeness_level` drops from `ok` to `unsafe`**, which prints
`⚠️ graph is UNSAFE ... treat query results as partial` on stderr:

```
$ git init -q . && git config user.email t@example.com && git config user.name t
$ mkdir pkg && printf 'package pkg\n\nfunc Helper() int { return 1 }\n' > pkg/lib.go
$ printf 'package main\n\nfunc main() {}\n' > main.go
$ git add -A && git commit -qm init
$ mv pkg realpkg && ln -s "$PWD/realpkg" pkg
$ git ls-files --cached
main.go
pkg/lib.go
$ entire-graph@main symbols --repo . --worktree
completeness_level: ok symbols: 3
$ entire-graph@PR symbols --repo . --worktree
completeness_level: unsafe symbols: 2
partial_failure: E_FILE_READ error pkg/lib.go - file listed but content was unavailable
```

(the two `symbols` lines are the `summary` record's `stats` and `partial_failures`, extracted from
`--format ndjson`). `filepath.WalkDir`, the fallback listing used when `git ls-files` fails, never
descends a symlinked directory and so cannot produce such a path at all.

This is the cost of making the boundary structural, and it is stated here rather than in a "nothing
else changes" line because a repository that hits it sees an `unsafe` graph, not a diagnostic about
a link.

## Failing-first proof

Production file restored to `origin/main`, new test file kept. It fails at **runtime**, not compile
time:

```
$ git show origin/main:internal/sem/provider.go > internal/sem/provider.go
$ go test ./internal/sem/ -run 'TestWorktreeRead|TestOversizeRegistryRefuses|TestWorktreeSourceReads' -count=1
--- FAIL: TestOversizeRegistryRefusesPathAboveRepositoryRoot (0.01s)
    provider_worktree_containment_test.go:103: oversize registry described a file above the repository root: {Bytes:4096 Hash:a2e659dacb4691e887ac0139f8893d04764ee197d70fb73d3190d56113d18e3e Lines:1}
--- FAIL: TestWorktreeReadRefusesPathAboveRepositoryRoot (0.01s)
    provider_worktree_containment_test.go:56: read escaped the repository root and returned "SECRET_TOKEN=hunter2\n"
--- FAIL: TestWorktreeSourceReadsRepositoryFilesAndClosesItsRoot (0.01s)
    provider_worktree_containment_test.go:124: worktree source returned no closer; the root it opens would leak for the process lifetime
--- FAIL: TestWorktreeReadRefusesSymlinkedDirectoryComponent (0.01s)
    provider_worktree_containment_test.go:80: read followed a symlinked directory out of the repository and returned "SECRET_TOKEN=hunter2\n"
FAIL
FAIL	github.com/entireio/entire-graph/internal/sem	0.686s
FAIL
```

The behavior-change test is failing-first too, against the same restored file — this is `main`
following an in-repo link that leaves and returns:

```
$ go test ./internal/sem/ -run 'TestSymlinkedDirectoryResolutionInsideRepository' -count=1
--- FAIL: TestSymlinkedDirectoryResolutionInsideRepository (0.01s)
    provider_worktree_containment_test.go:168: read(updown/a.ts) = "export const a = 1;\n"; a link leaving the repository is refused even when it returns to it
FAIL
FAIL	github.com/entireio/entire-graph/internal/sem	0.596s
FAIL
```

With the fix all six pass. These tests run in parallel, so the order of the `--- FAIL` blocks and the
per-test durations vary between runs; the failure lines themselves are the invariant. The two
symlink tests skip on Windows (`runtime.GOOS` guard): creating a directory symlink there needs
`SeCreateSymbolicLinkPrivilege`, which CI runners do not grant. The `../` traversal case needs no
symlink and runs on every platform.

## No regression

`edges --repo . --worktree --format ndjson` over this repository, `main` vs this branch:
**byte-identical**, 39,178 lines (39,177 records plus the header). Wall clock over 3 alternating
runs each — main: 13.24 / 13.84 / 14.37 s; branch: 13.43 / 13.75 / 14.20 s. Within run-to-run noise
(both arms drift upward together).

## Gate

- `gofmt -l -s .` — empty
- `go vet ./...` — clean
- `go build ./...` — ok
- `go test ./...` — all packages ok (2,803 tests, 21 packages)
- `go test ./internal/sem -race -run 'Parallel|Worktree|Oversize|Symlinked'` — ok (37 tests, includes
  the `Worktree: true` parallel-worker snapshots that exercise the shared root concurrently)

Schema unchanged. No new dependency. Provider stays no-egress.


ashtom:

- **4 commits ahead, 219 commits behind** current `main`.
- Test-merging onto `main` **conflicts in 1 file(s)**:

```
internal/sem/provider.go
```

### What this needs

### Entire trail findings

### Verified vs. assumed

suhaanthayyil:
Merged `main` (`bb41f223`) into this branch and re-verified. Head is now `f5d69dc`.

## Is the hole still open on `main`?

Yes. `5ebafdbe` ("harden Git path plumbing") hardened the *listing* and the `.git`-pointer
sweep, not the working-tree *content reader*. On `bb41f223` the worktree reader is still the
pre-change shape:

```
internal/sem/provider.go:9967   read := func(path string) (string, bool) {
internal/sem/provider.go:9968       full := filepath.Join(repo, filepath.FromSlash(path))
internal/sem/provider.go:9969       info, err := os.Lstat(full)
internal/sem/provider.go:9983   readPrefix := func(path string, limit int) (string, bool) {   // same shape
internal/sem/provider.go:10136      if digest, err := filedigest.File(pending.full); err == nil {
```

`filepath.Join` normalises a leading `../` before the `os.Lstat` runs, and `os.Lstat` only
inspects the final component, so an intermediate symlinked directory is followed. Reproduced
against `bb41f223` with a temporary in-package test (fixture: `repo/src/a.ts` plus a sibling
`secret.env` one level above the root):

```
$ go test ./internal/sem/ -run 'TestZZRepro138' -count=1 -v
--- FAIL: TestZZRepro138Traversal
    VULN read(../secret.env) escaped: "SECRET_TOKEN=hunter2\n"
    VULN readPrefix(../secret.env) escaped: "SECRET_T"
--- FAIL: TestZZRepro138SymlinkDir
    VULN read(link/secret.env) escaped: "SECRET_TOKEN=hunter2\n"
    VULN readPrefix(link/secret.env) escaped: "SECRET_T"
--- FAIL: TestZZRepro138Oversize
    VULN oversize described outside file: {Bytes:4096 Hash:a2e659dacb4691e887a... Lines:1}
```

All three read paths (content, prefix, deferred digest) still escape. The reachability the
change is premised on also still holds: `jsLocalImportCandidates` is at
`internal/sem/provider.go:16075` on `main` and still turns `../../secret.env` in a repository
file into the candidate `../secret.env`.

## What the merge required

One file conflicted, `internal/sem/provider.go`, in two hunks. Neither side was taken
wholesale.

**Hunk 1 — worktree source preamble.** `main` added a `sweepWarnings` fold
(`warnings = append(warnings, sweepWarnings...)`) and kept `newOversizeRegistry()`; this branch
replaced the preamble with the `os.OpenRoot` acquisition and `newOversizeRegistry(root, repo)`.
Resolved by keeping both, with `main`'s warning fold placed *before* the root acquisition —
the permission fallback passes `warnings` onward, so folding after it would have dropped the
sweep warning on that path.

**Hunk 2 — end of `openSource` plus new declarations.** `main` added `ignores: ignores` to the
worktree `openedSource` and a new `fallbackOversizeRegistry` type after the function; this
branch added `close: root.Close` and the four confinement helpers (`containedRealPath`,
`readFallback`, `readPrefixFallback`, `openManuallyConfinedWorktreeSource`). Both sides kept.
`main`'s new `ignores` field then had to be propagated into the fallback constructor as well,
which this branch introduced after `main` wrote that line — `openManuallyConfinedWorktreeSource`
now takes an `ignoreMatcher` and sets the field, so the permission-fallback source is not
silently missing the matcher that `analyze.go:989`, `analyze.go:1012`,
`search_replay_policy.go:148` and `search.go:1843` read.

## One behaviour this branch asserted is gone from `main`

`TestWorktreeSourceOpensWithAnExecuteOnlyRepositoryRoot` no longer describes the code and now
fails for a reason unrelated to containment. `main` added an unconditional listing preflight
that holds an `os.Root` over the repository *before* any listing runs:

```
internal/sem/provider.go:14909   root, err := newSweepDirectoryRoot(resolvedRepo)
internal/sem/provider.go:14911   return fmt.Errorf("%w: %w: hold selected root: %v", ...)
```

so an execute-only repository root is refused by `worktreeSourceFiles`, before `openSource`
reaches its own `os.OpenRoot` at all. Verified independently of this branch, on `bb41f223`:

```
$ go test ./internal/sem/ -run TestZZRepro138ExecOnlyRootOnMain -count=1 -v
    MAIN ALREADY REFUSES exec-only root: Git worktree preflight declined subprocess
    enumeration: filesystem worktree fallback cannot safely traverse the refused path:
    hold selected root: open .../repo: permission denied
--- PASS
```

The permission branch below the listing is therefore defensive only now (a root whose mode
changes between preflight and read; listing paths that bypass the preflight), not the
end-to-end capability the old test asserted. Rather than delete the coverage, the test was
replaced by two in commit `f5d69dc`:

- `TestExecuteOnlyRepositoryRootIsRefusedByTheListingPreflight` — pins `main`'s new refusal and
  its `hold selected root` reason, so a future change that silently re-admits such a repository
  is caught.
- `TestManuallyConfinedWorktreeSourceReadsAndConfines` — exercises the fallback reader directly:
  nil closer, in-repository read and prefix-read both work, and `../secret.env`, `../big.env`
  and the deferred digest are all refused.

Net coverage is up, not down. No other test was changed, skipped or weakened.

## Re-verification on the merged tree

```
$ go build ./...                 # clean
$ go vet ./internal/sem/         # clean
$ gofmt -l internal cmd          # empty
$ go test ./internal/sem/ -run '<the 11 containment tests>' -count=1
Go test: 11 passed

$ go test ./internal/sem/ -run 'Symlink|Oversize|Worktree' -count=1        ok  6.6s
$ go test ./internal/sem/ -run 'Snapshot|Source|Ignore|Cache' -count=1     ok  25.6s
$ go test ./internal/sem/ -run '<containment>|Symlink|Oversize|Worktree'   135 passed
```

## Mutation check

Two independent mutations, each applied to the merged tree, each confirmed applied by grepping
the file afterwards, each compiling (`go build ./...` clean).

**Mutation A** — put the rooted reader back on the unconfined path: `root.Lstat(name)` →
`os.Lstat(filepath.Join(repo, name))` and `root.Open(name)` → `os.Open(filepath.Join(repo, name))`
in `read`, in `readPrefix`, and in `oversizeRegistry.digestContained` (6 sites, all verified by
grep). Result — 4 failed, 7 passed:

```
FAIL TestWorktreeReadRefusesPathAboveRepositoryRoot
FAIL TestWorktreeReadRefusesSymlinkedDirectoryComponent
FAIL TestOversizeRegistryRefusesPathAboveRepositoryRoot
FAIL TestReadFallbackRefusesAnAbsoluteSymlinkThatEscapes
```

**Mutation B** — strip the containment check out of `containedRealPath` (drop the
`filepath.Rel` / `..` prefix test, return the resolved path unconditionally). Result:

```
FAIL TestManuallyConfinedWorktreeSourceReadsAndConfines
     fallback read escaped the repository root and returned "SECRET_TOKEN=hunter2\n"
     fallback readPrefix escaped the repository root and returned "SECRET_T"
     fallback oversize registry described a file above the repository root: {Bytes:4096 ...}
FAIL TestReadFallbackRefusesAnAbsoluteSymlinkThatEscapes
     read(abs-outside/secret.env) = "SECRET_TOKEN=hunter2\n"
```

Restored in both cases — 11 passed. Every containment test has teeth; the positive tests
(in-repository reads, the three in-repository symlink shapes, the 10-hop chain) stay green
under both mutations, so they are not passing by accident either.

suhaanthayyil:
The gate came back with one unresolved HIGH finding against the merged code. **Reproduced, confirmed, fixed in `c2fc0ee5`.**

> The fallback re-resolves the mutable `repo` pathname instead of using the already-open `os.Root`. If the repository path is a symlink, or is renamed and replaced after `openSource` pins it, a path that triggers the fallback (such as an absolute intermediate symlink) is validated against and read from the replacement tree, allowing content outside the originally selected worktree to enter output.

## Reproduction

`openSource` pins the worktree with `os.OpenRoot`, but `readFallback` / `readPrefixFallback` / `digestFallback` take the repository **pathname**, and `filepath.Join`, `filepath.EvalSymlinks` and `os.ReadFile` all ran at read time against whatever that name pointed at:

```go
func containedRealPath(repo, relPath string) (string, bool) {
	dir, base := filepath.Split(filepath.Join(repo, filepath.FromSlash(relPath)))
	realRepo, err := filepath.EvalSymlinks(repo)   // <- read time, not open time
```

Symlinked root, pinned, then repointed to a second tree:

```
before swap: ok=true content="package selected\n"
AFTER swap (root still pinned to selected): ok=true content="package replacement\n"
pinned root still reads: "package selected\n"
```

The held descriptor and the fallback disagree about which repository they are reading. The finding is right, and the severity is right: the fallback is reachable from repository-controlled input — `jsLocalImportCandidates` derives candidates from an import specifier written in a repository file, which is the reason this confinement could not be left to the callers in the first place.

## Fix

The finding offered two remedies — tie the fallback to the held root, or reject repository identity changes. The first is not available: the fallback exists precisely for the paths `os.Root` refuses to resolve (an intermediate symlink with an absolute target, a chain past its 8-hop limit), so the descriptor cannot resolve them by construction. So the second.

The selected directory's identity is taken **once**, when the source opens:

```go
func pinnedRootIdentity(root *os.Root, repo string) fs.FileInfo {
	if root != nil {
		if info, err := root.Stat("."); err == nil {
			return info
		}
	}
	if info, err := os.Stat(repo); err == nil {
		return info
	}
	return nil
}
```

`root.Stat(".")` asks the pinned descriptor itself; `os.Stat` is the answer for the execute-only root `os.OpenRoot` declines to open, which reaches `openManuallyConfinedWorktreeSource` and now pins its identity by name at open time rather than re-deriving it per read. Every fallback resolution must then land on that same directory before it reads:

```go
if pinned == nil {
	return "", false
}
realRepoInfo, err := os.Stat(realRepo)
if err != nil || !os.SameFile(pinned, realRepoInfo) {
	return "", false
}
```

An identity that could not be taken is refused rather than trusted. `digestFallback` re-resolves the same way, so the registry carries the same pin — otherwise the deferred digest would reintroduce the hole one call after the reader closed it.

## Verification

`TestFallbackRefusesAReplacedRepositoryRoot` drives the reproduction above and adds a non-over-refusal case: a repository read through its own unchanged path still returns the file.

| check | result |
|---|---|
| mutation: drop the `os.SameFile` guard (compiles, grep-confirmed) | `TestFallbackRefusesAReplacedRepositoryRoot` **FAIL** — `fallback read "package replacement\n" from the REPLACEMENT tree` |
| restore | pass |
| this PR's containment suite (12 tests, incl. `TestExecuteOnlyRepositoryRootIsRefusedByTheListingPreflight`, `TestManuallyConfinedWorktreeSourceReadsAndConfines`, `TestReadFallbackFollowsASymlinkChainLongerThanOSRootsLimit`) | 12/12 pass |
| `internal/sem` `TestProvider*/TestWorktree*/TestOversize*/TestSnapshot*/TestSource*/TestIgnore*` | ok 6.6s |
| `go test ./internal/cli ./internal/gitutil ./cmd/...` | ok |
| `go build ./...`, `go vet`, `gofmt -l` | clean |

darwin/arm64, Go 1.25.7. The test skips on Windows: repointing a directory symlink mid-scan is not the shape this race takes there, and the assertion would describe a platform behaviour I cannot verify from here.

suhaanthayyil:
The previous fix resolved the first HIGH (now marked stale) and the re-review raised a second one against it. **Reproduced in structure, confirmed, fixed in `c9a0caec`.**

> The fallback validates the resolved directory, then opens the resulting pathname separately. If an in-repository directory is renamed and replaced with a symlink after `EvalSymlinks` but before this read, the open follows it outside the repository; `readPrefixFallback` and `digestFallback` have the same TOCTOU escape, so the resolved directory must be pinned and the final component opened relative to that handle.

This is correct, and it is a different bug from the one I fixed. The first pinned the repository *root* identity; this one is a component *below* it. `containedRealPath` returned a pathname and the reader then re-opened that pathname — two resolutions of the same string with a writable window between them.

## Fix

Exactly what the finding prescribes. `containedRealDir` resolves the parent, pins it with `os.OpenRoot`, and `openContainedRegularFile` does `dirRoot.Lstat(base)` / `dirRoot.Open(base)` — so no component above the leaf can be swapped between the check and the read. All three readers (`readFallback`, `readPrefixFallback`, `digestFallback`) now share that one function.

The window between resolving the name and pinning it is closed by validating again afterwards and requiring the pinned handle to be the directory that second validation contains: a swap **before** the pin fails the re-validation's containment check, and a swap **back** after the pin fails the identity comparison, because the handle is still on the directory that was there when it was taken.

Two things fell out of routing all three through one function: the size bound is now taken from the **open** file rather than an earlier `Lstat` of the same name, and a file that grows past the bound between check and read is refused rather than truncated.

## On the test, honestly

My first attempt at a regression test did **not** work, and I only found that out by mutating. I reverted `openContainedRegularFile` to the pathname-based open, and both tests I had written still passed:

```
--- PASS: TestContainedRealDirPinsTheDirectoryItValidated
--- PASS: TestFallbackNeverReadsThroughASwappedDirectory
```

The first called the helper directly, so it could not see a change in the reader. The second is a 3000-iteration race probe, and the window between `os.Lstat(resolved)` and `os.Open(resolved)` is too narrow to hit reliably. A test that passes against the bug is worth nothing, so I am reporting that rather than quietly keeping it.

The property here is only *observable* under a race no deterministic test can schedule, so the discriminating test is a source-level guard — the idiom this package already uses for class invariants. `TestFallbackOpensOnlyThroughTheDirectoryHandle` asserts `openContainedRegularFile` reaches the filesystem only through the handle, and fails loudly if the function disappears, so it cannot pass vacuously. The deterministic test was retargeted at the readers' entry point, and the race probe is kept as a one-directional probe: it fails only if outside content is actually observed, so it can never fail spuriously.

| check | result |
|---|---|
| mutation: pathname-based open (compiles, grep-confirmed) | `TestFallbackOpensOnlyThroughTheDirectoryHandle` **FAIL** — `no longer opens the leaf through the pinned directory handle`; `reaches the filesystem by pathname via os.Open(`; `via os.Lstat(` |
| restore | pass |
| this PR's containment suite (12 tests) | 12/12 pass, unchanged |
| `internal/sem` `TestProvider*/TestWorktree*/TestOversize*/TestSnapshot*/TestSource*/TestIgnore*/TestFallback*/TestContained*/TestSymlinked*` | ok 7.1s |
| `go test ./internal/cli ./internal/gitutil ./cmd/...` | ok |
| `go build ./...`, `go vet`, `gofmt -l` | clean |

darwin/arm64, Go 1.25.7. Both new filesystem tests skip on Windows: swapping a directory for a symlink mid-read is not the shape this race takes there, and I will not assert a platform behaviour I cannot verify from here. The source guard runs everywhere.

suhaanthayyil:
The `windows-latest` job failed on the guard I added, not on the fix. Cause found and reproduced locally; corrected in `9abe0362`. The Findings gate is green on `c9a0caec` — both HIGHs are resolved.

```
--- FAIL: TestFallbackOpensOnlyThroughTheDirectoryHandle (0.01s)
    worktree_dir_pin_test.go:162: the fallback reaches the filesystem by pathname via os.Lstat(, ...
    worktree_dir_pin_test.go:162: the fallback reaches the filesystem by pathname via os.Stat(, ...
```

## Cause

The guard bounded its scan with `strings.Index(body, "\n}\n")`. Git for Windows checks the tree out with **CRLF**, so that terminator never matched, `end` came back `-1`, and the code fell through to leave `body` as the entire rest of the file. Every `os.Stat` and `os.Lstat` elsewhere in `provider.go` was then reported as a finding against this one function.

Reproduced locally rather than inferred — `provider.go` converted to CRLF in place, old guard restored:

```
--- FAIL: TestFallbackOpensOnlyThroughTheDirectoryHandle (0.01s)
    the fallback reaches the filesystem by pathname via os.Lstat(, ...
        	dirRoot, base, ok := containedRealDir(pinned, repo, relPath)
        	if !ok {
        	file, info, ok := openContainedRegularFile(pinned, repo, relPath)   <- past the end of the function
```

Same message the runner produced, and the body dump shows the scan running past the function into `readFallback` below it.

## Fix

Endings are normalized before the scan, and a missing terminator is now a hard failure instead of a silent fall-through — the unbounded scan is what turned a bounded check into a false positive, so it should not be reachable at all.

Verified in both directions on both endings, which is the matrix the original test was missing:

| | LF | CRLF |
|---|---|---|
| correct code | PASS | PASS |
| pathname-based mutation | FAIL | FAIL |

`internal/sem` (`TestProvider*/TestWorktree*/TestOversize*/TestSnapshot*/TestSource*/TestIgnore*/TestFallback*/TestContained*/TestSymlinked*`) ok 7.0s; `gofmt -l`, `go vet` clean.

Worth noting for the record: nothing about the production fix was wrong, and the failure was in a test I added to catch a race — the guard was checking real source text and was right to exist, it was just not portable. The two filesystem tests beside it skip on Windows deliberately; this one runs everywhere and now does so correctly.

ashtom:

**Verdict: clean gate — my previous comment is superseded**

### Correction

My last comment on this PR said it was stale (219 commits behind) and could not be evaluated without
a manual merge. That is no longer true: the branch has been brought up to date and is now **0 commits
behind** and merges clean. Please disregard the "needs a merge" part of that comment.

| Check | Tree | Result |
|---|---|---|

### Entire trail findings

Re-queried immediately before posting (`--status open --freshness current`): **none**.

suhaanthayyil:
Confirmed — nothing to change, and the supersession matches what I see.

Re-verified against current `main`: all three `test` legs and all three `build` legs pass, plus lint. Branch is 4 commits behind `main`; trial merge is clean.

I also checked that a merge preserves the files `main` gained today (`scripts/create-nightly-tag.sh`, `.github/workflows/nightly.yml`) — `git diff main..branch` shows them as deletions purely because the branch predates #186, which is a two-dot-diff artifact rather than anything this PR does. Worth stating explicitly because a whole-file conflict resolution silently reverted 186 lines on another PR today, and that is the view which makes such a revert look unremarkable.

The HIGH TOCTOU finding raised on the merged tree in the earlier round remains fixed on this head.

PR #140, opened 2026-08-19
fix(sem): bound AST walk recursion (CWE-674)

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/100

## The bug (CWE-674, uncontrolled recursion)

A source file of deeply nested expressions kills every entire-graph verb with
`fatal error: stack overflow`. Tree-sitter builds its tree iteratively on the C
heap, so the tree is as deep as the source; the Go walkers over that tree recursed
once per level with no bound. A Go stack overflow is a fatal, unrecoverable
process abort that `recover()` cannot catch, so the fix has to be a **limit
applied before recursing**, not a rescue afterwards.

Reproduced end to end on `origin/main` @ cceb4de against a repo holding one
1,803,606-byte Python file of 900,000 nested parentheses plus one ordinary module:

```
$ ./eg-main search --repo ./hostile --query "ordinary function"
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
EXIT=2
```

Nothing in front of the parser rejects it: `defaultMaxParseBytes` (`provider.go:37`)
is 4 MiB and the file is 1.72 MiB; `looksMinified` (`similarity.go:195`) only flags
files whose bytes live on lines longer than 5000 columns, and the fixture puts one
character per line.

## Every walker a parse can drive, bounded

`internal/sem` holds **nine** self-recursive functions taking a `*sitter.Node`
plus **three** recursive closures over one — enumerated mechanically (Tarjan SCC
over the package call graph), not by eye, which also established that no
*mutually* recursive node cycle exists. All of them, and what happened to each:

| walker | phase | disposition |
|---|---|---|
| `walkEntitiesScoped` (`parser.go`) | entity walk | bounded — the originally reproduced crash |
| `initializerTypeBodies`' walk | entity walk | bounded — only recursion into a field initializer |
| `collectParseErrorDetails`' walk | after the walk | bounded — ran on every `HasError` file, capped on RESULTS not depth |
| **`firstNameDescendant`** | **before the walk** | **bounded — reproduced abort** |
| **`firstDescendantOfType`** | **before the walk** | **bounded — second descent on the same path** |
| **`unwrapRustItemWrapperMacros`' walk** | **pre-parse** | **bounded — reproduced abort** |
| **`rAssignedValueKind`** | **before the walk** | **bounded — reproduced abort** |
| **`jsPatternBindingNames`** (`js_scopes.go`) | **entity walk** | **bounded — reproduced abort** |
| **`jsFunctionLikeNode`** (`js_scopes.go`) | entity walk | bounded — sibling on the same call |
| **`identifierDescendants`** (`parameters.go`) | **before the walk descends** | **bounded — reproduced abort** |
| **`jsMemberChainParts`** (`js_scopes.go`) | **relation phase** | **bounded — reproduced abort** |
| **`jsScopeWalker.walk`** (`js_scopes.go`) | **relation phase** | **bounded — reproduced abort** |

**Nothing is left unbounded.** An earlier revision of this PR left three walkers
out of scope — the two relation-phase ones and `identifierDescendants` — and a
review finding caught each. All three are now bounded; the reasoning that
excluded them is retracted below where it was wrong.

The two closures that a survey of the package also turns up —
`kotlinReceiverTypeNames`' `visit` (`depth > 4`) and `uniqueImplementedMethod`'s
`walk` (`depth > 16`) — recurse over the SYMBOL graph, not a parse tree, and
already carry their own explicit caps. They are not parse walkers.

### Finding 1 — `firstNameDescendant` / `firstDescendantOfType`

`entityFromNode` calls these on a whole declaration subtree while
`walkEntitiesScoped` is still one or two levels down, so their recursion is
independent of the walk's and reaches the stack first. The reviewer's example
reproduces — but **only above 4 MiB**, and that matters:

```
$ /usr/bin/time -l ./eg-parent diff --repo /tmp/hostile --base HEAD~1 --head HEAD
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
github.com/entireio/entire-graph/internal/sem.firstNameDescendant(...)
EXIT=2      4114186240 maximum resident set size
```

The file is 12,024,026 bytes — one C function declarator nested 6,000,000 deep.
Under `defaultMaxParseBytes` the provider would reject it, so the honest statement
is: **`Analyze` applies no size cap at all.** `analyze.go` hands `gitutil.ShowFile`
output straight to `ParseWithStatus`, so `graph diff` reaches these sizes on a real
repository. (`ProviderSnapshotOptions.MaxParseBytes < 0` removes the cap on the
provider path too.) Same command on this branch:

```
$ ./eg-fixed diff --repo /tmp/hostile --base HEAD~1 --head HEAD
deep.c (C)
  ~ module deep.c body changed (0 dependents)
Warnings
  ! E_PARSE_DEPTH_EXCEEDED deep.c
EXIT=0
```

### Finding 2 — `unwrapRustItemWrapperMacros`

Confirmed: this pass runs as language preprocessing, **before** the guarded walk
and on its own tree, so nothing downstream protects it. The `cfg_*!\s*\{` hint that
arms it is a regex over the whole file and need not be anywhere near the nesting —
one tokio-shaped line at the top of a deeply nested Rust file is enough. Reproduced
at 5,000,000 levels / 10,020,051 bytes: `fatal error: stack overflow`, frames in
`unwrapRustItemWrapperMacros.func1`, exit 2.

Truncating here leaves a `cfg_*!` wrapper nested past the limit unexpanded — the
same degradation as before this pass existed — and the file is still reported,
because the entity walk over the same source hits its own guard.

### Two more of the same class, found by the survey

`rAssignedValueKind` follows an R assignment chain (`a <- b <- function() 1`) whose
length the source controls. `jsPatternBindingNames` is reached from
`walkEntitiesScoped` through `jsEntityParameterNames` on every JS/TS callable, and
object patterns nest as deeply as they are written. Both abort the process from
`ParseWithStatus` on the parent commit. `jsFunctionLikeNode` is the sibling descent
on the same `jsEntityParameterNames` call; no input was found that drives it deep
(a declarator whose arrow function is buried under a million parentheses stops
being classified as a callable), but leaving one of the pair unbounded would only
relocate the abort if that classification ever changed.

### Finding — `identifierDescendants`, the walker this PR argued was safe

Raised against `parser.go:3785` after `main` was merged in, and **reproduced**.
`walkEntitiesScoped` checks its depth budget on entry and then, for every callable
it extracts, calls `astParameterNames` — which follows a parameter's `pattern` /
`declarator` field through `identifierDescendants` **before** the walk descends
into that callable. The walk's budget is therefore unspent and cannot bound this
descent, exactly like `firstNameDescendant`.

The claim this PR made for leaving it alone — that three C declarator shapes
(parenthesized, pointer, array) nested to 800,000 levels never drove it deep — was
a real measurement and the wrong conclusion. **C never takes that branch at all.**
`parameterBindingNames` splits on the field name: C's `parameter_declaration`
resolves through the `name` branch, which does not recurse, so C reads 0 levels at
any nesting and no amount of C could have falsified the claim. Rust and Python
parameter *patterns* take the `pattern` branch. Instrumenting the walker directly
on a 2,000-level fixture:

| fixture | levels reached |
|---|---|
| Rust tuple patterns — `fn f(((…a…)): u8) {}` | 2,001 |
| Rust reference patterns — `fn f(&&&…&a: u8) {}` | 2,001 |
| Python tuple patterns — `def f(((…a…))): pass` | 2,001 |
| C declarators — `void f(int ((…*b…))) { }` | **0** |

One level per source level, and a reference pattern is one byte per level. The
abort, on this branch's head before the fix, from `ParseWithStatus` on a 0.5 MB
file — no escape hatch, no size beyond the default cap needed:

```
$ go test -c -o sem.test ./internal/sem
$ EG_SEM_DEPTH_WALK_CHILD=1 ./sem.test -test.run '^TestParameterPatternDescentIsBoundedNotFatal$'
runtime: goroutine stack exceeds 25165824-byte limit
fatal error: stack overflow
github.com/entireio/entire-graph/internal/sem.identifierDescendants(...)   x hundreds of frames
EXIT=2
```

The fix bounds the recursive walker itself rather than either call site, so it
covers both entry points — `astParameterNames` and `directParameterChildren`,
Swift's list-less form — instead of one path into it. The budget starts fresh, as
`firstNameDescendant`'s does: the descent begins at one parameter entry and only
looks downward, so unlike `initializerTypeBodies` (whose result the entity walk
then descends *into*) it cannot compound with the caller's remaining budget.

After: exit 0, the file reported as a partial failure, the shallow symbol kept.
`TestDeepParameterPatternStillReportsTruncation` pins that a 6,000-level pattern
still yields `E_PARSE_DEPTH_EXCEEDED` with `Partial` true and the callable `f`
present — truncating the parameter descent does not turn a partial result into a
total failure, and does not swallow the truncation either.

**Re-survey.** The mechanical enumeration was re-run over the package call graph at
this head, after `main` was merged in, in case that merge brought new walkers: nine
self-recursive `*sitter.Node` functions and three recursive closures over one, no
mutually recursive node cycle. Every one is now bounded, and
`identifierDescendants` was the only one still reachable ahead of a guard.

### Relation phase — `jsScopeWalker.walk` and `jsMemberChainParts`

Raised in review against the entity-walk limit: relation construction **reparses**
each JS/TS file and walks its own tree *after* the entity phase has returned, so
the entity walk's budget never reaches it. Verified rather than accepted, and the
finding is half right.

**Not reproducible under the default cap.** The largest `.js` the provider's 4 MiB
`defaultMaxParseBytes` admits — 4,008,013 bytes, 2,000,000 nested parentheses —
completes both profiles with the shipped binary:

```
$ eg snapshot --repo /tmp/jsmax --profile full   EXIT=0   2.0 GB RSS
$ eg snapshot --repo /tmp/jsmax --profile fast   EXIT=0   2.0 GB RSS
"stats":{"files":3,"parsed_files":3,"partial_failures":2,"completeness_level":"unsafe"}
```

**Reproducible through the documented escape hatch**, which is the same route this
PR already cites for budgeting the entity-phase walkers —
`ProviderSnapshotOptions.MaxParseBytes < 0` removes the cap. A snapshot over an
8 MB `.js` file of 4,000,000 nested parentheses, at Go's default 1 GiB ceiling:

```
fatal error: stack overflow
github.com/entireio/entire-graph/internal/sem.(*jsScopeWalker).walk(...)   x85 frames
EXIT=2      3392847872 maximum resident set size
```

Both relation-phase walkers are unbounded and both are now bounded:

| walker | aborts at (16 MiB ceiling) |
|---|---|
| `jsScopeWalker.walk` | survives 100,000, aborts by 400,000 |
| `jsMemberChainParts` | aborts at 100,000 links, where `walk` still survives — `walk` hands it the whole member chain from its head *before* descending, so its recursion reaches the stack first |

Truncation reports through the channel the relation phase already has: a new
`jsScanState.depthTruncated` flag becomes an `E_PARSE_DEPTH_EXCEEDED` partial
failure carrying the entity phase's code and severity, so a file truncated in both
phases stays one record per code+file. The scope state stays usable — the
model is unchanged, a truncated walk is a partial result — with one exception:
`jsMemberChainParts` refuses a truncated chain rather than returning a short one,
because a partial chain would resolve the call to the **wrong** target.

#### Follow-up — one record, but not one phase's story

Raised in review against the record above and reproduced: sharing the entity
phase's code made `mergePartialFailures` **discard** the relation-phase record,
because it deduplicated by code+file and kept the record already present. The two
phases lose different things, so the report understated what was missing. On the
parent commit, one `.ts` file that truncates in both phases (a namespace call plus
a 6,000-deep nested initializer, verified truncating in both by asserting
`ParseWithStatus` status and `jsScanState.depthTruncated` before the snapshot):

```
before:  E_PARSE_DEPTH_EXCEEDED deep.ts
  effect=file record and symbols above the parser depth limit emitted; more deeply
         nested declarations were not walked, so this file counts against completeness
                                                       <- nothing about call classification

after:   E_PARSE_DEPTH_EXCEEDED deep.ts
  effect=... so this file counts against completeness; also: relation-phase scope walk
         truncated at the parser depth limit; calls nested deeper than that were not classified
  detail=AST nesting exceeded the 5000-level walk limit; declarations nested deeper than
         that were not extracted; also: AST nesting exceeded the 5000-level walk limit
         during relation construction
```

Still **one** record, deliberately: two records for one file would count that file
twice against completeness and list the same condition twice. A duplicate is now
**folded** instead of dropped — identity fields (code, file, severity) stay with
the record already present, and the incoming record's effect and detail text are
appended. Folding skips text the record already carries, so the cache-derived
selective path merging the same relation failures again is idempotent. The same
fold applies to `E_PARSE_TIMEOUT`, which loses symbol parsing in the entity phase
and call classification in the relation phase for exactly the same reason.

Failing-first at runtime on the previous head (`af2488d`), both tests compiled
against it unmodified:

```
--- FAIL: TestSnapshotReportsBothPhasesOfDepthTruncation (0.08s)
    relation-phase effect discarded by the merge; want "relation-phase scope walk
    truncated at the parser depth limit; calls nested deeper than that were not
    classified" inside "file record and symbols above the parser depth limit
    emitted; more deeply nested declarations were not walked, so this file counts
    against completeness"
--- FAIL: TestMergePartialFailuresFoldsRatherThanDrops (0.00s)
    effect "declarations below the limit were not walked" missing "relation-phase
    scope walk truncated at the parser depth limit; calls nested deeper than that
    were not classified"
```

`TestStreamSnapshotDedupesSameCodeFailuresAcrossPhases` was updated in the same
commit: it still requires one record with the entity phase's identity, and now
also requires that record to carry both phases' effect text.

Failing-first, against the parent commit:

```
--- FAIL: TestJSScopeWalkerIsBoundedNotFatal (0.71s)
    child process ... did not complete: exit status 2
        fatal error: stack overflow
        github.com/entireio/entire-graph/internal/sem.(*jsScopeWalker).walk(...)
--- FAIL: TestJSMemberChainIsBoundedNotFatal (0.61s)
    child process ... did not complete: exit status 2
        fatal error: stack overflow
        github.com/entireio/entire-graph/internal/sem.jsMemberChainParts(...)
```

`TestSnapshotOverDeeplyNestedJSCompletes` is a guard, not a fix pin — it runs the
whole provider over a truncating `.js` and requires the run to finish and say so.
It passes on both sides at that size, and is labelled as such in the test.

## Findings 3 and 4 — truncation is a PARTIAL RESULT

The two findings pull opposite ways: the diff treated truncation as too **fatal**,
completeness as too **benign**. It is neither. A depth-truncated file *parsed*, and
the declarations it did yield are *real* — so it is a partial result, and the two
existing categories both misdescribe it:

- not a **total failure**, because the output is valid, merely short;
- not an **intentional skip**, because unlike `E_FILE_TOO_LARGE` / `E_MINIFIED` the
  parser *did* open the file and *did* drop declarations. There is a genuine gap in
  understanding, and completeness exists to report exactly that.

`ParseStatus.Partial` names the distinction, and the three consumers now agree:

**(a) Analyze's file delta — kept.** A truncated walk yields `ParseError` with zero
entities, the shape `analyze.go` treats as total and answers by suppressing the
file's delta. Wrong here: the tree parsed and the walk stopped, so zero entities
means "no declarations above the limit", not "no signal". Reviewer's example, both
binaries, same repo:

```
=== parent commit ===                       === this branch ===
No semantic entity changes detected.        mod.py (Python)
                                              ~ module mod.py body changed
Warnings                                    Warnings
  ! E_PARSE_DEPTH_EXCEEDED mod.py             ! E_PARSE_DEPTH_EXCEEDED mod.py
    file diff suppressed; changes               diff kept, but declarations nested
    omitted because the file could              deeper than the parser walk limit
    not be parsed                               were not compared on one side
```

A genuinely unparseable file is *not* `Partial`, so it still suppresses its delta
(`TestAnalyzeStillSuppressesTotalParseFailure`).

**(b) Completeness — counted.** `E_PARSE_DEPTH_EXCEEDED` leaves
`intentionalSkipFailureCodes`. Ten Python files each carrying a shallow symbol above
6,000 nested parentheses — the reviewer's scenario exactly:

```
=== parent commit ===   {"files":10,"parsed_files":10,"partial_failures":10,"completeness_level":"ok"}
=== this branch ===     {"files":10,"parsed_files":10,"partial_failures":10,"completeness_level":"unsafe"}
```

### Interaction with the existing total-parse-error path

Raised in review against the `Partial` model itself, and reproduced. A file that
both exceeds the walk limit **and** contains syntax errors took the depth branch
unconditionally, masking `root.HasError()` and marking a malformed tree partial.
Base `def alpha` / `def beta`, head 6000 nested parentheses around a stray `$`:

```
parent:  entities=0  {ParseError:true Partial:true Code:E_PARSE_DEPTH_EXCEEDED}
         AnalyzeGitRange -> removed alpha, removed beta      <- phantom
fixed:   entities=0  {ParseError:true Partial:false Code:E_PARSE_ERROR}
         AnalyzeGitRange -> no changes; "file diff suppressed"
```

**Error dominates depth.** Truncation alone yields fewer entities that are still
correct; a malformed tree yields entities that may be *wrong*, so the caller must
stay free to suppress the file. When both hold the status is `E_PARSE_ERROR` with
`Partial` false, and the detail names both conditions. A deep but well-formed tree
is unchanged. Pinned by `TestDeepAndMalformedIsTotalNotPartial` (runtime-fails on
the parent with the phantom removal) and `TestDeepButWellFormedStaysPartial`.

**(c) Status code — unchanged**, `E_PARSE_DEPTH_EXCEEDED`, warning severity, now
carried alongside `Partial` so both consumers can tell a partial result from a
total failure without string-matching the code.

### The remaining phantom route — a WELL-FORMED tree that truncates to ZERO entities

Raised in review after the two fixes above. Both of those close the phantom-delta
class through a **malformed** tree. A well-formed one reaches it too: if the walk
truncates before it reaches any declaration, the side reports partial-and-usable with
**zero** entities, and `compareEntities` was handed an empty set against a populated
one. Base `function alpha` / `function beta` at top level, head the same two
declarations inside 6000 nested `{ }` blocks — a block is a statement and a function
declaration inside one is legal ES6, so there are no error nodes:

```
parent: head entities=0  {ParseError:true Partial:true Code:E_PARSE_DEPTH_EXCEEDED}
        AnalyzeGitRange -> removed alpha, removed beta   <- phantom; both still declared
fixed:  same status; entity comparison skipped
        AnalyzeGitRange -> module-scope change only + "entity comparison skipped"
```

The declarations are not missing from the head file, only hidden below the limit, so
every one of those removals is a phantom. `reconcileMoves` can then promote them into
cross-file MOVES, so the blast radius is not confined to the one file.

**Three depth cases stay deliberately distinct:**

| side | entities recovered | what the diff does |
|---|---|---|
| truncated | zero | entity comparison skipped; module-scope change still reported (new) |
| truncated | some | entity diff kept — fewer entities, but each one correct |
| not truncated | zero | entity diff kept — the removals are **real** |

The third row is the one worth being careful about: collapsing it into the first would
silently stop reporting genuine deletions, which is worse than the bug being fixed.
Each row has its own test, and the zero-entity fixture asserts its preconditions
(head: well-formed, `E_PARSE_DEPTH_EXCEEDED`, `Partial`, zero entities; base: clean
parse with `alpha` and `beta` present) so it cannot silently stop exercising the path.

Only an existing side can lack signal this way: on an add or a delete the missing side
parses no content, and its emptiness is the real change.

Because the diff now does three distinct things with a degraded parse,
`parseFailureWarning`'s `suppressed bool` becomes a `diffParseOutcome`
(`diffKept` / `diffEntitiesSkipped` / `diffSuppressed`). The skipped case's effect
string states that the comparison was skipped and that individual symbol changes are
omitted, rather than claiming a complete diff.

## The limit: `maxParseWalkDepth = 5000`, unchanged and re-verified

Bounding more walkers changes which walkers count depth, so the corpus evidence was
re-run rather than inherited. Every walk above was instrumented with a per-walker
max-depth probe and **56,261 real source files** were parsed through it, two corpora
measured separately (percentiles are per corpus):

| corpus | files | p50 | p90 | p99 | p99.9 | max | truncated at 5000 |
|---|---|---|---|---|---|---|---|
| first-party (entire-graph, cli, entire-api, entiredb, entire.io/{frontend,api}/src, entire.io/website) | 7,515 | 14 | 23 | 33 | 46 | **463** | **0** |
| third-party (entire.io/node_modules) | 48,746 | 10 | 21 | 37 | 184 | **418** | **0** |

Deepest level each walk actually reached, across both corpora:

```
walkEntitiesScoped     463    entiredb gen/proto/publicapi/admin/v1/admin.pb.go (generated protobuf)
initializerTypeBodies  359    lightningcss/node/ast.d.ts
collectParseErrorDetails 310  @babel/types/lib/index-legacy.d.ts
jsFunctionLikeNode      12    jsPatternBindingNames 8    firstNameDescendant 7    firstDescendantOfType 2
```

The name-resolution descents added here start at a declaration and only look
downward, so they are an order of magnitude shallower again than whole-file
nesting. 5000 caps the walk near 3.8 MB of stack (`walkEntitiesScoped` frames
measured at 752 bytes on darwin/arm64) against Go's 1 GB goroutine limit.

## Tests

Every test in `parser_depth_test.go` asserts on **behaviour**, never on the new
`ParseStatus.Partial` field, so the file compiles unmodified against the parent
commit and each pre-fix failure is a **runtime** failure rather than a build error.
A fatal stack overflow cannot be caught by `recover()`, so those tests re-exec the
test binary as a child process under an env guard and assert on its **exit status**;
the child lowers its goroutine stack ceiling with `debug.SetMaxStack` so the same
unbounded recursion aborts with a ~1 MB fixture instead of a 12 MB one and a few
hundred MB of RSS instead of 4.1 GB. That is honest because the defect is unbounded
recursion, not a byte count: an unguarded walker overflows whatever ceiling it is
given, and a guarded one stops at 5000 levels under any ceiling above ~4 MB.

Verbatim, against the parent commit:

```
--- FAIL: TestNameDescentIsBoundedNotFatal (0.32s)
    child process running TestNameDescentIsBoundedNotFatal did not complete: exit status 2
        fatal error: stack overflow
        recursing frames:
        github.com/entireio/entire-graph/internal/sem.firstNameDescendant(0x7f3af64b29f0, ...)
--- FAIL: TestRustMacroUnwrapIsBoundedNotFatal (0.22s)
        fatal error: stack overflow
        github.com/entireio/entire-graph/internal/sem.unwrapRustItemWrapperMacros.func1(0x5d9498c11b60)
--- FAIL: TestRAssignmentChainIsBoundedNotFatal (0.35s)
        fatal error: stack overflow
        github.com/entireio/entire-graph/internal/sem.rAssignedValueKind(0x1b2e1ae7e090, ...)
--- FAIL: TestJSPatternBindingsAreBoundedNotFatal (0.70s)
        fatal error: stack overflow
        github.com/entireio/entire-graph/internal/sem.jsPatternBindingNames(0x48b1498aea80, ...)
--- FAIL: TestDepthTruncationCountsTowardCompleteness (0.00s)
    completenessFailureCount = 0, want 1: a depth-truncated file is a real coverage gap
--- FAIL: TestSnapshotReportsDepthTruncationAndCountsItAgainstCompleteness (0.05s)
    completeness_level = "ok", want degraded: one of ten files was parsed incompletely
--- FAIL: TestAnalyzeKeepsDeltaForDepthTruncatedFile (0.11s)
    depth truncation must not suppress the file delta; got files []
    warnings [{Code:E_PARSE_DEPTH_EXCEEDED ... EffectOnCompleteness:file diff suppressed;
    changes omitted because the file could not be parsed ...}]
```

`TestAnalyzeSkipsEntityDiffWhenTruncatedSideIsEmpty` is demonstrated differently, and
the difference is stated rather than glossed: this change also turns
`parseFailureWarning`'s `suppressed bool` into a `diffParseOutcome`, which the sibling
`analyze_parsefailure_test.go` calls directly, so the package does not build against the
parent commit unmodified. Instead the **behavioural guard alone** was neutralised — the
`if !entitiesSkipped` around `compareEntities` forced back to unconditional, everything
else in place — which is exactly the parent's behaviour at that call site:

```
--- FAIL: TestAnalyzeSkipsEntityDiffWhenTruncatedSideIsEmpty (0.16s)
    phantom removed "alpha" from a truncated zero-entity side (svc.js): both are still
    declared in head, below the walk limit; changes [{Type:removed Kind:function
    Name:alpha ... BeforeStartLine:1 AfterStartLine:0 ...} {Type:removed Kind:function
    Name:beta ... BeforeStartLine:2 AfterStartLine:0 ...}]
```

`TestAnalyzeDiffsLegitimatelyEmptyFileNormally` (not truncated, genuinely zero entities)
and `TestAnalyzeKeepsEntityDiffWhenTruncatedSideHasEntities` (truncated, some entities)
pass with and without that guard: they pin that the fix does not over-suppress, which is
the failure mode that would be worse than the bug.

Guards that pass **both** before and after — they pin the absence of a regression,
not the fix: `TestParseNestingUnderTheWalkLimitIsNotTruncated`,
`TestNameDescentUnderTheWalkLimitStillNamesEntities`,
`TestDeepNameDescentStillReportsTruncation`,
`TestCollectParseErrorDetailsStillReportsShallowErrors`,
`TestAnalyzeStillSuppressesTotalParseFailure`.

`TestDeepNameDescentStillReportsTruncation` earns its place: the name descents
truncate **silently** (they return "no name found" rather than threading a flag
back), which is only safe because any subtree deep enough to exhaust their budget
is also descended by `walkEntitiesScoped` or `initializerTypeBodies`, whose guard
sets the status. That test pins the assumption instead of trusting it.

Fixtures are pure in-memory source strings — no symlinks, device files, exotic
names or separator assumptions — and the child re-exec uses `os.Args[0]`, so they
behave identically on `windows-latest` and need no `runtime.GOOS` guard.

## Gate

```
gofmt -l -s .    clean
go vet ./...     ok
go build ./...   ok
go test ./...    ok (7 packages with tests, 0 failures) -- re-run at 85b4f3b
go test -race ./internal/sem -run '<the depth + partial-failure tests>'
                 ok, 29/29 PASS, no data race
```

## Scope note for the parallel PRs in this series

`internal/sem/provider.go` is touched in exactly one region, the
`intentionalSkipFailureCodes` map literal. `internal/sem/provider_parallel.go` only
in the partial-failure translation. `internal/sem/analyze.go` only in the
total-vs-partial branch and `parseFailureWarning`. `internal/sem/js_scopes.go` is
touched deliberately: the two helpers `jsEntityParameterNames` calls, plus
`jsScopeWalker.walk` and `jsMemberChainParts` after review reproduced aborts in
both. `internal/sem/provider.go` gains two regions beyond the completeness map —
the relation-phase failure record next to the existing `jsScanPartialFailure`
call, and `mergePartialFailures` plus its three helpers, which the record above
made necessary. `internal/sem/search_cache.go` is comment-only. Leaving a
reproduced fatal abort — or a reproduced silent loss in the report that abort's fix
introduced — reachable from the entry point this PR is fixing would make the PR's
own claim untrue.


suhaanthayyil:
Responses to the two review findings.

## HIGH, `parser.go:365` — relation-phase walkers (addressed in 873f149)

Reproduced, with one correction to the finding's reachability claim.

**Refuted under the default cap.** The largest `.js` the provider's 4 MiB
`defaultMaxParseBytes` admits — 4,008,013 bytes, 2,000,000 nested parentheses —
completes both profiles with the shipped binary:

```
snapshot --profile full   EXIT=0   2.0 GB RSS
snapshot --profile fast   EXIT=0   2.0 GB RSS
"stats":{"files":3,"parsed_files":3,"partial_failures":2,"completeness_level":"unsafe"}
```

**Confirmed through `ProviderSnapshotOptions.MaxParseBytes < 0`**, the documented
escape hatch that removes the cap — the same route this PR already cites for
budgeting the entity-phase walkers. Snapshot over an 8 MB `.js` of 4,000,000
nested parentheses, Go's default 1 GiB ceiling:

```
fatal error: stack overflow
internal/sem.(*jsScopeWalker).walk(...)   85 frames
EXIT=2   3392847872 maximum resident set size
```

So the mechanism is exactly as described; only the entry point differs. Both
relation-phase walkers were unbounded and both are now bounded:

| walker | aborts at (16 MiB ceiling) |
|---|---|
| `jsScopeWalker.walk` | survives 100,000, aborts by 400,000 |
| `jsMemberChainParts` | aborts at **100,000 links, where `walk` still survives** — `walk` hands it the whole member chain from its head *before* descending, so its recursion reaches the stack first |

Thank you for naming the member-chain helper separately; it is the one that
aborts first, and bounding only `walk` would have moved the crash rather than
removed it.

Bounding was chosen over skipping the scan for depth-truncated files, for the
reason given: a file can be deep in a way that trips the scope walker without
tripping the entity walk. Truncation reports through the relation phase's
existing channel (`jsScanState.depthTruncated` -> `E_PARSE_DEPTH_EXCEEDED`
partial failure). `jsMemberChainParts` refuses a truncated chain rather than
returning a short one, because a partial chain would resolve the call to the
wrong target.

## MEDIUM, `parser.go:433` — malformed + deep masked as partial (fixed in af2488d)

Reproduced exactly as described. Base `def alpha` / `def beta`, head 6000 nested
parentheses around a stray `$`:

```
before:  entities=0  {ParseError:true Partial:true Code:E_PARSE_DEPTH_EXCEEDED}
         AnalyzeGitRange -> removed alpha, removed beta      <- phantom
after:   entities=0  {ParseError:true Partial:false Code:E_PARSE_ERROR}
         AnalyzeGitRange -> no changes; "file diff suppressed"
```

Fixed as suggested: error dominates depth. Truncation alone yields fewer entities
that are still correct; a malformed tree yields entities that may be wrong, so
the caller must stay free to suppress the file. When both conditions hold the
status is `E_PARSE_ERROR` with `Partial` false, and the detail names both so the
operator is not left guessing which to fix. A deep but well-formed tree is
unchanged and still degrades to a partial result.

This was a regression the `Partial` model introduced, so it belonged here. Pinned
by `TestDeepAndMalformedIsTotalNotPartial`, which fails at runtime on the parent
with the phantom removal, and `TestDeepButWellFormedStaysPartial` for the other
half of the pair.

## Remaining disclosure

`parameters.go:285 identifierDescendants` is the only recursive AST walker left
unbounded. Three C declarator shapes (parenthesized, pointer, array) nested to
800,000 levels never drove it deep, so it is documented at `maxParseWalkDepth`
rather than guarded on a hypothesis. If someone finds the shape that reaches it,
that is a real bug and I would want to know.

Gate on af2488d: `gofmt -l -s` clean, `go vet ./...` ok, `go build ./...` ok,
`go test ./...` ok (7 packages).

ashtom:

- **11 commits ahead, 219 commits behind** current `main`.
- Test-merging onto `main` **conflicts in 1 file(s)**:

```
internal/sem/analyze.go
```

### Note

I could not find a recursion bound in `main`'s `internal/sem` (`maxWalkDepth`, `recursionLimit` and
similar all return nothing), so the CWE-674 concern this addresses appears to still be open on
`main`. I did not build a reproduction, so treat that as an unverified indication rather than a
finding.

### What this needs

### Entire trail findings

### Verified vs. assumed

suhaanthayyil:
Merged `main` (`bb41f223`) into this branch and re-verified. Head is now `bff5056`.

## Is CWE-674 still open on `main`?

Yes — confirmed, not just indicated. There is no recursion bound anywhere in `internal/sem` on
`bb41f223`:

```
$ grep -rn 'maxParseWalkDepth\|maxWalkDepth\|recursionLimit\|DepthExceeded\|E_PARSE_DEPTH' internal/
(no matches)
```

and the walkers still take no depth parameter:

```
internal/sem/parser.go:3643      func walkEntitiesScoped(node *sitter.Node, src []byte, language, scope string, inFunc bool, entities *[]Entity)
internal/sem/parser.go:382       func collectParseErrorDetails(root *sitter.Node, src []byte, limit, lineOffset int) []string
internal/sem/parser.go:613       func unwrapRustItemWrapperMacros(src []byte) bool          // recursive closure inside
internal/sem/parser.go:6660      func firstNameDescendant(node *sitter.Node, src []byte) string
internal/sem/parameters.go:285   func identifierDescendants(node *sitter.Node, src []byte) []string
internal/sem/js_scopes.go:597    func jsPatternBindingNames(pattern *sitter.Node, src []byte) []string
internal/sem/js_scopes.go:646    func jsMemberChainParts(node *sitter.Node, src []byte) []string
```

Reproduced as a fatal process abort against `bb41f223` with a temporary in-package test that
re-executes itself as a child with `debug.SetMaxStack(24 MiB)` and 500,000 nesting levels
(~1 MB of source, under the 4 MiB parse cap, line-broken every 500 columns so `looksMinified`
does not fire):

```
$ go test ./internal/sem/ -run TestZZRepro140 -count=1 -v
--- FAIL: TestZZRepro140WalkEntities              child exit=exit status 2   fatal error: stack overflow
--- FAIL: TestZZRepro140RustMacroUnwrap           child exit=exit status 2   fatal error: stack overflow
--- FAIL: TestZZRepro140CollectParseErrorDetails  child exit=exit status 2   fatal error: stack overflow

recursing frames in the child dumps:
  github.com/entireio/entire-graph/internal/sem.walkEntitiesScoped                (x10)
  github.com/entireio/entire-graph/internal/sem.unwrapRustItemWrapperMacros.func1 (x6)
  github.com/entireio/entire-graph/internal/sem.entityFromNode                    (x2)
```

Inputs: `x = ((((…1…))))` as `deep.py`; a `cfg_net! { … }` header plus nested parens as
`deep.rs`. The third fixture (unbalanced parens, intended for `collectParseErrorDetails`)
also aborts, but in `walkEntitiesScoped` first, since the entity walk runs before the
error-detail walk — so that particular walker is not *separately* isolated by this repro. It
is isolated by this branch's own `TestCollectParseErrorDetailsIsBoundedNotFatal`, which is one
of the tests the mutation below flips.

## What the merge required

One file conflicted, `internal/sem/analyze.go`, one hunk — purely additive on both sides, at
the same insertion point after `budgetSkippedFileWarning`. `main` added three warning
constructors (`diffFileTooLargeWarning`, `diffNonBlobWarning`, `diffFileReadWarning`); this
branch added the `diffParseOutcome` type and its three constants. Both kept, `main`'s
constructors first, this branch's type immediately above `parseFailureWarning`, which consumes
it. Nothing was dropped from either side.

Everything else auto-merged: `internal/sem/analyze_test.go`, `internal/sem/provider.go`,
`internal/sem/provider_parallel.go`, `internal/sem/search_cache.go`.

## Does the bound still cover every walker after 219 commits of `main`?

Checked mechanically on the merged tree rather than by eye — enumerated every self-recursive
function whose signature takes a `*sitter.Node`, plus every `var x func(… *sitter.Node …)`
recursive closure, across all non-test files in `internal/sem`:

```
firstDescendantOfTypeAt  firstNameDescendantAt  identifierDescendantsAt
jsFunctionLikeNodeAt     jsMemberChainPartsAt   jsPatternBindingNamesAt
rAssignedValueKindAt     walkEntitiesScoped     jsScopeWalker.walk
+ 3 recursive closures in parser.go, all now func(*sitter.Node, int)
```

Every one carries a depth parameter and tests `depth >= maxParseWalkDepth`; `main` introduced
no new unbounded parse-tree walker. (`signatureFromNode` shows up in a naive scan but is not
self-recursive.) No tree-sitter walking exists outside `internal/sem`.

## Re-verification on the merged tree

```
$ go build ./...                 # clean
$ go vet ./internal/sem/         # clean
$ gofmt -l internal cmd          # empty
$ go test ./internal/sem/ -run '^(<the 31 tests in parser_depth_test.go>)$' -count=1 -v
31 passed, 0 failed   (14.9s)

$ go test ./internal/sem/ -run 'CompareEntities|ParseFailure|ParseError|Analyze|TypeScript|Parse'   ok  14.8s
$ go test ./internal/sem/ -run 'Snapshot|Partial|Completeness|Warning|Scope|Parameter|Member'       ok  20.1s
```

The second and third cover the signature-adaptation edits this branch makes to
`analyze_parsefailure_test.go`, `analyze_test.go` and `typescript_test.go`
(`parseFailureWarning(..., diffSuppressed)`, `compareEntities(..., false)`).

## Mutation check

Reverted the production change on the merged tree by making the limit unreachable —
`const maxParseWalkDepth = 5000` → `1 << 40`, so no `depth >= maxParseWalkDepth` test can ever
fire while every call site keeps compiling. Confirmed applied by grep
(`parser.go:208: const maxParseWalkDepth = 1 << 40`) and `go build ./...` clean.

22 of the 31 tests failed:

```
TestParseDeeplyNestedSourceIsBoundedNotFatal            TestNameDescentIsBoundedNotFatal
TestRustMacroUnwrapIsBoundedNotFatal                    TestRAssignmentChainIsBoundedNotFatal
TestJSPatternBindingsAreBoundedNotFatal                 TestParameterPatternDescentIsBoundedNotFatal
TestJSScopeWalkerIsBoundedNotFatal                      TestJSMemberChainIsBoundedNotFatal
TestCollectParseErrorDetailsIsBoundedNotFatal           TestSnapshotOverDeeplyNestedJSCompletes
TestDeepParameterPatternStillReportsTruncation          TestDeepNameDescentStillReportsTruncation
TestInitializerTypeBodiesWalkIsBudgeted                 TestParseTruncatedWalkStillEmitsShallowSymbols
TestSnapshotReportsDepthTruncationAndCountsItAgainstCompleteness
TestSnapshotReportsBothPhasesOfDepthTruncation          TestDeepButWellFormedStaysPartial
TestAnalyzeKeepsDeltaForDepthTruncatedFile              TestAnalyzeSkipsEntityDiffWhenTruncatedSideIsEmpty
TestAnalyzeKeepsEntityDiffWhenTruncatedSideHasEntities
TestAnalyzeSuppressesCombinedDepthAndMalformedEvenWithRecoveredEntities
TestAnalyzeDoesNotSuppressAddedFileEntitiesWhenTheOnlySideIsDepthTruncated
```

The 9 that still passed are exactly the controls that must not depend on the bound — the
under-the-limit cases (`TestParseNestingUnderTheWalkLimitIsNotTruncated`,
`TestNameDescentUnderTheWalkLimitStillNamesEntities`,
`TestCollectParseErrorDetailsStillReportsShallowErrors`), the ordinary-diff cases
(`TestAnalyzeDiffsLegitimatelyEmptyFileNormally`, `TestAnalyzeStillSuppressesTotalParseFailure`,
`TestDeepAndMalformedIsTotalNotPartial`, `TestMergePartialFailuresFoldsRatherThanDrops`,
`TestAnalyzeSuppressesOneSidedResultWhenTruncatedSideHasEntities`,
`TestDepthTruncationCountsTowardCompleteness`) — so the suite is not simply keyed on the
constant.

Restored to `5000`: 31 passed. No test was changed, skipped or weakened in this branch's merge.

suhaanthayyil:
Brought current with `main` (`7b3252aa`); the Up-to-date gate was 13 behind, then 9 more landed mid-round. Head is `4cc4283a`.

The first round conflicted in `internal/sem/analyze.go`, where main's rename work and this branch's depth-truncation work landed on the same lines. Both are additive and independent, so both are kept.

The parse-failure test keeps this branch's form:

```go
afterParseFailed := afterStatus.ParseError && !afterStatus.Partial && (len(afterEntities) == 0 || afterStatus.DepthExceeded)
```

Main's `ParseError && len(entities) == 0` does not know about `DepthExceeded`, and taking it would reopen both cases this branch closed — a depth-truncated side with zero entities treated as a total failure, and a side that is both malformed and truncated treated as partial. Main's explanatory comment is kept alongside; it describes the general rationale and is still accurate.

Main's unsupported-side emptying moved down beside the existing missing-revision emptying, which is the same operation on the same values, rather than staying interleaved with the truncation analysis. The two do not interact: a side with no parser carries no `ParseStatus.Partial`, so the truncation checks read nothing from it either way.

The second round merged clean.

| check | result |
|---|---|
| `go build ./...`, `go vet`, `gofmt -l` | clean |
| this branch's 39 tests (`parser_depth_test.go` + `analyze_parsefailure_test.go`), by exact name | 39/39 pass |
| `internal/sem` `TestAnalyze*/TestRename*/TestMixedSupport*/TestUnsupported*` (main's new area) | ok |
| mutation: `maxParseWalkDepth` 5000 → 1000000 (compiles, grep-confirmed) | 8 bounded-not-fatal tests **FAIL**; restore → pass |

The mutation was re-run on the merged tree rather than relied on from before: the bound is still load-bearing after both rounds.

ashtom:

**Verdict: clean gate — my previous comment is superseded**

### Correction

My last comment on this PR said it was stale (219 commits behind) and could not be evaluated without
a manual merge. That is no longer true: the branch has been brought up to date and is now **0 commits
behind** and merges clean. Please disregard the "needs a merge" part of that comment.

| Check | Tree | Result |
|---|---|---|

### Entire trail findings

Re-queried immediately before posting (`--status open --freshness current`): **none**.

suhaanthayyil:
Confirmed — nothing to change.

Re-verified against current `main`: all three `test` legs, all three `build` legs, and lint pass. Branch is 4 commits behind `main`; trial merge clean, and the merge result retains the files `main` gained today (the nightly workflow and its script) — the deletions visible in `git diff main..branch` are a two-dot-diff artifact of the branch predating #186, not a change this PR makes.

The recursion bound (`maxParseWalkDepth`) is still absent from `main`, so the fix this PR carries is still needed rather than superseded by anything that landed today.

ashtom (approved):
Approved after syncing with current main. Entire findings are clear, the merged-tree full suite passed locally, and required CI is completing on this head.

PR #146, opened 2026-08-21
RFD 0005: converge code search on one engine (P2.M1 / COR-786)

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/106

**RFC / discussion only. Not for merge yet.** This is a decision doc, no code changes. It opens P2 (Unified Memory) with P2.M1 / COR-786: "converge code search on one engine, deprecate duplicates."

Doc: `docs/adr/0004-converge-code-search.md` (continues this repo's existing `docs/adr/` convention; Status is Proposed).

### The proposal in three lines

**entire-graph becomes the precision symbol PRODUCER**, emitting a SCIP index per `repo@commit` into peregrine's already-reserved `SymbolSource::Scip` / `SymbolPrecision::Semantic` slots. **peregrine remains the find-text-fast serving engine** and the single front door. **entire-graph remains the understand-structure engine** (relations, impact, brain). Each side then retires its duplicate half.

### What the doc found (grounded, file:line in the appendix)

- **The seam is designed-in, not invented here.** `peregrine/src/symbols/types.rs:111-115` and `:118-122` reserve `SymbolSource::{TreeSitter,Scip}` and `SymbolPrecision::{Syntax,Semantic}`; `:265-276` already deserializes them. peregrine's own spec says the fields "exist to support future SCIP overlay without schema changes." Nothing constructs `Scip` today. No `.pg` version bump needed to fill it.
- **peregrine's persisted symbol resolution is a documented heuristic.** `resolve.rs:118` says so in its own words ("simple name-matching + same-directory preference"), and `:174-176` breaks ties by picking the first candidate. The better import-aware resolver exists but only runs per-request at `POST /api/resolve`, never persisted. The README describes only the second one.
- **Kotlin extraction is dead code in peregrine.** `parser.rs:27` drops the grammar on a tree-sitter version conflict, so `lang_kotlin.rs` is unreachable and every Kotlin file yields zero symbols, despite Kotlin being advertised tier 1. entire-graph extracts it today.
- **entire-graph has started building a second trigram engine.** Branch `codex/eg-graphify-lexical-sidecar` carries a dormant on-disk trigram-postings index (~1.4k lines, `EGLEXI1`/`EGPOST1`), zero callers. The RFD proposes closing it rather than merging.
- **entire-graph's semantic coverage is a strict superset**: 36 semantic languages vs peregrine's 15 tier-1, all 15 included.
- **Honest cons are recorded**, not hidden: SCIP is lossy (only 15 of 30 relation types map), the exported monikers carry entire-graph-internal identity rather than portable ones, PR #97 materializes the whole index in memory, and peregrine's S3 object is repo-keyed rather than commit-keyed, which the interface section has to reconcile.

### Sections to review

Options A/B/C with con lists, the interface (artifact, delivery, freshness contract, fallback), migration steps with owners, what deprecation concretely deletes and when, parity-harness gates, non-goals, and 8 open questions.

### Explicit non-goals

Embedding backbone convergence is **COR-788** and gets its own RFD. Prose and knowledge search (entire-search, librarian) are untouched. This is not an entire-graph deprecation, and it is not the MCP surface design (that is P2.M3).

### Asks

- @evisdren - you own peregrine and most of the migration steps. Are the ingestion seam, the fallback contract and the parity gates the right shape? Open questions 2 (moniker scheme), 3 (size/latency budgets) and 7 (feed store addressing) need your call. Also please sanity-check the Kotlin and dual-resolver findings.
- @alishakawaguchi - the front-door half. `search_entire_code` is already the live surface; does the P2.M3 direction here line up with what you are doing on the gateway plane?
- @ashtom - PR #97 is step 1. Does promoting `--format scip` out of experimental, and freezing the omission note as contract, seem reasonable?
- Alex (librarian) - mostly an FYI: this RFD deliberately does not touch prose/knowledge search. Flag it if you think the boundary is drawn wrong.

Open questions 1, 4, 5 (who owns the parity harness) are unassigned. 5 is the biggest scheduling risk.


ashtom:

**Verdict: changes required — 1 HIGH finding on the ADR itself**

### State

- 6 ahead, 25 behind `main`; **merges clean**. Draft.
- Adds `docs/adr/0005-converge-code-search.md`. Relevant to **#152**, which adds `0006` — the
  numbering only stays contiguous if this lands first.

| Check | Tree | Result |
|---|---|---|
| revert-the-fix check | — | N/A — docs-only PR |

### Note

The HIGH finding says assigning the object-store write to the entire-graph producer violates the
provider boundary. That is worth taking seriously against this repo's stated contract: `AGENTS.md`
requires the provider stay **no-egress** — "never add remote fetches, hosted API calls, telemetry".
An ADR that ratifies the producer writing to an object store is exactly the kind of decision that
should be settled in the ADR rather than discovered in review of the implementation.

### Entire trail findings — 4 open and current

Re-queried immediately before posting.

```
01M11YK8E7ZN  Medium  open    current    docs/adr/0005-converge-code-search.md:229  The required omission-note contract is frozen as an unversioned shape, leaving c…
01M11YK8DSNQ  High    open    current    docs/adr/0005-converge-code-search.md:183  Assigning the object-store write to the entire-graph producer violates the provi…
01M10SVGXRYZ  Medium  open    current    docs/adr/0005-converge-code-search.md:267  The no-regression gate compares against peregrine's native layer, but that basel…
01M10SVGXAVB  Medium  open    current    docs/adr/0005-converge-code-search.md:174  A 20-byte `indexed_commit` represents only SHA-1, while Entire Graph supports Gi…
```

### Verified vs. assumed

**Verified by execution:** merge onto `bb41f223`; `go vet`/`gofmt`; the full test run for every package this
PR touches, on the merged tree; trail findings re-queried immediately before posting.

suhaanthayyil:
Merged `origin/main` into the branch: clean, no conflicts, 25 commits of main absorbed. `MERGEABLE` at `5f62c7a4`; the ADR number `0005` is still free on main (`docs/adr/` holds `0001`–`0004`), so `0006` stays available for #152.

The merge did change the ground under the RFD: **PR #97 is merged** (2026-08-27), so `internal/sem/compact_snapshot.go` and `--format scip` are on main. Statements that the producer was unmerged, that #97 was open/draft, and that there was "No SCIP on main" were true when written and are false in the merged tree. Those are corrected in the same commit, along with the `capabilities --json` drift in the appendix, which main has since reconciled (live output now reports 185 supported / 36 semantic / 149 inventory-only / 30 relation types, matching `docs/language-support.md`).

---

### HIGH `01M11YK8DSNQ` — object-store write assigned to the producer — **CONFIRMED, fixed**

Reproduced. The finding is right and the ADR contradicted itself in three places.

The contract it violates is explicit — `AGENTS.md:224`:

> Contract rules that must not break: schema `1.x` is frozen and additive-only (`docs/adr/0001-ga-schema-contract.md`); the provider is **no-egress** (never add remote fetches, hosted API calls, telemetry, or runtime grammar downloads)

and `AGENTS.md:17`: "Built-in analysis is local and no-egress (no network, no model, no keys)."

The old Delivery paragraph said the artifact "lands in the same object store as the `.pg` index, **written by the producer**", and the TLDR names entire-graph as the producer. That is object-store credentials and network access in the graph binary. It also contradicted this RFD's own case against Option B, which rests on entire-graph being "a one-shot, no-egress binary, not a serving system".

Delivery is now split three ways, and the split is stated as a decision rather than left to implementation:

> **Delivery, and the no-egress boundary.** entire-graph does not write to the object store. The provider is no-egress by ratified contract — "never add remote fetches, hosted API calls, telemetry, or runtime grammar downloads" (`AGENTS.md:224`) — and this RFD's own case against Option B rests on it being "a one-shot, no-egress binary". A producer that had to hold object-store credentials and reach the network would spend exactly the property that makes its output trustworthy, and would change that boundary by implication in an implementation review rather than on purpose here. So the write is split three ways:
>
> - **entire-graph** runs `snapshot --format scip`, which writes the SCIP protobuf to stdout and the omission note to stderr (`internal/cli/root.go:271,445`). It has no file sink, no network client and no credentials, and needs none: both halves of the artifact leave over pipes.
> - **A publisher wrapper**, deployment-side and not the graph binary, captures those two streams and uploads them to the same object store as the `.pg` index, under the `(repo_id, commit_sha)` key above. It owns the credentials, retries, retention and the paging for a failed publish.
> - **The indexer tier** (`deploy/fleet/apps/peregrine/base/index-deployment.yaml`) reads the artifact at build time.
>
> Whether that wrapper is a standalone job or a step inside peregrine's indexer tier is open question 6. That it is *not* entire-graph is decided here.

The mechanism is checked against the merged code, not assumed: `--format scip` has no output-file flag, writes the index to stdout, and reserves stderr for the note (`internal/cli/root.go:271`, `:445`). The binary has nowhere to put bytes except its own pipes, so the boundary is enforced by the shape of the command.

Open question 6 was reworded to match — it asked where "the producer" runs, which is now settled; it asks where the *publisher* runs, which is not:

> 6. **Publisher runtime placement.** The producer itself is settled — entire-graph emits over pipes and never uploads (see *Delivery*). Open is where the publisher wrapper that captures and uploads those streams runs: a step inside peregrine's indexer tier, or a separate job. Affects failure isolation and paging.

---

### medium `01M11YK8E7ZN` — omission note frozen as an unversioned shape — **PARTLY REFUTED, gap fixed**

Refuted on the facts: the note is **not** unversioned. It has carried a version field since #97 — `internal/sem/compact_snapshot.go:299`:

```go
Version: "entire-graph-scip-omissions/v1",
```

The real gap is the one behind the finding: a bare `v1` has no minor, so a consumer cannot tell additive evolution from a break and cannot implement ADR 0001's rule 3 ("warn (not fail) when they see a newer supported-major minor"). The ADR only gestured at this, in Consequences, as "ADR 0001's additive-only `1.x` rules **should** extend to" the note. That is now a decision in *The interface*:

> Being contract, it is versioned like one. The note already carries a version field (`"version": "entire-graph-scip-omissions/v1"`, `internal/sem/compact_snapshot.go:299`), so "freeze the schema" in step 1 means freezing a compatibility **policy**, not a field list. ADR 0001's `1.x` rules (`docs/adr/0001-ga-schema-contract.md`) extend to the note unchanged: the major is the compatibility boundary and a consumer refuses an unknown one; minors are additive only, never removing a field, changing a field's meaning, or making an optional field required; readers ignore unknown fields and warn rather than fail on a newer minor within a supported major. The version tag therefore carries a minor as well — `entire-graph-scip-omissions/v1.0`, then `/v1.1` for each additive change — because a bare `v1` gives a tolerant reader nothing to warn on and no way to tell compatible evolution from a break. A breaking change is `/v2` with a migration note, and is out of scope for the feed's `v1` line.

Step 1's gate no longer reads as "freeze this field list", which was the trap:

> Freezing means stamping that shape `entire-graph-scip-omissions/v1.0` under the additive-only policy in *The interface*, not freezing it against all further additions

Consequences drops the "should".

---

### medium `01M10SVGXRYZ` — no-regression gate passes zero-to-zero on Kotlin — **CONFIRMED, fixed**

Reproduced against the ADR's own text. The primary metric is scored "against peregrine's native layer", the gate is "recall must not drop in any language", and the RFD itself establishes that peregrine's native layer emits nothing for Kotlin (`parser.rs:27` excludes the grammar, `extract.rs:93-96` yields no symbols). A feed emitting zero Kotlin symbols passes `0 → 0`. The existing ground-truth clause covers only cross-file `resolved_to`, not symbol recall, so it does not close this.

Added as a gate bullet:

> - **Absolute floor, because the baseline is not trustworthy everywhere.** Non-regression against peregrine's native layer is a floor of zero wherever that layer is empty or broken. Kotlin yields no symbols at all today (`parser.rs:27`, `extract.rs:93-96`), so a feed that emitted nothing for Kotlin would pass a zero-to-zero comparison and could reach feed-on without ever fixing the coverage gap this RFD cites as a motivation. Every language therefore also has to clear an **absolute recall floor against hand-labelled ground truth** on a sampled subset of the corpus, not only a delta against the native layer. A language whose native baseline is empty or known-degraded — Kotlin at minimum — may not go feed-on on the delta comparison at all; the ground-truth measurement is the only evidence that counts for it. The floor per language is agreed with Evis before the run, like the budgets.

Step 6's gate cell was tightened to match, so the table and the prose cannot drift: "Recall no worse than feed-off *and* above the per-language absolute floor, precision better, budgets held".

---

### medium `01M10SVGXAVB` — `indexed_commit: [u8; 20]` is SHA-1 only — **CONFIRMED, fixed**

Both halves verified. `indexed_commit` is 20 bytes (`src/index/format.rs:33-44`), and entire-graph does read SHA-256 repositories: `internal/gitutil/git.go:1529-1530` sizes tree-metadata records for 64-hex object ids ("SHA-256 object IDs need at most 64 bytes"), and `internal/gitutil/git_test.go:1582` branches on `rev-parse --show-object-format == "sha256"`. A 32-byte commit id cannot ride in that header, and the exact-commit join is the entire freshness contract, so the ADR could not stay silent.

Scoped rather than fudged — new paragraph next to the addressing contract:

> **Object-format scope: SHA-1 only, for now.** [...] **the feed is defined for SHA-1 repositories only.** The publisher must refuse to publish a feed for a SHA-256 repository rather than truncate or re-derive an id, and the consumer treats an absent feed there as feed-off and serves from its native layer — the same path as any other missing feed, so no new failure mode. Widening `indexed_commit` to a length-tagged object id is a `.pg` header change and the one part of this design that would need a format bump; it is deferred, not assumed away. Step 6's corpus must record each repository's object format, so the scope note is confirmed rather than trusted.

Two consistency follow-ons: the corpus gate now records each repository's object format, and the "No `.pg` version bump is required" claim is scoped to the symbol table so it does not read as contradicting the header-widening note.

---

### The four stale findings

Re-queried with `--freshness any`. All four are marked `stale`, and the reason is that the ADR was already edited to answer them. Verified rather than assumed:

- `01M0XGF5HE53` (aggregate counts cannot drive per-language fallback) — answered by freshness clause 2: the note "now carries `language_tiers` (each language as `semantic` or `inventory-only`) and `partial_failures` (the failure records, with path and code, not just `partial_failure_count`)". Present in the merged code (`compact_snapshot.go`, `language_tiers` / `partial_failures` fields).
- `01M0XGF5GYRP` (inventory-only symbols advertised as `Scip`/`Semantic`) — answered by *Precision must follow the tier, not the transport*: "only `semantic` languages may be written as `Scip` / `Semantic`; an `inventory-only` language either keeps the native layer or is written with `Syntax` precision".
- `01M0XGF5HV9Z` (feed-off cannot survive step 9) — not resolved, but no longer undisclosed: the ADR states "**This conflicts with step 9 and the RFD does not yet resolve it**" and open question 9 lays out the three ways out and requires picking one before step 5. That is the right shape for an open RFD; it is a decision for the reviewers, not a text fix.
- `01M10SVGWPMD` (note not cryptographically bound to the protobuf) — partly answered: the note now carries `commit` and `tree`, and freshness clause 1 makes an exact-`indexed_commit` match mandatory with staleness a hard miss. What remains — an artifact digest binding the two files atomically — is a real residual and belongs with the publisher wrapper's contract. Flagging it as still worth a decision rather than claiming it closed.

---

### Not touched

Option A's cons list still calls #97's in-memory materialization "an unvalidated scale risk on a monorepo" while open question 4 reports it measured at ~1.5x the streaming formats on a mid-size repo and still unmeasured on a monorepo. Both statements are true in their own scope, and OQ4 explicitly supersedes the earlier assumption, so this was left alone rather than rewritten.

Docs-only change; no build or test behaviour is affected.

ashtom:

**Verdict: no longer stale — but findings remain open**

### Correction

My last comment said this was stale (219 commits behind) and could not be evaluated without a manual
merge. The branch is now **0 behind** and merges clean, so disregard that part.

### Entire trail findings — still open and current

```
01M19NB9VEJF  Medium  open    current    docs/adr/0005-converge-code-search.md:245  `partial_failures` does not contain a language field—only path, code, severity, …
```

suhaanthayyil:
## Round 8 Medium — `docs/adr/0005-converge-code-search.md:245` — **VALID**, ADR corrected in `f3ace42e`

### The claim, confirmed against the shipped artifact

Freshness-contract item 2 said *"Counts say something was skipped; **only records say which language
and which file**"*, and described `partial_failures` as *"the failure records, with path and code"*.

`sem.PartialFailure` (`internal/sem/provider.go:255-261`) is `code`, `severity`, `file_path`,
`effect_on_semantic_completeness`, `detail`. No language, and two of the five fields the RFD did not
mention at all.

Not from the struct — from a real export. `entire-graph snapshot --repo . --format scip` over a repo
with a 23 MB TypeScript bundle, stderr note verbatim:

```json
{
  "record_type": "scip_omissions",
  "version": "entire-graph-scip-omissions/v1",
  "format": "scip",
  "partial_failure_count": 1,
  "language_tiers": { "Go": "semantic", "TypeScript": "semantic" },
  "commit": "61c203fb...", "tree": "6e353a9d...",
  "partial_failures": [
    {
      "code": "E_FILE_TOO_LARGE",
      "severity": "warning",
      "file_path": "vendor_bundle.ts",
      "effect_on_semantic_completeness": "file record emitted but symbol parsing skipped",
      "detail": "file is 23600020 bytes, above max parser input 4194304 bytes; content was never held in memory"
    }
  ]
}
```

A consumer implementing the per-language fallback from that paragraph would have gone looking for a
field that is not there.

### Which side moves: the ADR, not the implementation

Adding `language` to `PartialFailure` would be additive and therefore *allowed* by the `v1.x` policy,
but it would be the wrong fix:

1. **`PartialFailure` is one shared record type** across `search`, `index`, `def`, `neighbors`,
   `impact` and the snapshot summary. Several of its producers have no language in hand at all —
   `E_REPO_IGNORE_UNREADABLE` over an ignored subtree, a failed git listing, walk errors. The field
   would be blank often enough that a consumer *routing per-language fallback on it* would fall back
   for nothing. A field that is absent is better than one that is silently empty half the time.
2. **The information is already in the artifact, with no format change and no bump.** Every discovered
   file becomes a SCIP `Document` — including one whose parse was skipped
   (`compact_snapshot.go:464`) — and the export sets `Document.language` from the same file record the
   tier was derived from (`:451-457`). Decoding the index emitted above:

   ```
   doc path="main.go"           language="Go"
   doc path="vendor_bundle.ts"  language="TypeScript"
   ```

   `partial_failures[].file_path` → `Document.relative_path` → `language` → a key in `language_tiers`.
   The contract's requirement is satisfiable today.
3. Step 1 freezes the note as `entire-graph-scip-omissions/v1.0`. The document being frozen against is
   the one the other team implements from, so it has to describe the shape that exists at the moment
   of the freeze.

### What changed in the ADR

Item 2 now states the record's real five fields, says explicitly that there is no language field and
why that is deliberate, and names the join a consumer must perform instead.

It also names a wrinkle that was previously left to be discovered in implementation and is directly
load-bearing for line 262's *"partition by the note's `language_tiers`"*: **`Document.language` is the
SCIP enum name while `language_tiers` is keyed by entire-graph's own language name.** They differ for
a closed set — `C++`/`CPP`, `C#`/`CSharp`, `F#`/`FSharp`, `Objective-C`/`Objective_C`,
`Protocol Buffers`/`Protobuf`, `Starlark`/`Skylark`, `Bash`/`ShellScript`, and the rest at
`compact_snapshot.go:852-892`. A consumer that joins naively mis-keys every one of them. The mapping
is total and one-way, so the ADR puts it in the written contract of open question 2 next to the
moniker scheme rather than leaving each consumer to re-derive it. **Not** changed on the wire here —
flagging it as the follow-up it is.

### Tests

`internal/cli/adr0005_feed_contract_test.go` — precedent is `agentguide_test.go`, which checks
`AGENTS.md`'s claims against the code the same way.

- `TestADR0005NamesTheFailureRecordItActuallyGets` — drives the expected field list off
  `reflect.TypeOf(sem.PartialFailure{})` rather than a hard-coded list, scoped to freshness-contract
  item 2 so a field name in an unrelated paragraph cannot satisfy it. Adding a field to the struct
  (a `language`, one day) fails this until the paragraph is brought with it — which is the direction
  the drift actually runs.
- `TestSCIPFeedResolvesAFailingFilesLanguage` — runs the documented join end to end: real
  `snapshot --format scip`, `E_MINIFIED` on a TypeScript file, note parsed from stderr and index from
  stdout, every failure's `file_path` resolved to a Document with a non-empty language, and that
  language present in `language_tiers`.

### Mutation verification

|---|---|---|
| ADR paragraph restored to its pre-fix text | n/a | **FAIL** — all five field assertions plus the `records say which language` assertion |
| `doc.Language` assignment disabled (`if language != "" && false`) | yes | **PASS** — correctly: `documentFor` falls back to `encoder.files[filePath]`, so the join still resolves |
| both language branches disabled (`if doc.Language == "" && false`) | yes | **FAIL** — `failure E_MINIFIED names "bundle.ts" whose document carries no language` |

The middle row is why the join is worth relying on: it survives the first-choice source being absent.
`grep` confirmed each mutation applied before running.

### Gate

`go test ./internal/cli` — 752 pass. `gofmt`, `go vet` clean. Every `file:line` citation added to the
ADR was re-read on this branch head.

PR #147, opened 2026-08-21
Add nightly LoCoMo CI (entire-graph arm, OIDC — no stored credential)

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/107

Nightly full LoCoMo (n=1,540) for the **entire-graph arm only**, at 06:00 UTC.

Addresses both stated requirements directly: external folks cannot trigger it, and there is no Azure key stored in this repository to extract.

## External triggering is closed structurally

- **`schedule` only fires on this repo's default branch.** A fork cannot schedule a workflow upstream, so the nightly path is unreachable from outside.
- **`pull_request_target` is absent, deliberately** — it is the one trigger that hands repository secrets to fork code. A comment in the file says so, to stop it being added later.
- **`workflow_dispatch` is gated** by the `benchmark` environment's required reviewers, on top of the write access the trigger already requires.
- **`if: github.repository == 'entireio/entire-graph'`** stops the job even if this file is copied into a fork with its own schedule.
- **All 5 actions pinned to full commit SHAs** (verified against the GitHub API, not written from memory).

## No stored credential

The job takes a short-lived GitHub OIDC token, federates into Azure via workload identity federation, and reads the Foundry key from Key Vault at run time. Nothing persistent exists in GitHub to steal; rotation happens in Key Vault without touching this repo; deleting the federated credential revokes CI access immediately.

**Stated limit, in the file itself:** the harness authenticates to Foundry with an `api-key` header (`patches/0001-*`), so the key is necessarily in the job environment while the benchmark runs, and any job with code execution can read its own environment. Eliminating that entirely would mean patching `llm_client.py` to use bearer auth — which would change the measured harness away from the published configuration. The chosen mitigation is that this workflow never runs untrusted code: default branch only, no fork triggers, SHA-pinned actions.

## The gate checks integrity, not score

`bench/memory/ci/summarize_run.py` fails the run on: fewer than 1,540 questions scored, `FAIR_MODE` inactive, arm-asymmetric settings present, any dropped search, or a score below a plausibility floor.

It deliberately **does not alert on a score delta.** Run-to-run noise here is **0.65pp**, and an identical entire-graph config re-run 26 hours apart drifted **2.21pt** (`LOCOMO-COMPARISON.md` §7). A nightly threshold inside that band would fire on noise every few nights and be ignored within a week. The score is a trend across runs, not a comparison against last night.

Result selection is by `metadata.run_id`, never by file timestamp — selecting the newest file has repeatedly picked up stale diagnostic artifacts and reported them as a finished run.

## Cost

~**$190/run** (~$5.7k/month) at Foundry list prices for `gpt-5.6-sol` ($5/M in, $30/M out) with the launcher's `--top-k-cutoffs 200`. Ingest is **$0** — the entire-graph arm makes no model calls to build its index. Widening cutoffs back to `200,50,20,10` would quadruple answerer/judge calls to ~$475/run, which is why the launcher does not.

## Verification done

- Workflow YAML parses; triggers are exactly `[schedule, workflow_dispatch]`; asserted no `pull_request_target`
- All action SHAs resolved against the GitHub API and confirmed to match their tags
- `cmd/entire-graph` build path confirmed to exist
- Summarizer tested on synthetic artifacts: clean run → exit 0; incomplete + `FAIR_MODE` off + dropped search → exit 1 with all three errors reported
- Confirmed it ignores a stale diagnostic result file carrying a **later** timestamp, selecting on `metadata.run_id`

## Before this can run

Operator setup is documented in `RUNNING-LOCOMO.md` → "Nightly CI" and is **not** in the repo: Azure AD app + federated credential (subject `repo:entireio/entire-graph:environment:benchmark`), Key Vault secret `azure-ai-api-key`, GitHub `benchmark` environment with required reviewers and `main`-only branches, and five repository *variables* (identifiers, not credentials).

Until that exists the scheduled run will fail at the Azure login step. Worth a `workflow_dispatch` run to verify wiring before trusting the schedule.


ashtom (approved):
Reviewed the direct OIDC conversion and follow-up hardening. Verified the locked Python environment, fail-closed result/provenance gates, workflow syntax, pinned upstream reconstruction, and full Go checks. No blocking findings remain.

PR #150, opened 2026-08-27
fix(sem,gitutil): list only regular files in a committed tree

ashtom:
https://entire.io/gh/entireio/entire-graph/trails/110

## The bug

A tracked symlink is a blob whose content is its **target path**. The committed-tree listing
passes `--name-only`, so it cannot tell one from a regular file, and lists it as source — handing
that path to a parser as though it were code.

```console
$ ln -s real.go link.go && git add -A && git commit -m fixture
$ entire graph snapshot --repo . --format ndjson | grep '"record_type":"file"'
  link.go  bytes=7      # the string "real.go", parsed as source
  real.go  bytes=21
```

The **working-tree** listing already refuses non-regular entries (`openSource`'s worktree branch
tests `fs.ModeSymlink` and `IsRegular`). So one repository produced two different graphs depending
on `--worktree`: different file records, symbols, relations and counts.

## Why `ListFiles` couldn't be patched in place

Not a forgotten filter. `git ls-tree --name-only` emits no mode, and **a symlink *is* a blob**, so
no object-type test separates them:

```console
$ git ls-tree -r HEAD
120000 blob 91fd0bf…	link.go
100644 blob 3129c81…	real.go
```

Only the mode distinguishes them. `ListRegularFiles` reads it and keeps entries whose **file-type
bits** are regular — `100644`, `100755` — excluding `120000` symlinks, `160000` gitlinks and trees.
It tests the type bits rather than enumerating permission spellings, so an unexpected mode is
refused rather than admitted.

`ListFiles` is **deliberately left unchanged**. Its only other caller is the dependents co-change
fallback, which enumerates paths for `git log`, where a symlink costs nothing.

## Cache versions are bumped

`provider-records-v8 → v9` and `search-snapshot-v12 → v13`.

The enumeration rule is not part of either cache key. Without the bump, a pre-fix and a post-fix
graph of the same tree share an entry, and a stale graph containing the symlink could be served as
current.

## Lineage

This is the committed-tree counterpart of `5f06a438` *"list the working tree the way git does and
bound every file read"*, which gave the worktree path its regular-file filter and left the
committed path untouched.

Worth stating explicitly, because this repository has a lot of symlink hardening already: all 44
of those commits target **`.git` metadata** — gitdir, gitfile, commondir, HEAD, pointers,
objects/refs. `gitutil.ListFiles` had never been touched by any of them. This is a different
problem: not "can a repository escape through its `.git` structure", but "is a tracked symlink
source code".

## Tests

This repository's own HEAD carries 615 `100644` and 6 `100755` entries and nothing else, so **the
regression is not reachable from it** — the tests build a fixture that carries a symlink, and skip
where the platform cannot create one.

- `TestIsRegularTreeMode` — the mode predicate, including `040000`, malformed input and the
  no-leading-zero spelling `40000`.
- `TestListRegularFilesExcludesSymlinksAndGitlinks` — asserts `ListFiles` still returns the symlink
  (unchanged by design) while `ListRegularFiles` does not.
- `TestCommittedAndWorktreeAgreeOnSymlinks` — end-to-end, asserting the **invariant** that broke
  (committed and worktree agree) rather than the symptom.

Failing-first, against unpatched `main`:

```
--- FAIL: TestCommittedAndWorktreeAgreeOnSymlinks
    committed snapshot listed a symlink as a source file
    committed and worktree disagree: map[link.go:true real.go:true] vs map[real.go:true]
```

## Gate

`gofmt -s -l .` empty · `go vet ./...` clean · `go build ./...` clean · `go test ./...` green
across all 8 packages.

Not independently validated: Windows and Linux were not verified natively. The tests skip rather
than fail where `os.Symlink` is unavailable, which on Windows depends on developer mode or
elevation — so a Windows CI run that reports green may have skipped the symlink cases, and the
mode predicate test is the part that always runs there.

## Provenance

Found while reviewing #97's SCIP export, where it surfaced as "SCIP requires `Document.relative_path`
to identify a regular file." It is not a SCIP bug: the export was faithfully reflecting a file
record the provider should not have emitted. Split out of that PR because the fix is provider-wide
and its cache-key bump deserves to be the headline rather than a footnote in a feature branch.

---

> [!NOTE]
> **Medium Risk**
> Changes which paths enter the semantic graph and invalidates on-disk caches; behavior is correctness-critical but scoped to enumeration with regression tests.
> 
> **Overview**
> Fixes **committed-tree snapshots treating tracked symlinks as source files** (blob content is the link target, e.g. `"real.go"` parsed as code) and **worktree vs committed graphs disagreeing** on the same repo.
> 
> **gitutil** adds `ListRegularFiles` (mode-filtered `ls-tree` with `--end-of-options`), `IndexNonRegularPaths` (index modes when `lstat` lies under `core.symlinks=false`), and `validateTreeish` on tree listings. **`ListFiles` stays broad** for co-change history but gets the same revision hardening.
> 
> **sem** switches committed enumeration to `ListRegularFiles` and skips index symlinks/gitlinks in worktree listing even when they look like regular files on disk.
> 
> **Provider and search snapshot cache versions** bump (`v8→v9`, `v12→v13`) so pre-fix graphs with symlink file records are not reused.
> 
> Tests cover mode filtering, option-shaped revisions, and end-to-end `StreamSnapshot` parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 324548bfa4e25434c5eca4e8c8c9df9667ff9f45. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ashtom:
bugbot run

ashtom:

### State

- 4 ahead, **13 behind** `main`; **merges clean** despite the distance.
- Its green checks ran against a base 13 commits old, so I re-ran the gate on the merged tree.

| Check | Tree | Result |
|---|---|---|

### Note

This overlaps **#166** in subject matter — both are about how a committed tree's non-regular
entries are classified. #166 classifies symlinks by tree mode in the diff path; this one lists only
regular files in a committed tree. They do not conflict textually, but they should be read together
so the two paths end up agreeing about what a non-regular entry means.

Re-queried immediately before posting.

```
01M128WBZSBF  Medium  open    current    internal/gitutil/git.go:182  This reads the entire index through `run`/`bytes.Buffer` and then `strings.Split…
01M128WBZENZ  Medium  open    current    internal/sem/provider.go     The index-mode veto silently drops a dirty regular file that replaces a tracked …
```

### Verified vs. assumed

**Verified by execution:** merge onto `bb41f223`; `go vet`/`gofmt`; the full test run for every package this
PR touches, on the merged tree; the PR's own tests with and without its production change; trail findings re-queried immediately before posting.

PR #152, opened 2026-08-27
RFD 0006: reference positions in the SCIP export

ashtom:
https://entire.io/gh/entireio/entire-graph/trails/111

## What this is

An RFD, not a fix. It documents a measured gap in `snapshot --format scip` and proposes what to do
about it. Opened as a **draft** because it asks for a decision.

## The finding

`snapshot --format scip` emits **25,716 SCIP reference occurrences** for this repository, and **none
of them is located at a reference**. Every one carries the enclosing caller's whole declaration
span, mean width **100.66 lines**.

```
CALLS relations                                    19,233
positioned evidence records                        18,988
  …byte-identical to the caller symbol's own span  18,988
  …narrower                                             0
unpositioned                                          246
```

```
caller '_safe_id'  spans 102-104;  CALLS evidence is 102-104
caller '_iso'      spans 111-117;  CALLS evidence is 111-117
caller '__init__'  spans 121-141;  CALLS evidence is 121-141
```

Not language-specific: Go 99.9%, Python 100%, TypeScript 100%, Java 100%, C 98.1%, C# 100%,
Zig 100%. Same rule everywhere — the apparent per-language spread in *width* is just average
function size.

## Why it matters, and how much

The relations are **true**. Only the positions are wrong.

| Consumer question | Today |
|---|---|
| Where is `Foo` defined? | ✅ correct |
| Which symbols reference `Foo`? | ✅ correct set |
| What symbol is at `file.go:50`? | ❌ resolves as a reference to everything that caller calls |

The third row is the one [RFD 0005](https://github.com/entireio/entire-graph/pull/146) has peregrine
joining on — *"peregrine joins on name plus position"*.

So this is **not** "the SCIP export is broken". Go-to-definition, symbol metadata and the type
hierarchy are sound. Find-references is a correct set with unusable positions.

## The proposal

**Option A now** — emit an occurrence only where evidence is genuinely a single line; count the rest
in the omission note. A SCIP `Occurrence` is a positional claim, and emitting one from a span nobody
measured asserts something the graph does not know. References drop 25,716 → ~48 here. Definitions
(9,877) and implementation relationships (45) are untouched.

**Option D as the real fix** — thread scanner offsets through instead of reducing to name sets. Not
irretrievable: `maskBytes` preserves length and newlines so offsets map 1:1, and JS/TS already keeps
tree-sitter positions before discarding them. Sized medium-to-large: ~40 `Evidence{` sites plus
per-language extractors, several returning position-less name sets.

**Option C explicitly rejected.** `neighbors` and `impact` already reconstruct a plausible call line
from `Evidence.Detail` at query time, and porting that into the export is the obvious cheap move. It
is a heuristic display aid — it takes the *first likely* line and counts the rest. As an interchange
claim it is worse than what it replaces: a 100-line span is detectably unusable, a confidently wrong
narrow position is not.

## A second gap, found along the way

Relation dedup keys on `(from, to, type)` and discards later records regardless of evidence
(`provider.go:1415`); `receiverCalls` dedupes by `receiver.method` before that. **Repeated calls to
the same callee collapse into one relation.** Even perfect spans could not yield one occurrence per
reference without also emitting multiple evidence entries — which is schema-safe, since `evidence`
is already an array and ADR 0001's `1.x` rules are additive-only.

## Provenance

Measured three ways independently — by hand, by a Fable 5 agent, and by Codex on `gpt-5.6-sol` at
xhigh reasoning. All three reproduced the same result; the two agents each corrected part of my
framing, and Codex is the reason Option C is rejected rather than recommended.

The appendix carries a runnable command; its output matches the table.

## Numbering

**0006**, not 0005 — 0005 is claimed by the open RFD in #146. (Two `0004`s were narrowly avoided the
same way last week.)

## Not in scope

No code changes. Option A is a small encoder change and should be its own PR once the decision is
made; Option D is provider work that deserves scheduling on its own merits rather than being rushed
to prop up a format.

---

> [!NOTE]
> **Low Risk**
> Documentation-only; no code, schema, or export behavior changes until a separate PR implements the proposed Option A.
> 
> **Overview**
> Adds **RFD 0006** (proposed, open for comment) documenting a measured gap in `snapshot --format scip`: reference occurrences are emitted with **caller declaration spans** (~100 lines on average), not actual call-site lines, while relation sets remain correct. The RFD recommends **Option A**—emit reference occurrences only when evidence is single-line and record the rest via the existing omission note—and defers **Option D** (provider threading of offsets, per-site evidence, dedup changes) as the real fix; it explicitly rejects porting CLI `resolveCallSite` heuristics into export.
> 
> Also links the new RFD from `docs/README.md` and spells out implications for **RFD 0005** (peregrine join on name plus position). **No runtime or encoder changes** in this PR—decision and follow-up work only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f49e39e5032960f126b0ed0105fda536f8cc0e5e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -31,6 +31,7 @@ authoritative in `entire graph capabilities --json`.
 | [ADR 0002](adr/0002-committed-tree-cache-key.md) | Accepted committed-tree cache-key decision; authoritative for committed-tree cache identity |
 | [ADR 0003](adr/0003-working-tree-search-snapshot-cache.md) | Superseded clean-tree search-snapshot reuse decision |
 | [ADR 0004](adr/0004-working-tree-cache-security-boundary.md) | Accepted working-tree cache bypass security decision; authoritative for working-tree eligibility |
+| [RFD 0006](adr/0006-reference-positions.md) | Proposed decision on SCIP reference-position fidelity; not yet decided, open for comment |
 
 Completed plans, superseded references, branch diaries, and point-in-time proof
 logs are listed in the [archive](archive/README.md). Archived documents are kept
--- a/docs/adr/0006-reference-positions.md
+++ b/docs/adr/0006-reference-positions.md
@@ -1,0 +1,213 @@
+# RFD 0006 - Reference positions: what the graph can honestly tell a SCIP consumer
+
+Status: Proposed (RFC, open for comment)
+Date: 2026-08-27
+
+## TLDR
+
+`snapshot --format scip` emits 25,716 SCIP reference occurrences for this repository. **None of them
+is located at a reference.** Every one carries the enclosing caller's whole declaration span,
+averaging ~100 lines, because that is what relation evidence records. The relations themselves are
+correct; only the positions are not.
+
+A SCIP `Occurrence` is a positional claim, so emitting one from a span we did not measure asserts
+something the graph does not know. This RFD proposes stopping that, and treats reference-position
+fidelity as the provider question it actually is.
+
+Definitions, symbol metadata and implementation relationships are unaffected and remain sound.
+
+## The measurement
+
+Reproduced independently three times — by hand, by a Fable 5 agent, and by Codex (`gpt-5.6-sol`,
+xhigh) — against `8935cfe9`.
+
+| | |
+|---|---|
+| `CALLS` relations | 19,233 |
+| evidence records carrying a position | 18,988 |
+| **whose span is exactly the `from_id` symbol's own span** | **18,988** |
+| whose span is narrower | **0** |
+| evidence records carrying no position at all | 246 (138 `route_endpoint_match`, 108 `top_level_call_site`) |
+
+`call_site` evidence: 11,886 records, 11,883 multi-line, mean width **100.66 lines**.
+
+The counts above are what the command in the appendix prints at `8935cfe9`. Independent runs landed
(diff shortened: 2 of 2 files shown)


ashtom:

**Verdict: advisory — docs only**

- Live PR head `f49e39e5`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `f49e39e5` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|
| revert-the-fix check | — | N/A — docs-only PR |

### Findings

**1. (advisory) ADR number depends on #146 landing first.** `main` has ADR 0001–0004. This PR adds
`docs/adr/0006-reference-positions.md`; **#146** adds `0005-converge-code-search.md`. There is no
collision, but if #146 is abandoned the numbering gains a permanent gap at 0005. Worth confirming
#146 lands first, or renumbering to 0005.

No code changes, so there is no fix to revert and nothing to discriminate. `go test` on the
touched packages passes on the merged tree.

### Independent review state

- Cursor Bugbot: ran. Comment `bugbot run` to trigger it.

PR #153, opened 2026-08-27
fix(sem): three SCIP export follow-ups from the #97 review

ashtom:
https://entire.io/gh/entireio/entire-graph/trails/112

Closes the SCIP export follow-ups raised after #97, updated and re-reviewed against current `main`.

## Relationship semantics

- `is_reference` is emitted only when both endpoints are method/function-like. Type relationships and unresolved or type-like external targets fail closed, while member overrides still merge references as SCIP intends.
- A missing implementation source is now reported separately from a missing target.

## Structural omission accounting

- Redundant `CONTAINS` and `DEFINES` relations are skipped only when they agree with the native SCIP structure that represents them.
- A missing child increments `missing_target_relations`; a present child with a missing parent increments the new additive `missing_source_relations`; a present but mismatched relationship increments the unsupported-relation count.
- Native `DEFINES` redundancy is verified against the symbol's actual file identity rather than assumed.

## Cargo package version

Cargo and Python project versions are now read with a pinned, real TOML parser instead of a line scanner. For a semantic root `[package]`, a literal version wins, and an exact `version.workspace = true` or `version = { workspace = true }` declaration may inherit `[workspace.package].version`.

This deliberately does **not** assign a repository-wide version to a virtual Cargo workspace with no root package. Such a workspace can contain mixed or non-inheriting member versions; until the SCIP exporter has member-aware package identities, it continues to use the unknown version `0`. This is the explicit product decision for Trail finding `01M1FZ232PHHNSSH92AYSQM9GR`.

The parser is `github.com/pelletier/go-toml/v2` v2.4.3, a pinned pure-Go dependency with no non-standard transitive dependencies and no runtime egress. Regressions cover comments, quoted and dotted keys, multiline values, arrays of tables, nested arrays, malformed TOML, duplicate/redefined keys, exact inheritance syntax, literal overrides, and virtual workspaces.

## Deliberately not included

- The reference-position policy from the same review remains an RFD 0006 decision.
- Moniker ownership and container ancestry remain coupled to RFD 0005's package-identity decision.

## Verification on `af34a0f3`

- `gofmt -s -l .` — clean
- `git diff --check origin/main...HEAD` — clean
- `go mod tidy -diff` and `go mod verify` — clean
- `go vet ./...` and `go build ./...` — pass
- `go test -timeout 30m ./...` — pass
- `go test -race -timeout 30m ./...` — pass
- focused TOML/SCIP tests — pass for 20 repetitions and under the race detector
- three independent source reviews plus fresh Cursor Bugbot — clean
- GitHub builds and tests — pass on Ubuntu, macOS, and Windows

Entire's Checks, Findings, and Up-to-date gates pass. Its aggregate check is red only because Trail #112 still awaits explicit reviewer approval.

---

> [!NOTE]
> **Medium Risk**
> Changes SCIP protobuf relationship semantics and omission JSON (additive field), which downstream SCIP consumers may rely on; manifest parsing swap affects every symbol's package version for Cargo/Python repos.
>
> **Overview**
> Addresses post-#97 SCIP export follow-ups: **tighter relation encoding**, **honester stderr omission accounting**, and **proper manifest version parsing** for `snapshot --format scip`.
>
> **SCIP relationships and structural relations:** Implementation-family edges no longer always set `is_reference`; it is emitted only when **both** endpoints are method/function-like (member overrides still merge references; type inheritance does not). `DEFINES` / `CONTAINS` are treated as redundant only when they match native structure (`ContainerID`, file-scoped `DEFINES`); extension `CONTAINS`, bad `DEFINES` sources, and missing endpoints are counted in `unsupported_relation_counts`, `missing_target_relations`, or the new optional **`missing_source_relations`** (e.g. implementation/containment with no local source). Docs in `snapshot-format.md` match this contract.
>
> **SCIP package version:** `Cargo.toml` / `pyproject.toml` versions are read via pinned **`github.com/pelletier/go-toml/v2`** instead of a conservative line scanner. Root `[package]` literals win; exact `version.workspace = true` / `{ workspace = true }` can inherit `[workspace.package].version`; **virtual workspaces** without a root package still fall back to unknown `0`. Broad regression tests cover TOML edge cases and SCIP omission behavior (`scip_followups_test.go`).
>
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af34a0f3933c896d10cea768ffc0a213788ffbc7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ashtom:

**Verdict: clean**

- Live PR head `582741a1`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `582741a1` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

### Independent review state

- Cursor Bugbot: ran (neutral). Comment `bugbot run` to trigger it.

ashtom:

**Verdict: unchanged since my last review, but 2 finding(s) have appeared since**

no longer does.

```
01M11V5SW1V2  Low     open    current    internal/sem/scip_project_version.go:112  The inline-table branch only checks for an opening `{`, so malformed input such …
01M11V5SVT7J  Medium  open    current    internal/sem/scip_project_version.go:108  Valid Cargo inheritance declarations with trailing comments are rejected because…
```

### Verified vs. assumed

suhaanthayyil:
Addressed the Bugbot finding on `tomlTableInheritsFromWorkspace`.

**Confirmed real and reproduced** before fixing — both spellings broke with a trailing comment:

| input | before | after |
|---|---|---|
| `version.workspace = true # inherit` | `false` | `true` |
| `version = { workspace = true } # inherit` | `false` | `true` |
| `version.workspace = true` | `true` | `true` |

`tomlTableRawValue` returned the whole rest of the line, so `"true # inherit"` never equalled `"true"`, inheritance read as absent, and the crate fell back to the unknown version — the SCIP package-identity collapse this PR exists to stop. It failed silently: manifest parsed, no error, identity just wrong.

Reused the **existing quote-aware `stripTOMLComment`** (provider.go) rather than adding a second stripper — which is what the finding was pointing at with "quoted versions already strip comments; this path does not". A version that legitimately contains a hash (`"1.0#rc1"`) is preserved, and that case is pinned in the test.

Regression test verified to fail without the fix and pass with it. Full `internal/sem` suite green (75.8s).

suhaanthayyil:
bugbot run

ashtom:
bugbot run

PR #154, opened 2026-08-28
fix(sem): bind the schema version to on-disk cache entry validity

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/113

## Problem

The two on-disk caches persist **serialized records**, but neither entry's identity records the
schema those bytes were serialized under. Nothing else in either envelope stands in for it:

- `providerRecordsCacheVersion` / `searchSnapshotCacheVersion` track the **caching machinery**,
  not the record wire format. They move independently of `sem.SchemaVersion`.
- `provider_version` does not separate two schemas either. It is the constant `dev` for every
  local build, and `.github/workflows/release.yml` stamps the constant `v0.0.0-ci` on every
  non-tag CI build.

So a build that changes the wire format, without also changing the injected binary version,
reads the previous schema's entries and replays them as its own output.

## Reproduction (before this change)

Binary A at the current schema populates a cache. Binary B is the same tree with one wire field
renamed (`SymbolRecord.signature` -> `signature_v2`) plus a `SchemaVersion` bump — i.e. exactly a
schema change — and the same `dev` provider version:

```
$ egA index  --repo . --format json --cache-dir $C          # populate
$ egB index  --repo . --format json --cache-dir $C
index_cache_hit= True                                        # schema bump did not invalidate

$ egB snapshot --repo . --format ndjson                      # WARM (A's cache)
{...,"signature":"func Beta() int","body_hash":"b8e16..."}

$ egB snapshot --repo . --format ndjson                      # COLD
{...,"signature_v2":"func Beta() int","body_hash":"b8e16..."}
```

Same binary, two different wire shapes, decided by cache state. The warm snapshot header also
still declared the *old* `schema_version`, so a downstream consumer (entire-brain ingests this
NDJSON) receives records that disagree with the producing binary.

## Fix

Make the schema part of each cache entry's address and validity:

- `cachedProviderRecords` gains a `schema_version` field. The record bytes are replayed verbatim,
  so the envelope is the only place the schema can be recorded. Entries written before the field
  existed decode as `""` and are correctly rejected as a different schema.
- `validCachedSearchSnapshot` checks the schema already carried on the stored snapshot header,
  alongside the provider/profile/tree checks it does today. No new field needed.

The schema is also part of both cache keys. A mismatch therefore addresses a different artifact
instead of letting schema-N and schema-(N-1) builds overwrite each other's entry; the read-time
schema checks remain defense in depth for stale or hand-edited artifacts.

**Rollout note:** both existing *records* and *search* entries were written under schema-independent
addresses. After this lands, each cache family cold-misses once under the new schema-qualified
address. Old artifacts remain at their previous addresses until ordinary cache cleanup; they are
not overwritten in place. That one-time cold rebuild is intentional because the cached bytes'
wire compatibility was never encoded in their address.

## Scope of the bug

Today the caches are protected from a schema change only **incidentally**, by release versions
happening to change between releases. Nothing in the cache knows what a schema is. Every build
that shares or pins a version loses that accidental protection — all local builds (`dev`), all
non-tag CI builds (`v0.0.0-ci`), and any packaging that pins a version string. This change makes
the protection deliberate instead of incidental.

Two hand-built test fixtures now set `SchemaVersion` on their headers. Real snapshots always
stamp it (`newProviderSnapshot`); only the fixtures omitted it.

## Tests

`internal/sem/cache_schema_version_test.go` — both caches, driven against the **persisted bytes**
rather than in-memory structs so the "field is simply absent" upgrade case is covered:

- records cache: foreign schema -> miss; schema absent -> miss; each with a control asserting the
  build's own entry is still served
- search cache: foreign/absent schema rejected, both in-memory and through a write/read round trip

Confirmed binding by mutation: deleting the `cache.SchemaVersion == SchemaVersion` clause (and the
header clause) makes the new tests fail; restoring it makes them pass.

## Verification

`go test -race -count=1 -run 'Cache|Schema|Snapshot|Records' ./internal/sem/` — clean, zero data races.

Because this change can only ever *invalidate* cache entries, it was also checked for a
cache-defeat slowdown against `origin/main` on the same machine, same subset:
baseline 100.4s / 279 tests, this branch 116.1s / 284 tests. The delta tracks the five new
round-trip tests, not extra rebuilds.


suhaanthayyil:
Adversarial re-review. The gate is real (reverting the `cache.SchemaVersion == SchemaVersion` clause fails `TestProviderRecordsCacheMissesEntryFromAnotherSchemaVersion/schema version absent`), both `writePath` callers are covered, and an absent field fails closed into a rebuild rather than a hard error. Three issues.

**1. HARD BUILD BREAK with #160.** #160 deletes `func (entry cacheEntry) writePath()` from `internal/sem/cache_entry.go` and rewrites its callers, including `search_cache_test.go:132` → `filepath.Join(root, relative)`. This PR adds `internal/sem/cache_schema_version_test.go:191`, which calls `entry.writePath()`. Verified: `git grep writePath origin/audit/graph-cache-write-confinement -- internal/sem/` returns nothing. Merge both in either order and package `sem` does not compile. Whichever lands second has to be rebased onto the other's accessor.

**2. Contradicts #155 on the same field.** #155 adds `CheckReadableSchemaVersion`, which rules `SnapshotHeader.SchemaVersion` compatible across a whole major and merely *warns* on a newer minor. This PR requires exact equality on the same field of the same struct. After both land, a header declaring `"1.0"` against a `"1.1"` build is refused by the search-snapshot cache and accepted by compact-snapshot load. Neither PR references the other. The distinction you actually want — "cache identity is exact, interchange compatibility is per-major" — is defensible but has to be written down and ideally expressed as two named constants, not one constant read two ways.

**3. Drift is unbound, so the motivating repro can recur.** Nothing ties `SchemaVersion = "1.1"` (`provider.go:39`) to the record shape. Rename `SymbolRecord.signature` without bumping the constant and stale entries are served again — the exact failure this PR exists to prevent. #156 freezes `Result` only, not the record types. A shape digest over the persisted record structs (in the spirit of #156's `jsonWireShape`, applied to `SymbolRecord`/`RelationRecord`) would make the version self-enforcing.

Minor: merging this invalidates every existing user's provider-records and search-snapshot caches, forcing one full rebuild. Worth a line in the PR body.

ashtom:

**Verdict: clean, with one operational note**

- Live PR head `b3a71ecd`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `b3a71ecd` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

**1. (note, not a defect) Every existing on-disk records-cache entry is invalidated by this change.**
`cachedProviderRecords` gains `schema_version`, and `validCachedProviderRecords` now requires it to
equal `SchemaVersion`. Entries written before this field existed decode as `""` and are correctly
rejected — which is the intent — but it does mean the first run after this lands rebuilds cold.
`providerRecordsCacheVersion` is deliberately *not* bumped; the `""`-rejection covers the same
ground. That reasoning holds, and I confirmed the search-snapshot side differs: it reads
`cache.Snapshot.Header.SchemaVersion`, which already existed, so those entries validate normally
rather than being discarded.

Otherwise clean: the bound is on the right field, and the tests fail without the change.

### Independent review state

suhaanthayyil:
Thanks — taking the note and one cross-PR defect it did not cover.

**Your note (cold rebuild) — agreed, and now written down.** Your read is right: `cachedProviderRecords` gains `schema_version`, pre-existing entries decode as `""` and are rejected, and that is the intent. `providerRecordsCacheVersion` is deliberately not bumped because the `""`-rejection covers the same ground. Your separate check that the search-snapshot side validates normally (it reads `cache.Snapshot.Header.SchemaVersion`, a field that already existed) is also correct. I have added that to the PR body so the first-run cold rebuild is not a surprise.

**A defect neither of us flagged in review: this PR and #160 do not compile together.**

#160 deletes `func (entry cacheEntry) writePath()` from `internal/sem/cache_entry.go` and rewrites its callers to `filepath.Join(entry.root, entry.relative)`. This PR added a NEW caller of it at `internal/sem/cache_schema_version_test.go:191`. Merge the two in either order and package `sem` fails to build:

```
$ git merge origin/audit/graph-cache-write-confinement   # onto this branch, pre-fix
$ go vet ./internal/sem/
vet: internal/sem/cache_schema_version_test.go:195:29: entry.writePath undefined (type cacheEntry has no field or method writePath)
```

Fixed in `f2e15c35` by using the same idiom #160 uses, which compiles on both branches and makes the two merge-order independent. The assertion is unchanged.

Verified:
- `go test ./internal/sem/ -run 'TestProviderRecordsCacheMissesEntryFromAnotherSchemaVersion|TestLoadSearchSnapshot|TestCacheEntryWrite'` on this branch alone: pass.
- Same, on the trial merge of this branch with #160: 13 pass.
- Mutation, on the merged tree: restore `entry.writePath()` → `vet` fails with the error above; revert → clean. The break is real and the fix is what removes it.

**On the three points in my own earlier review that this does not close:** the #154/#155 disagreement about what `SchemaVersion` means (exact for cache identity, per-major for interchange) is still an unwritten decision, and nothing binds the constant to the record shape, so a rename without a bump can still serve stale entries. Both are follow-ups rather than blockers for this PR; say the word if you would rather they land here.

suhaanthayyil:
Follow-up on the open item from my last comment: the `SchemaVersion` question is now **settled and written down**, in an amendment to `docs/adr/0001-ga-schema-contract.md` landed on #156 (`a9db513f`).

The resolution is that this PR and #155 are **both correct** and are answering different questions about the same string:

| question | rule | this is |
|---|---|---|
| may I *read* bytes another build wrote? | same major; warn on newer minor | #155, `CheckReadableSchemaVersion` |
| may I *reuse* a cache entry I wrote? | **exact** match; absent fails closed | **this PR** |
| may the persisted payload's shape change? | additive within a major | #156 |

The justification for the difference, so it is not just an assertion: a tolerance band exists to avoid discarding data you cannot regenerate, and a cache is by definition data you can. Applying the per-major rule to cache validity would serve entries written by a build whose record shape has since changed — the exact failure this PR exists to prevent. So no change is needed here; the rule was right, only the written distinction was missing.

#156 additionally closes the third leg: the frozen shape's digest is now pinned beside the exact version string, so the shape cannot change without the version question being asked.

ashtom:

Head moved `b3a71ecd` → `f2e15c35`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

```
01M17E648GGQ  Medium  open    current    internal/sem/search_cache.go:808  The schema is validated only after opening the existing schema-independent cache…
```

### Verified vs. assumed

suhaanthayyil:
## Finding response — `internal/sem/search_cache.go:808`

**VALID.** Head is now `cb5ee7ae`.

### What was wrong

The read-time check added by this PR does reject a foreign-schema entry — that half works. But it is only half a gate, because the entry's **address** never contained the schema. Neither key function hashed it:

- `searchSnapshotKey` hashed cache-version, repository-path, repository-key, provider-version, tree, profile, max-parse-bytes, max-files, only-files, builtin-secret-rules, policy.
- `providerRecordsKey` hashed the same set plus commit and mode.

So two builds at different `SchemaVersion` resolve to the **same artifact path**. Each one's store replaces the other's entry, and the read-time check then turns the survivor into a miss for whichever build reads next. The pair never warms: every run rebuilds and re-poisons the entry for the other build, and nothing reports it, because a miss followed by a successful rebuild is indistinguishable from a cold cache. The check catches the wrong answer; it cannot stop the mutual eviction that produced it, because by the time it runs the other build's entry is already gone.

### Fix

`internal/sem/search_cache.go` and `internal/sem/records_cache.go`: the schema version is now a key input.

```go
writeCacheKeyString(hash, "schema-version", schemaVersion)
```

Each key function keeps its existing signature and delegates to a `…ForSchema` variant that takes the schema explicitly, so a test can address the entry a build at another schema would have written. Production reaches them only through the existing entry points, which supply `SchemaVersion`. The read-time comparison stays exactly as it is — it is still the right defence for an entry written before the field existed or edited by hand.

### Test

`TestCacheKeysAddressTheSchemaVersion` in `internal/sem/cache_schema_version_test.go`. It asserts both key families give schema `1.1` and schema `9.9` different addresses, and that the production entry points address this build's schema and no other (without that second half the first proves nothing about what the tool actually reads).

### Mutation verification

Deleted both `writeCacheKeyString(hash, "schema-version", schemaVersion)` lines, `grep`-confirmed the deletion applied, rebuilt (the now-unused parameter still compiles):

```
GREP-AFTER-MUTATION (must be empty):
---build---
---test (MUST FAIL)---
  [FAIL] TestCacheKeysAddressTheSchemaVersion
     cache_schema_version_test.go:267: records cache addresses schema 1.1 and schema 9.9 to the same entry …
```

Restored, `grep`-confirmed both lines back, test passes. `go test ./internal/sem -run 'Cache|Snapshot'` — 268 passed. `go vet` and `gofmt -l` clean.

ashtom (approved):
Approved after syncing with current main and addressing all Entire findings across both on-disk cache families. Records and search builds now validate exact schema before writes; refreshed current-head findings are zero.

PR #155, opened 2026-08-28
fix(sem): enforce the ADR 0001 schema major when reading compact snapshots

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/114

## Problem

ADR 0001 (`docs/adr/0001-ga-schema-contract.md`) ratifies the GA schema contract and states the
rules for **reading** it back:

> 1. **Major = compatibility boundary.** Consumers refuse an unknown *major* version.
> 3. **Tolerant readers required.** Consumers ignore unknown fields within a supported major, and
>    warn (not fail) when they see a newer supported-major minor, since additive facts may have
>    been skipped.

`entire-graph` is itself such a consumer: `snapshot-query` reads compact snapshots back off disk,
and the bench preflight round-trips them. Neither honoured rule 1 or rule 3 — the header's
`schema_version` was decoded into `SnapshotHeader` and then never looked at.

`CompactSnapshotFormatVersion` does not cover this. It versions the **array encoding** (`["h",1,…]`),
not the record schema the header declares, and the two move independently.

## Reproduction (before this change)

Every one of these loaded silently and answered the query:

```
$ eg snapshot-query --input v_major2.ndjson  --symbol Beta   # header schema_version "2.0"
exit=0, 1 record
$ eg snapshot-query --input v_major9.ndjson  --symbol Beta   # "9.9"
exit=0, 1 record
$ eg snapshot-query --input v_missing.ndjson --symbol Beta   # no schema_version at all
exit=0, 1 record
$ eg snapshot-query --input v_junk.ndjson    --symbol Beta   # "abc"
exit=0, 1 record
```

A snapshot from a 2.x producer was decoded straight into this build's 1.x structs, where every
field that major renamed arrives as a zero value rather than an error.

## Fix

`CheckReadableSchemaVersion` (new, `internal/sem/schema_compat.go`) applies the ADR to a version
read off a persisted artifact, and `DecodeCompactSnapshot` calls it right after unmarshalling the
header. A version that cannot be placed on either side of the compatibility boundary — absent,
not `major.minor`, or an unreadable component — is refused for the same reason an unknown major is.

`LoadCompactSnapshot` carries the other half of the rule: a readable major written under a
**newer minor** loads, and reports `W_NEWER_SCHEMA_MINOR` on the index, because the additive facts
that minor introduced were not read. `snapshot-query` prints it to stderr, keeping stdout the
clean record stream.

After:

| header `schema_version` | before | after |
|---|---|---|
| `1.1` (this build) | served | served |
| `1.0` (older minor) | served | served |
| `1.7` (newer minor) | served, silently | served + `W_NEWER_SCHEMA_MINOR` on stderr |
| `2.0` / `9.9` | served, silently | refused, exit 1 |
| absent | served, silently | refused, exit 1 |
| `abc` / `1.1.0` | served, silently | refused, exit 1 |

## Tests

`internal/sem/schema_compat_test.go` — a ten-row table over `LoadCompactSnapshot`, driven by
rewriting the **serialized header line** so the absent-field case is expressible; plus direct
coverage of `DecodeCompactSnapshot` and of `CheckReadableSchemaVersion` against the package
constant (which also guards that the constant itself stays `major.minor`).

Confirmed binding by mutation: removing the `CheckReadableSchemaVersion` call from the decoder
fails the newer-major, absent, malformed and unreadable-minor rows; restoring it passes.

Existing compact decoder cases built bare `["h",1,{}]` headers, which now fail at the schema gate
before reaching the case under test. They go through a `compactHeaderLine()` helper that declares
a readable version, so each case still fails for its own reason.

## Verification

`go test -race -count=1 -run 'Compact|Schema|SCIP|Snapshot' ./internal/sem/ ./internal/cli/` — clean, zero data races.


suhaanthayyil:
Adversarial re-review. The enforcement is real (`TestLoadCompactSnapshotEnforcesSchemaMajor/"newer major"` fails on `main`, which loads a newer major silently), both ingest points are gated, the producer always stamps, and it fails closed on a missing or unparseable version. Three issues.

**1. INCOMPLETE — the warn-on-newer-minor half reaches a human on only one of two consumers.** `internal/bench/compact_preflight.go:103` also calls `LoadCompactSnapshot` and drops `index.SchemaWarnings` on the floor; only `snapshot_query.go:36` prints them. So a bench preflight against a newer-minor snapshot proceeds with no signal at all, which is precisely the case the ADR wants surfaced. Either surface it there too or make the warning a returned value the caller cannot ignore.

**2. A patch-level version becomes an unreadable artifact.** The parser hard-fails a three-component `"1.1.0"`. That is a one-way door: any future decision to move to semver-shaped versions makes every snapshot written after it unreadable by every binary shipped before it. If two components is the contract, say so in the ADR and reject at *write* time as well, so the failure is caught by the producer rather than by every downstream reader.

**3. Conflicts with #154 on the same field.** #154 requires exact equality on `SnapshotHeader.SchemaVersion` for cache-entry validity; this PR rules the same field compatible across a whole major. Both cite ADR 0001, neither references the other, and #154 does not use `CheckReadableSchemaVersion`. After both land a `"1.0"` header is refused by the cache and accepted by the loader. The two rules may both be correct, but they need to be an explicit written decision ("cache identity ≠ interchange compatibility") rather than an accident of which PR you read.

ashtom:

**Verdict: clean**

- Live PR head `6f09cf11`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `6f09cf11` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking.

**Guard checked in both directions (§4).** `schemaMajorMinor` rejects anything that is not exactly
`major.minor`. I checked this against the contract rather than assuming it is over-strict: ADR 0001
says "The wire format carries `schema_version` in `major.minor` form", and `SchemaVersion` is
`"1.1"`, so refusing a third component is the contract, not over-rejection. The newer-minor path
warns rather than fails, as clause 3 requires.

### Independent review state

suhaanthayyil:
Thanks — no changes needed on this branch, and I want to record why I agree with the one thing you checked hardest.

You checked `schemaMajorMinor`'s strictness against the contract rather than assuming it was over-strict, and reached the right answer: ADR 0001 says the wire format carries `schema_version` in `major.minor` form and `SchemaVersion` is `"1.1"`, so refusing a third component is the contract. That is the reading I intended.

Two items from my own earlier review are still open and are follow-ups, not blockers:

1. `internal/bench/compact_preflight.go:103` also calls `LoadCompactSnapshot` and drops `index.SchemaWarnings`; only `snapshot_query.go:36` prints them. So the warn-on-newer-minor half reaches a human on one of two consumers.
2. The relationship between this PR's rule (per-major compatibility) and #154's (exact equality, same field) is a real distinction — cache identity is exact, interchange compatibility is per-major — but it is currently an accident of which file you read rather than a written decision.

Happy to fold either into this PR if you would rather they not become follow-ups.

suhaanthayyil:
Follow-up on the open item from my last comment: the `SchemaVersion` question is now **settled and written down**, in an amendment to `docs/adr/0001-ga-schema-contract.md` landed on #156 (`a9db513f`).

This PR's rule is confirmed as correct and unchanged: interchange readability is **per-major, warning on a newer minor**, exactly as ADR 0001 clauses 1-3 require. #154's exact-equality rule on the same field is also correct, because it answers a different question — cache reuse, where there is no second party, no compatibility promise, and no migration path, so rebuilding is both cheap and always right.

The amendment states the distinction as a table and gives the reasoning: a tolerance band exists to avoid discarding data you cannot regenerate; a cache is by definition data you can. Neither PR needs to change.

Still open from my earlier pass, and genuinely this PR's: `internal/bench/compact_preflight.go:103` calls `LoadCompactSnapshot` and drops `index.SchemaWarnings`, so the newer-minor warning clause 3 requires reaches a human on only one of the two consumers. Happy to fold that in here if you want it in this PR rather than as a follow-up.

suhaanthayyil:
### The schema-warning contract reached a human on only one of its two consumers

`internal/bench/compact_preflight.go:103` called `sem.LoadCompactSnapshot` and dropped the returned `index.SchemaWarnings` on the floor.

There are exactly two production callers of that loader:

- `internal/cli/snapshot_query.go:31` — loops the warnings and prints `graph warning code=... detail=...` to stderr.
- `internal/bench/compact_preflight.go:103` — discarded them.

(The remaining hits — `internal/sem/golden_test.go:300`, `internal/sem/compact_snapshot_query_test.go`, `internal/cli/root_test.go:468`, `internal/cli/snapshot_query_test.go:25` — are tests, not consumers.)

So the ADR 0001 clause 3 signal, populated at `internal/sem/compact_snapshot_query.go:48-50`, was surfaced on one path and silently swallowed on the other.

### Why a hard error rather than a print

Surfacing it in the preflight the same way `snapshot-query` does would be the wrong shape for this call site.

`snapshot-query` reads an artifact **somebody else wrote**. Warn-and-serve is exactly right there, and it is unchanged.

The preflight does not. `runCompactPreflight` streams a snapshot out of `sem.StreamSnapshot` in-process and reads the compact bytes straight back; the header it reads is the one `leanHeader` (`internal/sem/provider.go:806`) just stamped with this build's own `sem.SchemaVersion`. A newer-minor signal there therefore cannot mean "a newer producer wrote this". It can only mean the writer and the reader inside one binary disagree about the schema, and the artifact was read incompletely.

That matters because the preflight is a **validation** path whose whole job is to certify the round-trip. Its two neighbouring integrity checks cannot catch this one:

- the canonical-hash comparison hashes decoded records against native records **both produced by this build's structs**, so fields a newer minor added are absent from both sides;
- `samePublicProjection` compares those same two struct populations, for the same reason.

Left as a warning, the function would go on to return `ProjectionMatchesNative: true` and publish `ndjson_bytes_per_projected_fact` / `compact_bytes_per_projected_fact` for facts it never read.

So it is refused, in the same class as the hash mismatch, the projection mismatch, and the query mismatch already in `verifyCompactPreflight` — and raised **before** the hash comparison, so the diagnosis names the schema rather than the downstream symptom it would cause.

No new reporting channel was invented. The error travels the one that already exists: `measureRepoWithPreflight` (`internal/bench/bench.go:183-186`) assigns `preflightErr` to `RepoMetrics.Error`, which is emitted as `"error"` in the JSON report and returned as `measureWorkerResponse.Error` by the isolated worker. Carrying the warning in the `compactPreflight` struct instead was rejected: `RepoMetrics` has no warning field, and adding one would create a channel nothing reads.

### Fix

`internal/bench/compact_preflight.go:109-130` — refuse when `loaded.SchemaWarnings` is non-empty, naming every warning's code and detail.

### Test

`TestCompactPreflightRejectsArtifactLoadedUnderANewerSchemaMinor` (`internal/bench/compact_preflight_test.go`).

It exercises the real newer-minor path rather than a stub. `compactPreflightInputs` re-runs the preflight's own producer over a temp repo and rewrites the header's `SchemaVersion` before it reaches the hasher, the projection, and the compact encoder — so the artifact is fully self-consistent and differs from a good one **only** in the version it declares. `newerSchemaMinorVersion` derives the version from `sem.SchemaVersion` (so it does not rot on a minor bump) and asserts via `sem.CheckReadableSchemaVersion` that it really is a newer minor of a readable major before using it, so the test cannot pass for the wrong reason. A control leg runs the identical pipeline at this build's own schema and requires it to pass, making the refusal attributable to the declared version and nothing else.

Rewriting the live header struct is enough here, unlike the sem fixture's serialized-line surgery in `internal/sem/schema_compat_test.go`: the only case that reaches this code is a version that is present and parses, since an absent or unplaceable one is already refused outright by the loader.

### Mutation verification

Production guard removed, test kept. The mutation compiles (`go vet` clean — the now-unused `strings` import was removed with the block).

Grep proof that the file actually changed:

```
$ grep -n 'SchemaWarnings|did not load losslessly|"strings"' internal/bench/compact_preflight.go
(no match: guard removed)
$ grep -c TestCompactPreflightRejectsArtifactLoadedUnderANewerSchemaMinor internal/bench/compact_preflight_test.go
1
```

Result with the guard removed:

```
--- FAIL: TestCompactPreflightRejectsArtifactLoadedUnderANewerSchemaMinor (0.03s)
    compact_preflight_test.go:141: preflight certified an artifact declaring schema 1.2,
        whose additive facts it never read
FAIL	github.com/entireio/entire-graph/internal/bench
```

Restored, and the same grep proving it back:

```
$ grep -n 'SchemaWarnings|did not load losslessly|"strings"' internal/bench/compact_preflight.go
9:	"strings"
120:	if len(loaded.SchemaWarnings) > 0 {
121:		details := make([]string, 0, len(loaded.SchemaWarnings))
122:		for _, warning := range loaded.SchemaWarnings {
126:			"compact preflight snapshot did not load losslessly: %s",
```

### Green

- `go test ./internal/bench/ -run 'TestCompactPreflight|TestMeasureRepo|TestRunMeasureWorker|TestIsolatedWorker|TestReducePhaseEvents|TestBuildReport'` — 19 passed
- `go test ./internal/sem/ -run 'TestLoadCompactSnapshot|TestDecodeCompactSnapshot|TestCheckReadableSchemaVersion'` — 14 passed
- `go test ./internal/cli/ -run 'TestSnapshotQuery|TestRunSnapshotQuery'` — 2 passed
- `gofmt -l`, `go vet`, `go build ./...` clean

No existing test was weakened or skipped.

ashtom:

**Verdict: clean gate, 3 open findings to resolve**

Head moved `6f09cf11` → `59ff5118`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### Entire trail findings — 3 open and current

```
01M1379P0CJK  Low     open    current    internal/cli/snapshot_query.go:39      The mandatory compatibility warning is written without checking `fmt.Fprintf`'s …
01M1379P09JX  Medium  open    current    internal/sem/compact_snapshot.go:1031  The streaming decoder discards the `newerMinor` result, so direct `DecodeCompact…
01M1379P065G  Medium  open    current    internal/sem/compact_snapshot.go:1031  `CheckReadableSchemaVersion` reports a newer 1.x minor as readable, but this dec…
```

### Verified vs. assumed

suhaanthayyil:
All three findings **valid and fixed** (`db70f709`). The two Mediums are one defect seen from two sides, so they are fixed together.

## Medium ×2 — `compact_snapshot.go:1031`, the decoder discards `newerMinor`

Correct. The decoder computed the signal *in order to decide readability* and then threw it away, so every caller that owed a reader the warning had to re-derive it from the header. `LoadCompactSnapshot` did exactly that (with a comment explaining why); `internal/bench/compact_preflight.go:134` is a second direct caller. A decoder that computes a mandatory signal and drops it hands each caller the same bug to rediscover.

`DecodeCompactSnapshot` now **returns** `[]ProviderWarning`, so the obligation is in the signature, and `LoadCompactSnapshot` consumes them instead of recomputing — one derivation rather than two that can drift.

**Why returned and not emitted through the record stream**, since that was the more obvious fix and it is wrong: `emit` feeds `SnapshotSemanticHasher.Add` and the preflight's `samePublicProjection`, both of which type-switch over the record kinds via `publicSnapshotRecord`. A `ProviderWarning` pushed through that channel would change the **canonical semantic hash** and break the projection comparison. A warning is metadata *about* the stream, not a record *in* it.

## Low — `snapshot_query.go:39`, the mandatory warning's write is unchecked

Correct, and it is the one place it can be caught. A dropped warning is silent by construction — nothing downstream can tell it was owed one — so serving the records after failing to deliver it leaves the consumer believing it got everything, which is precisely what clause 3 exists to prevent. The write is now checked and the command fails.

## Mutation verify (both compile; each confirmed applied by `grep` after the edit)

| mutation | result |
|---|---|
| decoder discards `newerMinor` again | **FAIL** — `TestDecodeCompactSnapshotReturnsTheNewerMinorWarning`, `TestLoadCompactSnapshotSurfacesTheDecoderWarning` |
| drop the mandatory-warning write check | **FAIL** — `TestSnapshotQueryFailsWhenTheMandatoryWarningCannotBeWritten` |
| unmutated | pass |

**One process note, since it changed the outcome.** On my first pass the second mutation was **not caught** — no test covered a failing stderr, and I had written the fix without one. Rather than report it as an untestable negative I injected a failing writer through `Options.Stderr` and built the missing coverage; the mutation is caught now. Along the way the first version of that test failed for the wrong reason (`snapshot-query requires at least --symbol or --from` — the flag parse runs before the warning loop), which I only noticed because the test asserts on the error *text* rather than just `err != nil`.

`go test ./internal/sem/` 78.5s, `./internal/cli/` 23.6s, `./internal/bench/` — all pass. Eight call sites updated for the new signature; no behaviour change for any of them beyond receiving the value.

ashtom (approved):
Approved after syncing with current main and addressing all current Entire findings. Newer-minor compact additions are tolerated without weakening current-schema framing, schema-version comparison is word-size independent, and the refreshed review reports zero current findings.

PR #156, opened 2026-08-28
test(sem): freeze the persisted Result wire shape against silent schema drift

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/115

## Problem

`sem.Result` is the persisted diff/analyze/checkpoint payload, but its JSON wire shape was not frozen. A field could be added, removed, renamed, retyped, or have wire-affecting tag behavior changed while `SchemaVersion` remained `1.1` and the existing suite stayed green.

A simple reflected field list was not sufficient: reachable nested types, custom JSON/text marshalers, `omitzero` hooks, anonymous-field promotion, tagged-field dominance, interface-backed runtime values, named non-struct underlyings, and same-named types from different import paths can all change emitted bytes without changing a naive descriptor.

## Fix

Add a test-only, closed wire-contract guard for the persisted `Result` graph:

- freeze canonical JSON field descriptors, including valid taggedness, anonymous status, all tag options, and recursively full-path-qualified types;
- freeze every reachable user-defined named type, including named scalar/container underlyings;
- require the frozen shape digest and exact `SchemaVersion` pin to move together;
- reject reachable interfaces whose runtime concrete union is not statically frozen;
- reject value- and pointer-receiver `json.Marshaler` and `encoding.TextMarshaler` implementations;
- reject `IsZero() bool` hooks on exact `omitzero` fields;
- exclude only fields that `encoding/json` excludes, including the exact `json:"-"` rule while preserving `json:"-,"`.

ADR 0001 now records these enforcement boundaries. This PR changes tests and documentation only; the production `Result` bytes and `SchemaVersion` remain unchanged.

## Findings addressed

The branch now covers every finding raised in the Entire trail: nested `ProviderWarning` shape coverage, exact schema/digest binding, all JSON tag options, ignored-field reachability, and custom-marshaler escape hatches. Follow-up review also closed text marshaling, anonymous/tagged promotion, named non-struct, `omitzero`, interface, and full import-path collision cases.

## Verification

- focused 12-test contract suite: pass
- `go test ./internal/sem -count=1`: pass on the exact merge with current `main` (476.918s)
- `git diff --check`: pass

Full cross-platform CI and a fresh Entire review are running on head `ea09a12b73e09ddac0099295b00e8d85c11dc363`.


suhaanthayyil:
Adversarial re-review. A shape tripwire is a good idea, but as written this one does not freeze the wire shape it claims to. Three gaps, all live today.

**1. `omitempty` is stripped, and ADR 0001 treats optional→required as breaking.** `jsonWireShape` does `name, _, _ := strings.Cut(field.Tag.Get("json"), ",")`, so the option half of the tag is discarded. `EntityChange` carries nine `,omitempty` fields and `Result` three (`model.go:54-99`). Delete `,omitempty` from `old_path` and every record's emitted JSON changes — consumers that keyed on the field's absence break — and this test stays green. Capture the full tag, or at minimum a normalized `name+omitempty` pair.

**2. Nested persisted types are not frozen.** `Result.Warnings []ProviderWarning` is part of the persisted `Result` wire, but `ProviderWarning`'s own fields appear in no shape list. Same for any other named struct reachable from `Result`. Either walk reachable struct types transitively or state explicitly which types the contract covers.

**3. Nothing binds the shape to the version.** `TestResultCarriesThisBuildsSchemaVersion` asserts only `strings.HasPrefix(SchemaVersion, "1.")`. You can edit `resultWireShape` and not bump the minor and everything passes. That makes this a "someone looked at it" marker rather than a contract. Hashing the shape list and asserting the digest is pinned alongside the exact `SchemaVersion` string would close it — and would also give #154 the shape-to-version binding it currently lacks.

Also worth deciding jointly with #154 and #155: after all three land, `SchemaVersion` simultaneously gates the `Result` wire shape, the compact-snapshot record shape, and cache-entry identity. A `Result`-only additive field then requires a minor bump under this PR's ADR reading, which under #154 invalidates every user's provider-records and search-snapshot caches — a full rebuild forced by a change to an artifact those caches do not contain.

ashtom:

**Verdict: clean — and the frozen contract genuinely discriminates**

- Live PR head `a3b727d3`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `a3b727d3` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|
| `go test ./internal/sem/` | merged onto main | pass |
| mutation: rename a wire field | merged onto main | **fail** — caught |
| mutation: add an optional field | merged onto main | **fail** — caught |

### Findings

Nothing blocking.

**This is a test-only PR, so the ordinary "revert the fix" check does not apply.** I ran the
stronger check instead — mutating the wire shape and confirming the test catches it. Both
directions fail as they should:

```
# baseline
ok    github.com/entireio/entire-graph/internal/sem

# mutation A: rename a wire field   Base `json:"base"` -> `json:"base_rev"`
--- FAIL: TestResultWireShapeIsFrozen
    want: [checkpoint string base string head string files ... producer_version string]

# mutation B: add an optional field  Sneaky string `json:"sneaky,omitempty"`
--- FAIL: TestResultWireShapeIsFrozen
```

The test pins the *property* (the full field/type list) rather than an incidental spelling, and the
failure message tells the next person which ADR 0001 clause applies. This is the check I most often
find missing; it is done correctly here.

### Independent review state

suhaanthayyil:
Thanks — and your mutation check is the right one to run on a test-only PR. Two of the three gaps from my own earlier pass were real and are now closed in `b0c2f9c9`; the third is a follow-up.

**Why your mutation B passed but the contract still had a hole.** You mutated by ADDING an optional field, which changes the field LIST and was caught. The gap was the opposite direction: `jsonWireShape` did `name, _, _ := strings.Cut(tag, ",")` and threw the option half away, so DELETING `,omitempty` from an existing field left the list byte-identical. That is a real wire change — the field stops being absent for a zero value and starts arriving as `""`/`0`/`null` — and ADR 0001 classes optional→required as breaking.

Measured on the published head of this branch:

```
# published test, mutate FileChange.OldPath `json:"old_path,omitempty"` -> `json:"old_path"`
go test ./internal/sem/ -run TestResultWireShapeIsFrozen  ->  PASS   (missed)

# published test, rename ProviderWarning.Detail `json:"detail,omitempty"` -> `json:"details,omitempty"`
go test ./internal/sem/ -run TestResultWireShapeIsFrozen  ->  PASS   (missed)
```

The second is gap two: `ProviderWarning` is `Result.Warnings`' element type, is persisted with the rest of the payload, and appeared in no shape list, so its fields could be renamed in silence.

**Fixed.** `omitempty` is now part of the frozen shape (`name type omitempty`), `ProviderWarning` is frozen, and `TestEveryPersistedTypeReachableFromResultIsFrozen` walks `Result`'s type graph in both directions — a new named struct reaching the payload fails until someone freezes it and decides what its arrival means for `SchemaVersion`, and a frozen type that stops being reachable fails as the breaking change it is. That is what stops this becoming a snapshot of whichever types someone thought of.

Mutation verify, all compiling:

| mutation | before | after |
|---|---|---|
| drop `,omitempty` from `FileChange.OldPath` | PASS (missed) | **FAIL** |
| rename `ProviderWarning.Detail`'s wire name | PASS (missed) | **FAIL** |
| add a new reachable named struct type | n/a | **FAIL** |
| unmutated | PASS | PASS |

**Still open** (gap three, deliberately not in this commit): nothing binds the shape to the version. `TestResultCarriesThisBuildsSchemaVersion` asserts only `strings.HasPrefix(SchemaVersion, "1.")`, so `resultWireShape` can be edited without a minor bump and everything stays green. Closing it means hashing the shape list and pinning the digest beside an exact `SchemaVersion` — which is a decision about what that constant means, and #154 and #155 are each making a different one about the same string. I would rather settle that once across the three than three times.

suhaanthayyil:
Closing the third gap I left open last round — the one where "the shape is frozen" and "the version is stamped" did not add up to a contract. `a9db513f`.

## The decision, written down

I said this needed settling once across #154/#155/#156 rather than three times. It is now an amendment in `docs/adr/0001-ga-schema-contract.md`. The short version: **`schema_version` answers two different questions, and the two rules only looked contradictory because neither was written down.**

| question | rule | enforced by |
|---|---|---|
| may I *read* bytes another build wrote? | same major; warn on newer minor | `CheckReadableSchemaVersion` (#155) |
| may I *reuse* a cache entry I wrote? | exact match; absent fails closed | cache-entry validity (#154) |
| may this payload's shape change? | additive within a major; a break needs `2.0` | frozen shape + pinned digest (#156) |

The reason they differ is not stylistic. A tolerance band exists to avoid discarding data you **cannot regenerate**; a cache is by definition data you **can**. Reading the per-major rule as governing cache validity would serve entries written by a build whose record shape has since changed. Reading the exact rule as governing interchange would refuse snapshots the ADR promises to accept. Both PRs were right; only the absence of a written distinction was wrong.

## The binding

`resultWireShapeDigest()` hashes the frozen shape, and `TestResultWireShapeIsBoundToTheSchemaVersion` pins that digest **next to** the exact `SchemaVersion` string. You can no longer move the shape without editing the version line in the same test — which is exactly the moment the ADR question gets asked. The failure message spells out the three cases and what each requires.

## Mutation verify (all compiling)

| mutation | `…IsFrozen` | `…IsBoundToTheSchemaVersion` |
|---|---|---|
| add an optional field, leave the frozen list alone | **FAIL** | pass |
| **the hole:** update the frozen list to match, no version bump | pass | **FAIL** |
| bump `SchemaVersion` 1.1→1.2 without re-pinning | pass | **FAIL** (`have "1.2", pinned "1.1"`) |
| unmutated | pass | pass |

Row two is the one that matters: that is precisely the case that stayed green before this commit.

**One honest note on method.** My first attempt at the third mutation was a silent no-op — my `sed` pattern missed the aligned `SchemaVersion         = "1.1"` declaration, so nothing changed and the test "passed" the mutation. I caught it by checking the declaration rather than trusting the result, and re-ran it correctly. Flagging it because a no-op mutation that reports as "not caught" is the exact failure mode that makes mutation testing worthless, and it would have been easy to write it up as a real gap.

ashtom:

Head moved `a3b727d3` → `a9db513f`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|
| revert-the-fix check | — | N/A — test-only PR |

**Re-confirmed:** this is still a test-only PR, so the revert check does not apply. I re-ran the
mutation check on the new head and the frozen shape still discriminates — renaming a wire field and
adding an optional field both fail the test.

```
01M17GYP7SD0  Medium  open    current    internal/sem/result_shape_contract_test.go:141  The shape records only `omitempty`, so wire-affecting options such as `string` a…
01M17EMWN4YH  Low     open    current    internal/sem/result_shape_contract_test.go:169  The reachability walk follows every exported field, including fields tagged `jso…
```

### Verified vs. assumed

suhaanthayyil:
Both findings are **valid and fixed** (`e24efe1e`), and both are one level below holes this PR had already closed — which is the right level for them to have been found at.

## Medium — the shape records only `omitempty`

Correct. `,string` makes a numeric or boolean field marshal as a **quoted** string, so adding or removing it changes what every consumer parses; `,omitzero` changes when the field is emitted. Neither was visible to the freeze.

Measured, adding `,string` to `EntityChange.DependentsCount` (`3` → `"3"` on the wire):

| helper | result |
|---|---|
| pre-fix (omitempty only) | **PASS — the wire change was missed** |
| fixed (full option list) | **FAIL — caught** |

Options are now recorded in full and **sorted**, so the shape is a property of the tag rather than of the order someone wrote its options in. That is a small point but it matters here: an unsorted list would make `json:"x,omitempty,string"` and `json:"x,string,omitempty"` read as different wire shapes when they are identical.

## Low — the reachability walk follows `json:"-"` fields

Also correct, and the consequence is the opposite direction from the one above: it makes the contract **over**-strict. A type reachable only through a `json:"-"` field never reaches the wire, so demanding a frozen shape for it would fail the freeze for a change no consumer can observe.

Measured, adding a struct reachable only through a `json:"-"` field:

| walk | result |
|---|---|
| pre-fix | **FAIL — wrongly demanded a shape for an unobservable type** |
| fixed | pass — correctly ignored |

## No golden churn

No field in the current payload uses `,string` or `,omitzero`, so the recorded shape is byte-identical and the **pinned digest is unchanged**. This commit adds the ability to see those options; it does not move the contract.

`go test ./internal/sem/ -run 'TestResult|TestEveryPersisted|TestCompact|TestLoadCompact'` — 46 pass. All four mutations above compile, and I confirmed by `grep` after each edit that the mutation actually applied.

ashtom (approved):
Reviewed the exact current head after all Entire findings were fixed. The canonical Result wire-shape guard, ADR update, focused regression set, and merged-tree internal/sem suite are clean.

PR #157, opened 2026-08-28
feat(search): version the search response envelope

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/116

## Problem

`search` was the only machine-readable surface with **no version discriminator at all**. Measured
on origin/main against one fixture repository, one command per row:

| discriminator | surfaces |
|---|---|
| `schema_version` | `diff`, `commit`, `capabilities`, `symbols`, `edges`, `snapshot`, `snapshot --format compact-ndjson` |
| `format_version` | `impact`, `neighbors`, `def`, `index`, `stats` |
| **neither** | **`search --format json`**, **`search --format ndjson`** |

```
$ eg search --repo . --query "widget render" --format json | python3 -c "import sys,json;print(list(json.load(sys.stdin).keys()))"
['query', 'repo_root', 'commit', 'tree', 'profile', 'results', 'verify_command', 'stats', 'warnings', 'partial_failures', 'completeness']
```

`search` is the documented entry point ("start here") and JSON is its **default** format, so the
one surface a consumer is most likely to parse was the one it could not version-check.

## Fix

`SearchResponse` gains `format_version`, on the same axis its five sibling query commands already
emit, and the ndjson `search_header` record carries it too so a streaming consumer can branch
before parsing a result record.

Deliberately **not** `sem.SchemaVersion`. This is a query response, not one of the persisted `1.x`
record artifacts ADR 0001 freezes. Conflating the two would make every search-payload change read
as a change to the ingestion schema — and search's payload blocks evolve far faster than the
snapshot records do. `SearchFormatVersion` starts at `1`, matching `impact`/`def`/`neighbors`/
`index`/`stats`.

`SearchRepository` stamps it at the **one** place a content-bearing response leaves the package,
mirroring how `AnalyzeGitRangeWithOptions` stamps `SchemaVersion`. That was not cosmetic: the
first version of this change stamped both success returns, and a mutation test showed only one of
them was actually reachable from the test fixtures — a third success path added later would have
emitted `format_version: 0`, which a consumer cannot distinguish from "field absent".

After:

```
$ eg search --repo . --query "widget render" --format json | head -c 60
{"format_version":1,"query":"widget render","repo_root":"/...

$ eg search --repo . --query "widget render" --format ndjson | head -1 | head -c 90
{"commit":"f34f61f...","format_version":1,"profile":"fast",...,"record_type":"search_header",...
```

## Tests

- `internal/cli/search_format_version_test.go` — both output formats carry the field, decoded into
  a `*int` so "absent" is distinguishable from `0`
- `internal/sem/search_format_version_test.go` — the real provider path stamps it, over both a
  hit and a no-result query

Confirmed binding by mutation: deleting the stamp fails the sem test on both subtests; deleting
the ndjson header field fails the cli test. Restoring each passes.

## Verification

`go test -race -count=1 -run 'Search|Format' ./internal/sem/ ./internal/cli/` — clean, zero data races.

### DIFF
--- a/internal/cli/search.go
+++ b/internal/cli/search.go
@@ -468,11 +468,15 @@ func writeNdjsonSearch(out interface{ Write([]byte) (int, error) }, response sem
 		encoder.SetEscapeHTML(false)
 		if err := encoder.Encode(map[string]any{
 			"record_type": "search_header",
-			"query":       response.Query,
-			"repo_root":   response.RepoRoot,
-			"commit":      response.Commit,
-			"tree":        response.Tree,
-			"profile":     response.Profile,
+			// The header record carries the envelope version for the same reason
+			// the JSON payload does: a consumer reading the stream must be able to
+			// branch on the shape before it parses a single result record.
+			"format_version": response.FormatVersion,
+			"query":          response.Query,
+			"repo_root":      response.RepoRoot,
+			"commit":         response.Commit,
+			"tree":           response.Tree,
+			"profile":        response.Profile,
 		}); err != nil {
 			return err
 		}
--- a/internal/cli/search_format_version_test.go
+++ b/internal/cli/search_format_version_test.go
@@ -1,0 +1,71 @@
+package cli
+
+import (
+	"bytes"
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/entireio/entire-graph/internal/sem"
+)
+
+// Every other machine-readable surface tells a consumer which shape it is
+// looking at. Measured on origin/main against one fixture repository:
+//
+//	schema_version : diff, commit, capabilities, symbols, edges, snapshot,
+//	                 compact-ndjson
+//	format_version : impact, neighbors, def, index, stats
+//	NEITHER        : search --format json, search --format ndjson
+//
+// search is the documented entry point ("start here") and JSON is its DEFAULT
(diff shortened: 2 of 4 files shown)


ashtom:

**Verdict: clean**

- Live PR head `dd6976de`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `dd6976de` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

### Independent review state

suhaanthayyil:
Thanks — nothing to change here. Confirmed on my side too: the envelope version is stamped at the one place the response is built, and reverting it fails the new tests.

One cross-PR note for whoever sequences these, since it touches the same constant three PRs now read: after #154, #155 and this one land, `SchemaVersion` simultaneously gates the search response envelope, the compact-snapshot record shape, cache-entry identity, and (via #156) the persisted `Result` wire shape. That is four contracts on one string. It works, but the coupling means an additive change to any one of them forces a minor bump that invalidates every user's on-disk cache under #154's exact-equality rule. Worth one written line about which of those are the same contract and which merely share a spelling — a follow-up, not a blocker here.

PR #158, opened 2026-08-28
fix(cli): stop init-agents writing through a link into the git directory

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/117

## Impact

`init-agents` follows a repository-committed symlink into the repository's own git directory and writes the managed block through it. A hostile checkout that commits

    CLAUDE.md -> .git/config

gets the pointer block appended to git's own config, and git then refuses to operate on the repository at all:

    fatal: bad config line 9 in file .git/config

`.git/hooks/*` is the same primitive aimed somewhere worse. The tool also reports the write as `updated .../CLAUDE.md`, so the file it actually damaged is never named.

This is the same class as #136/#137 at a call site those did not reach. #137 confined `init-agents` to the repository root; `--report` / `--record-baseline` (#136) additionally refuse a symlinked component, and `outputpath.go` says why in as many words: os.Root "stops the write leaving the root, but it happily FOLLOWS a link that stays inside it (on every platform, Windows included), and `.git/config` is inside it." `init-agents` kept only the containment half, and `runInitAgents` records the reasoning that made that look sufficient — a link "that stays inside the repository is still followed — which is what the documented AGENTS.md/CLAUDE.md alias needs". That is true of repository *content* and false of `.git`, which is inside the root and is not content.

`docs/trust-and-security.md` claimed `init-agents` "writes through exactly three repository paths". It wrote through a fourth.

## Repro (before this change)

    mkdir -p /tmp/hostile && cd /tmp/hostile && git init -q .
    printf 'package main\nfunc main(){}\n' > main.go
    ln -s .git/config CLAUDE.md
    git add -A && git commit -qm x

    entire-graph init-agents --repo .
    # wrote    /tmp/hostile/.entire/graph-agent.md
    # updated  /tmp/hostile/AGENTS.md
    # updated  /tmp/hostile/CLAUDE.md      <- actually .git/config

    git status
    # fatal: bad config line 9 in file .git/config

`ln -s .GIT/config CLAUDE.md` does the same on macOS and Windows, and is the reason the fix folds case.

## Fix

Refuse a managed target whose resolution lands inside a git directory — `.git` at any depth, including a nested checkout's `vendor/dep/.git` and a linked worktree's `.git` pointer file.

The check sits on `resolveContainedNameWithOptions`, the single resolver every stat, read and write of a managed target goes through, rather than on the preflight. `ensureContainedInRepo` is explicitly documented as not atomic with the write, so a preflight-only check would be passed by a link swapped in afterwards; on the resolver each operation re-asks, and the one that finds the swapped link refuses it. It also covers the directory target, where `.entire -> .git` would otherwise make the git directory the guide's home.

The comparison folds case, which is the second half of the bug. A symlink target is text the *repository* chose, and macOS and Windows resolve it case-insensitively, so `CLAUDE.md -> .GIT/config` opens `.git/config` and walks past an exact `".git"` comparison. Folding where the filesystem does not can only refuse a path through a directory genuinely named `.GIT`, which is not a git directory and not an instruction file's home either, so the trade fails closed.

"Inside a git directory" is not restated. `sem.hasGitDirComponent` already answers it for the indexer — whole component, any depth, `.git` files included — and the new `sem.PathLandsInGitDir` sits beside it, reusing that rule and documenting the one deliberate difference. `hasGitDirComponent` itself stays exact: its only other caller, `skipVendoredDir`, judges names that came out of a directory *walk*, where the kernel's matching rules never enter, so folding there would change indexing semantics for no gain. A writer that will not read inside a git directory (#132) must not write there either.

The refusal is classified apart from `isRootEscape`: the link stays comfortably inside the root, so calling it an escape would send the reader hunting a link that leaves when none does.

Everything `docs/agents.md` documents still works and is now pinned by tests — the shared-instruction-file alias (`CLAUDE.md -> AGENTS.md`) and an in-repository symlink to an ordinary file (`AGENTS.md -> docs/guide.md`). Both docs are corrected.

## Mutation verification

Two mutations, because the fix has two independent halves.

**A — disable the predicate** (`if false && strings.EqualFold(...)`): all 6 refusal subtests fail with `init-agents wrote through a link into the git directory`, as does the directory-alias test. The 2 alias subtests keep passing in both states, so they are not vacuous and the refusal is not achieving them by breaking the feature.

**B — keep the predicate, drop only the fold** (`component == ".git"`): fails exactly the 2 subtests that exist for it, `uppercase git spelling` and `mixed case git spelling`, and nothing else. The fold is load-bearing and separately pinned.

The refusal tests also assert no partial install: the other managed files must not exist after the error. The case-spelling cases probe the filesystem for case-insensitivity and skip on a case-sensitive volume rather than assuming from GOOS.

## Gate

`go build ./...`, `go vet ./...` and `gofmt` are clean, and `go test ./...` passed in full before the final case-fold commit; the init-agents suite (`-run TestInitAgents`, 139 tests) is green in 5s.

The full `-race` gate was not completed locally. `mise run check` fails on this machine with `panic: test timed out after 10m0s` in both `internal/cli` and `internal/sem` and zero `--- FAIL` subtests, which is Go's default per-package timeout under a load average around 20. The control settles it: the same `go test -timeout 9m ./internal/cli/` on an unmodified `origin/main` worktree times out at **540.7s**, against **542.7s** on this branch. The timeout is environmental and pre-existing, not this change. CI on a clean runner is the gate that matters here.

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -88,7 +88,14 @@ by its name, so a repository whose administrative directory is not called `.git`
 — `git init --separate-git-dir=admin`, or a checkout driven by `GIT_DIR` — is
 covered by the same refusal.
 
-Hard links are also
+The landing must also be an agent-instruction file. An alias exists so that
+`AGENTS.md` and `CLAUDE.md` can share one instruction file, so a target that is
+markdown by extension, or a rules file such as `.cursorrules`, is written; a
+target that is some other existing file — a `Makefile`, `.envrc`, or
+`.github/workflows/ci.yml` — is refused rather than having a managed block
+appended to it. A target that does not exist yet is still created, which is what
+the dangling-alias case below relies on, and a target this command wrote on an
+earlier run stays writable whatever it is named. Hard links are also
 supported, but every hard-linked pathname names the same inode: an update
 through `AGENTS.md` or `CLAUDE.md` is therefore visible through any other hard
 link to that file, including one outside the project. A directory, named pipe,
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -83,7 +83,12 @@ directory — `.git` at any depth, including a nested checkout's and a linked
 worktree's `.git` pointer — is refused as well. `.git` is inside the project root
 but is not project content, and no instruction file belongs there; writing a
 managed block into `config` or a hook would corrupt the repository rather than
-configure an agent. Hard links are also
+configure an agent. The git directory is recognised by its structure rather than
+by its name, so a repository whose administrative directory is not called `.git`
+— `git init --separate-git-dir=admin`, or a checkout driven by `GIT_DIR` — is
+covered by the same refusal.
+
+Hard links are also
 supported, but every hard-linked pathname names the same inode: an update
 through `AGENTS.md` or `CLAUDE.md` is therefore visible through any other hard
 link to that file, including one outside the project. A directory, named pipe,
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -77,7 +77,13 @@ content from that validated snapshot.
 Each path must be missing or resolve to a regular file. Symlinks to regular
 files are supported, with the target written either relatively or as an
 absolute path, as long as it stays inside the project root; a symlink that
-resolves outside is refused and nothing is installed. Hard links are also
+resolves outside is refused and nothing is installed. Staying inside the project
+root is necessary but not sufficient: a symlink that lands inside a git
+directory — `.git` at any depth, including a nested checkout's and a linked
+worktree's `.git` pointer — is refused as well. `.git` is inside the project root
+but is not project content, and no instruction file belongs there; writing a
+managed block into `config` or a hook would corrupt the repository rather than
+configure an agent. Hard links are also
 supported, but every hard-linked pathname names the same inode: an update
 through `AGENTS.md` or `CLAUDE.md` is therefore visible through any other hard
 link to that file, including one outside the project. A directory, named pipe,
--- a/docs/trust-and-security.md
+++ b/docs/trust-and-security.md
@@ -179,9 +179,14 @@ these rules existed misses instead of re-emitting the paths it named.
   rebuild, nothing else. Queries never modify repository source files.
 - `init-agents` writes through exactly three repository paths, disclosed in
   [agent activation](agents.md): `.entire/graph-agent.md` and managed blocks
-  in `AGENTS.md` and `CLAUDE.md`. A hard-linked pathname elsewhere names the
-  same inode and therefore observes the same update; the activation guide
-  calls out that filesystem property explicitly.
+  in `AGENTS.md` and `CLAUDE.md`. A repository-committed symlink at one of
+  those paths may redirect the write to another file, which is what makes the
+  alias support in the activation guide work; the redirection is confined to
+  the project root, and additionally refused when it lands in a git directory,
+  so a hostile checkout cannot aim a managed block at `.git/config` or a hook.
+  A hard-linked pathname elsewhere names the same inode and therefore observes
+  the same update; the activation guide calls out that filesystem property
+  explicitly.
 - `index --report <path>` writes a Markdown graph report to the path you give
   it.
 - `verify --record-baseline <path>` creates parent directories as needed and
--- a/docs/trust-and-security.md
+++ b/docs/trust-and-security.md
@@ -182,8 +182,10 @@ these rules existed misses instead of re-emitting the paths it named.
   in `AGENTS.md` and `CLAUDE.md`. A repository-committed symlink at one of
   those paths may redirect the write to another file, which is what makes the
   alias support in the activation guide work; the redirection is confined to
-  the project root, and additionally refused when it lands in a git directory,
-  so a hostile checkout cannot aim a managed block at `.git/config` or a hook.
+  the project root, and additionally refused when it lands in a git directory —
+  recognised by structure, so an administrative directory not named `.git` is
+  covered too — so a hostile checkout cannot aim a managed block at
+  `.git/config` or a hook.
   A hard-linked pathname elsewhere names the same inode and therefore observes
   the same update; the activation guide calls out that filesystem property
   explicitly.
--- a/docs/trust-and-security.md
+++ b/docs/trust-and-security.md
@@ -184,8 +184,9 @@ these rules existed misses instead of re-emitting the paths it named.
   alias support in the activation guide work; the redirection is confined to
   the project root, and additionally refused when it lands in a git directory —
   recognised by structure, so an administrative directory not named `.git` is
-  covered too — so a hostile checkout cannot aim a managed block at
-  `.git/config` or a hook.
+  covered too — or on an existing file that is not an agent-instruction file, so
+  a hostile checkout cannot aim a managed block at `.git/config`, a hook, a
+  `Makefile`, `.envrc`, or a CI workflow.
   A hard-linked pathname elsewhere names the same inode and therefore observes
   the same update; the activation guide calls out that filesystem property
   explicitly.
--- a/internal/cli/agents.go
+++ b/internal/cli/agents.go
@@ -252,7 +252,7 @@ func runInitAgents(opts Options, args []string) error {
 	if guideWasMissing {
 		guideCreated, err = writeNewContainedFile(repoRoot, guideName, []byte(agentGuide), 0o644)
 	} else {
-		err = writeContainedFile(repoRoot, guideName, []byte(agentGuide), 0o644)
+		err = writeContainedFile(repoRoot, guideName, []byte(agentGuide), 0o644, guideName, agentsName, claudeName)
 	}
 	if err != nil {
 		return rollback(err, guideCreated)
@@ -265,7 +265,7 @@ func runInitAgents(opts Options, args []string) error {
 	}
 	fmt.Fprintf(opts.Stdout, "wrote %s\n", guidePath)
 
-	if err := writeContainedFile(repoRoot, agentsName, agentsContent, 0o644); err != nil {
+	if err := writeContainedFile(repoRoot, agentsName, agentsContent, 0o644, guideName, agentsName, claudeName); err != nil {
 		return fmt.Errorf("init-agents: AGENTS.md: %w", err)
 	}
 	fmt.Fprintf(opts.Stdout, "updated %s\n", agentsPath)
@@ -289,7 +289,7 @@ func runInitAgents(opts Options, args []string) error {
 	if err == nil && os.SameFile(agentsInfo, claudeInfo) {
 		return nil
 	}
-	if err := writeContainedFile(repoRoot, claudeName, claudeContent, 0o644); err != nil {
+	if err := writeContainedFile(repoRoot, claudeName, claudeContent, 0o644, guideName, agentsName, claudeName); err != nil {
 		return fmt.Errorf("init-agents: CLAUDE.md: %w", err)
 	}
 	fmt.Fprintf(opts.Stdout, "updated %s\n", claudePath)
@@ -1365,13 +1365,34 @@ func readContainedFile(root *os.Root, name string, limit int64) ([]byte, error) 
 
 // writeContainedFile is os.WriteFile confined to root. The perm argument applies only when the
 // file is created, matching os.WriteFile, so an existing file keeps its mode.
-func writeContainedFile(root *os.Root, name string, content []byte, perm os.FileMode) error {
+func writeContainedFile(root *os.Root, name string, content []byte, perm os.FileMode, managed ...string) error {
 	resolved, err := resolveContainedName(root, name)
 	if err != nil {
 		return err
 	}
-	file, err := root.OpenFile(resolved, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm)
+	// O_TRUNC is deliberately NOT in this open, and the file is judged from the
+	// HANDLE rather than from the resolved name.
+	//
+	// Both follow from the same gap. resolveContainedName decides on a path, and
+	// the path is not the thing written: OpenFile re-resolves it, so a link
+	// swapped in between the two is followed (os.Root refuses a link that
+	// ESCAPES the root, atomically, but follows one that stays inside — and
+	// `.git` is inside). Judging the open handle removes the window, because the
+	// handle IS what the write lands on.
+	//
+	// And truncation is already the damage. Opening `.git/config` with O_TRUNC
+	// destroys it before any guard below can object, so the truncate happens
+	// only after the handle has been accepted.
+	file, err := root.OpenFile(resolved, os.O_WRONLY|os.O_CREATE, perm)
 	if err != nil {
+		return err
+	}
+	if err := refuseSharedInode(root, file, name, resolved, managed); err != nil {
+		_ = file.Close()
+		return err
+	}
+	if err := file.Truncate(0); err != nil {
+		_ = file.Close()
 		return err
 	}
 	_, writeErr := file.Write(content)
@@ -1381,6 +1402,75 @@ func writeContainedFile(root *os.Root, name string, content []byte, perm os.File
 	}
 	return closeErr
 }
+
+// refuseSharedInode rejects an open managed target that is a HARD link.
+//
+// Every other guard here reasons about a path, and a hard link has no path to
+// reason about. `ln .git/config CLAUDE.md` gives git's config a second name in
+// the working tree: it resolves to "CLAUDE.md", carries no `.git` component, and
+// Lstat reports an ordinary regular file, so PathLandsInGitDir cannot see it.
+// Writing the managed block through that name appends to git's own config —
+// exactly the corruption the git-directory refusal exists to prevent, reached by
+// a route it does not cover. Measured before this guard: `init-agents` exited 0
+// and `.git/config` was rewritten.
+//
+// The rule is link count rather than "is it inside .git", because the handle
+// cannot be asked where else its inode is named. That means it also refuses a
+// hard link to a harmless file. An instruction file with a second name is
+// vanishingly rare, sharing one is what the DOCUMENTED alias (a symlink) is for,
+// and the failure mode on the other side is silent corruption of the repository
+// — so this fails closed, on the same reasoning as the case-folded `.GIT`
+// refusal above.
+func refuseSharedInode(root *os.Root, file *os.File, name, resolved string, managed []string) error {
+	info, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	links, known := fileLinkCount(info)
+	if !known || links <= 1 {
+		return nil
+	}
+	// A shared inode is only dangerous if one of its OTHER names is somewhere
+	// this command must not write. Hard-linking two managed instruction files
+	// together is documented and supported, so the question is not "is this
+	// file linked" but "is every one of its names a managed target".
+	//
+	// Counting is what answers that without being able to enumerate an inode's
+	// names: if as many managed targets share this inode as the inode has
+	// names, then all of its names are accounted for and all of them are ours.
+	// One unaccounted name means some other path — `.git/config`, a hook —
+	// reaches the same bytes, and the write must be refused.
+	device, inode, identified := fileIdentity(info)
+	if !identified {
+		return nil
+	}
+	accounted := uint64(0)
+	for _, candidate := range managed {
+		candidateResolved, resolveErr := resolveContainedName(root, candidate)
+		if resolveErr != nil {
+			continue
+		}
+		candidateInfo, statErr := root.Stat(candidateResolved)
+		if statErr != nil {
+			continue
+		}
+		candidateDevice, candidateInode, candidateIdentified := fileIdentity(candidateInfo)
+		if candidateIdentified && candidateDevice == device && candidateInode == inode {
+			accounted++
+		}
+	}
+	if accounted >= links {
+		return nil
+	}
+	return fmt.Errorf(
+		"%w: %s resolves to %s, which has %d names but only %d of them are managed instruction files; an unaccounted name may reach a file this command must never write, such as one in the git directory",
+		errSharedInodeManagedTarget, name, filepath.ToSlash(resolved), links, accounted,
+	)
+}
+
+// errSharedInodeManagedTarget marks a managed target that shares its inode with
+// another directory entry.
+var errSharedInodeManagedTarget = errors.New("refusing to write through a hard link")
 
 // writeNewContainedFile is the create-only counterpart to writeContainedFile. The returned
 // FileInfo identifies the entry this call acquired even if writing or closing it subsequently
--- a/internal/cli/agents.go
+++ b/internal/cli/agents.go
@@ -710,11 +710,88 @@ func resolveContainedNameWithOptions(root *os.Root, name string, allowMissingDir
 			errGitDirManagedTarget, name, filepath.ToSlash(resolved),
 		)
 	}
+	if administrative, inside := landsInGitAdministrativeDirectory(root, resolved); inside {
+		return "", fmt.Errorf(
+			"%w: %s resolves to %s, inside %s, which is this repository's git directory whatever it is named",
+			errGitDirManagedTarget, name, filepath.ToSlash(resolved), filepath.ToSlash(administrative),
+		)
+	}
 	return resolved, nil
 }
 
 // errGitDirManagedTarget marks a managed target whose resolution lands in the git directory.
 var errGitDirManagedTarget = errors.New("refusing to write into the git directory")
+
+// landsInGitAdministrativeDirectory reports the git administrative directory a resolved landing is
+// inside, whatever that directory is NAMED, and whether it found one.
+//
+// sem.PathLandsInGitDir above answers a question about a NAME: it refuses a landing that spells a
+// `.git` component. That is the only shape it can see, and git does not require that shape. A
+// repository created with `git init --separate-git-dir=admin` keeps a `.git` POINTER FILE holding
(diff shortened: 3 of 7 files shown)


suhaanthayyil:
Adversarial re-review. Gate placement is right — every sink funnels through `resolveContainedNameWithOptions` (`statContainedFile:1247`, `readContainedFile:1251`, `writeContainedFile:1268`, `writeNewContainedFile:1288`, `mkdirAllContained:553`) and I found no unguarded sibling writer in `agents.go`. But `PathLandsInGitDir` is a check on the resolved *name*, and that leaves four ways in.

**1. Hardlink — no race needed.** `PathLandsInGitDir(rel)` takes a repo-relative path string, so it can only see `.git` when a `.git` component survives resolution. `ln .git/config CLAUDE.md` resolves to `"CLAUDE.md"`, zero `.git` components, `Lstat` reports a regular file. `readAndValidateInstructionFile` then reads git's config and `writeContainedFile` `O_TRUNC`s the shared inode and writes it back with the managed block appended. That is exactly the corruption this PR is written to prevent. `docs/trust-and-security.md` now states unconditionally that a hostile checkout cannot aim a managed block at `.git/config` or a hook; the hardlink paragraph next to it carves out no exception for `.git`. Either weaken the doc sentence or check `nlink > 1` (or compare the opened file's inode against a `.git` walk) before writing.

**2. TOCTOU inside a single operation.** `resolveContainedName` Lstat-walks and returns a *name*; the sink then calls `root.OpenFile(resolved, O_CREATE|O_TRUNC)`, which re-resolves. `os.Root` refuses a link that *escapes* the root atomically at the kernel, but it *follows* one that stays inside — and `.git` is inside. So a symlink planted between the walk and the open is followed. The comment that "each one re-asks, and a link swapped between them is refused by the operation that finds it" is true across operations and false within one. Open first, then fstat the handle and judge that.

**3. Name heuristic, not the resolved gitdir.** `GIT_DIR`, `--git-dir`, a worktree gitlink, or any checkout whose admin directory is not literally named `.git` is not refused. This inherits the weakness of the read-side `hasGitDirComponent` it cites as precedent rather than fixing it.

**4. `init-agents --repo <path>/.git`.** The test is on the repo-relative resolved path, so pointing `--repo` at the git directory itself installs instruction files straight into it.

Separately: this is a denylist of one entry. `CLAUDE.md -> .github/workflows/ci.yml`, `-> Makefile`, `-> .envrc` all still receive the managed block, and each is as good a persistence primitive as a git hook.

Cross-PR: this PR resolves the full symlink chain and then judges the landing by name, while #160 never resolves and refuses any symlinked component outright. Applied to each other's domain they disagree — #160's rule would reject the `AGENTS.md -> docs/guide.md` alias this PR documents as legitimate, and this PR's rule would permit `<cache>/search -> <cache>/elsewhere`. `internal/cli/outputpath.go:892` carries a third idiom. Three path-confinement models and no shared helper is how the next one of these gets missed.

ashtom:

**Verdict: clean, with a merge-order note**

- Live PR head `6f8926e2`; current `main` `bb41f223`; **2 ahead, 13 behind**.
- **Every green check on this PR ran against `6f8926e2` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking in the change itself.

**1. (merge order) This PR and #164 both edit `internal/cli/agents.go` and conflict with each
other**, although each merges cleanly onto `main` on its own:

```
$ git merge-tree --write-tree refs/prs/158 refs/prs/164   ->  CONFLICT
```

They are independent improvements (this one refuses a `.git` landing; #164 bounds the read size), so
both should land — but the second one merged will need a manual resolution. Worth sequencing
deliberately rather than discovering it at merge time.

**2. (deliberate trade, flagging for the record)** `PathLandsInGitDir` folds case with
`strings.EqualFold`, so on a case-sensitive filesystem a directory genuinely named `.GIT` is also
refused. The PR argues this explicitly and fails closed; I agree with the trade, and note only that
it is a real (if tiny) over-rejection on Linux.

### Independent review state

suhaanthayyil:
Thanks. Taking the merge-order finding, with one correction to it, and agreeing on the case-fold trade.

**Finding 1 (merge order with #164) — real, but narrower than stated.** You wrote that the two "both edit `internal/cli/agents.go` and conflict with each other". They do both edit `agents.go`, but `agents.go` merges cleanly; the conflict is in `agents_test.go` only:

```
$ git merge-tree --write-tree 6f8926e2 c22b8460
Auto-merging internal/cli/agents.go
Auto-merging internal/cli/agents_test.go
CONFLICT (content): Merge conflict in internal/cli/agents_test.go
```

One conflict, one file. It is an append/append: #164 adds `TestInitAgentsRefusesOversizeInstructionFileWithoutWrites` at the end of the file, this branch adds `refusedEvenIfTheKernelCanReachIt` to `aliasPlant` and extends `untraversableAliasPlants` just above it. Resolution is "keep both", with no interaction between the two changes. Agreed they should both land; the sequencing is worth being deliberate about, but the cost of getting it wrong is smaller than the note implies.

**Finding 2 (`.GIT` over-rejection on a case-sensitive filesystem) — agreed, and deliberate.** Failing closed on a directory literally named `.GIT` is the trade I intended; thank you for putting it on the record rather than treating it as a bug.

**What this PR still does not close**, from my own earlier pass, so it is not mistaken for settled by a clean verdict:

1. `PathLandsInGitDir` judges the resolved NAME, so `ln .git/config CLAUDE.md` resolves to `"CLAUDE.md"` with zero `.git` components and `Lstat` reporting a regular file. The managed block then lands on git's config through the shared inode. No race required.
2. TOCTOU within one operation: `resolveContainedName` Lstat-walks and returns a name, and the sink re-resolves on `OpenFile`. `os.Root` refuses a link that ESCAPES the root, atomically, but follows one that stays inside — and `.git` is inside.
3. The check is a name heuristic, not the resolved gitdir, so `GIT_DIR`, `--git-dir`, and a worktree gitlink are not covered.
4. It is a denylist of one entry: `CLAUDE.md -> .github/workflows/ci.yml`, `-> Makefile`, `-> .envrc` all still receive the managed block.

Each is a separate change and I would rather not smuggle them into a PR you have already verified. Say which of them you want as follow-ups and which should block this.

suhaanthayyil:
Closing the hardening items I listed last round. **Two of the four are fixed in `b8c5dbb4`; two are genuinely out of scope for this branch and I say why rather than leaving them implied.**

## 1. Hard link — was live, exploitable, and now refused

Not theoretical. Measured on this branch before the change:

```go
os.Link(repo/".git/config", repo/"CLAUDE.md")   // a HARD link, not a symlink
Run(..., "init-agents", "--repo", repo)
// run err = <nil>          ← exited 0
// .git/config REWRITTEN with the managed block appended
```

`PathLandsInGitDir` is structurally unable to see this. Every guard on this branch reasons about a path — its components, what it resolves to, whether a component is a symlink — and a hard link has no path to reason about. `CLAUDE.md` *is* git's config: it resolves to `"CLAUDE.md"`, carries no `.git` component, and `Lstat` reports an ordinary regular file. The name is innocent; the inode is not.

**The obvious fix — refuse any hard link — would have broken a documented feature.** `docs/agents.md` offers hard-linking `AGENTS.md` and `CLAUDE.md` as a supported way to share one instruction file, and `TestInitAgentsWritesSameFileOnlyOnce/CLAUDE_hard-links_to_AGENTS` covers it. My first attempt did exactly that and broke it; the test was right and I changed the rule, not the test.

The rule that is actually correct: **is every one of this inode's names ours?** Counting answers it without needing to enumerate an inode's names — when as many managed targets share the inode as the inode has names, all names are accounted for and all are managed. One unaccounted name means some other path reaches the same bytes. The attack has `Nlink=2` with 1 managed name → refused. The documented alias has `Nlink=2` with 2 managed names → allowed.

## 2. TOCTOU inside one operation — fixed, same root cause

You resolve a NAME and then `OpenFile` re-resolves it, so a link swapped between the two is followed — `os.Root` refuses a link that *escapes* the root atomically, but follows one that stays inside, and `.git` is inside. The handle is now what gets judged, because the handle is what the write lands on.

`O_TRUNC` also came out of that open. Opening `.git/config` with `O_TRUNC` destroys it *before* any guard can object; truncation now happens only after the handle is accepted. Mutation MUT3 below is exactly this.

## Mutation verify (all compiling)

| mutation | attack test | documented-alias test |
|---|---|---|
| drop the shared-inode guard | **FAIL** | pass |
| blanket refusal of any hard link | pass | **FAIL** |
| restore `O_TRUNC` in the open | **FAIL** | pass |
| unmutated | pass | pass |

Row two is the one that matters: it proves the guard is *precise* and not a blanket that would quietly disable a supported feature. `go test ./internal/cli/` — pass, 22.1s.

## 3 and 4 — NOT fixed, and deliberately so

**Name heuristic vs. resolved gitdir** (`GIT_DIR`, `--git-dir`, a worktree gitlink). Real, and unfixed. `sem.PathLandsInGitDir(rel)` takes only a repo-relative string; making it consult the *resolved* git directory means threading repository state into a pure path predicate that `internal/sem` also uses on the read side. That is a change to a shared boundary, not a patch to this one, and it deserves its own PR rather than riding along here.

**Denylist of one entry** (`CLAUDE.md -> .github/workflows/ci.yml`, `-> Makefile`, `-> .envrc`). Also real. But this PR's stated scope is the git directory, and each of those needs a policy decision about which paths a repository may aim a managed block at — that is a different question from "never write inside `.git`", and answering it by quietly widening this predicate would make the change harder to review, not safer.

Both remain open. Happy to take either as a follow-up if you would rather they not wait.

suhaanthayyil:
## Both "out of scope" hardenings closed

Two commits on this branch: `b512ce2a` (git directory by structure, not by name) and `21d73dba` (write only into an agent-instruction file). Both hypotheses reproduced first; neither is refuted.

---

### 1. The name heuristic misses the git directory git actually uses — **VALID**

`sem.PathLandsInGitDir` (`internal/sem/provider.go:10315`) takes a repo-relative STRING, so it can only refuse a landing that SPELLS `.git`. Git does not require that spelling.

**Reproduced before the fix** (real `git init`, binary built from `b8c5dbb4`):

```
$ git init --separate-git-dir="$R/admin" "$R"
$ cat "$R/.git"                      # a gitlink FILE, not a directory
gitdir: .../repro-a1/admin
$ git -C "$R" rev-parse --absolute-git-dir
.../repro-a1/admin                   # the REAL admin dir, no `.git` component

$ ln -s admin/config "$R/CLAUDE.md"
$ entire-graph init-agents --repo "$R"
wrote .../.entire/graph-agent.md
updated .../AGENTS.md
updated .../CLAUDE.md                 # exit 0
$ tail -6 "$R/admin/config"
<!-- entire-graph:begin -->
This repo has the entire-graph code graph installed...
<!-- entire-graph:end -->
$ git -C "$R" status
fatal: bad config line 9 in file .../repro-a1/admin/config
```

Two more, same shape:

| planted alias | result before the fix |
|---|---|
| `AGENTS.md -> admin/hooks/pre-commit` (repo made with `GIT_DIR=admin git init`, **no `.git` entry at all**) | exit 0, managed block appended to an executable hook |
| `.entire -> admin/refs/heads` | exit 0, guide written into the real ref store; `git branch -a` → `warning: ignoring broken ref refs/heads/graph-agent.md` |

That last one is the case **no filename rule can catch** — the landing is `graph-agent.md`, an ordinary markdown name. Only *where* it lands is wrong. It is why finding 2 below does not subsume this one.

**Fix — shape (2), the separate check at the `internal/cli` write sink.** `internal/cli/agents.go:686`, calling `landsInGitAdministrativeDirectory` (`:743`) / `isGitAdministrativeDirectory` (`:760`).

Shape (1) — threading real repo state into `PathLandsInGitDir` — was rejected because that predicate is the *shared* boundary: `internal/sem` calls it as `hasGitDirComponent` (`internal/sem/provider.go:10287`, used at `:13303`) to decide what the indexer walks. Widening it would change what gets indexed, in a package whose own comments record that this rule has already had to be narrowed back from over-suppression twice. The write sink is where the extra knowledge belongs, and the shared predicate is byte-identical to what was on this branch before (see the diff check below).

Design points worth reviewing:

- **Structure, not `git rev-parse`.** The check applies git's own `is_git_directory()` — a `HEAD`, plus `objects/` and `refs/`, with a `commondir`/`gitdir` alternative for a linked worktree's admin directory. Shelling out would make the answer depend on a binary being installed, on `$GIT_DIR` in this process, and on the CWD — none of which describe the repository named by `--repo`.
- **Re-asked, not cached.** It sits at the same chokepoint as the existing name rule (`resolveContainedNameWithOptions`, `:675`), so every stat, read and write re-asks it. A value resolved once at startup could not refuse a directory that becomes a git directory between two operations — which is the property this file already holds itself to for the preflight.
- **The `.git` pointer text is deliberately NOT read.** `sem`'s `gitDirExcluder` already records why: the pointer is attacker-controlled input, so git resolves it and then asks `is_git_directory()` of the target anyway. With that structure test applied, the pointer adds nothing — a landing inside the target has the target among its own ancestors, and a target outside the repo is refused by `os.Root` first. Reading the text *without* the structure test would only hand a repository a way to make an ordinary directory of its own unwritable.

Worth noting for the record: **the read side of this boundary was already stricter than the write side.** `gitDirExcluder` resolves separate-git-dir pointers and applies the structure test. This branch's own comment argues "a writer that will not read there must not write there either" — that was not true until now.

---

### 2. The denylist is one entry, and cannot be finished — **VALID**

**Reproduced before the fix**, all exit 0, all victims rewritten:

| alias | victim after the run |
|---|---|
| `CLAUDE.md -> Makefile` | `all:\n\techo hi` + managed block |
| `CLAUDE.md -> .github/workflows/ci.yml` | workflow + managed block |
| `CLAUDE.md -> .envrc` | `export FOO=bar` + managed block |

One precision worth stating plainly rather than overclaiming: the injected text is *this tool's own fixed markdown*, so the primitive is **silent corruption of a file that was already doing a job**, not attacker-controlled execution. That is the same harm class this branch already accepted as sufficient to refuse `.git/config`.

**Policy chosen: an allowlist on the landing, not more denylist entries.**

> A managed target may resolve onto an **existing** file only if that file is itself an agent-instruction file — markdown by extension, or a rules file that carries no markdown extension (`.cursorrules` and siblings).

`refuseNonInstructionLanding` (`internal/cli/agents.go:840`), `isInstructionFileName` (`:806`), hooked at `:695` for file resolutions only (a directory resolution names a directory, which is not this question).

Why not a denylist: it cannot be finished, and it fails the wrong way round. Every build system, task runner and shell init file yet to be invented is one more entry someone must remember; a destination nobody thought of is admitted **silently**. The allowlist inverts that — a name missing from it costs a refusal the user resolves by renaming the target, never a corrupted file. And it is bounded by what the feature is *for*: `docs/agents.md` offers the alias so `AGENTS.md` and `CLAUDE.md` can share ONE INSTRUCTION FILE, so a landing that is not an instruction file is not a use of the feature and refusing it costs the documented case nothing.

Three landings are deliberately still accepted, each because there is nothing there to protect:

1. **One that does not exist.** `init-agents` creates all of its own targets, and the documented dangling alias (`CLAUDE.md -> AGENTS.md` before `AGENTS.md` exists) depends on it. Nothing is destroyed: the new file holds this command's markdown at 0644 and appears untracked in `git status`. This also keeps `TestInitAgentsAllowsCaseDistinctMissingManagedTargets` — which writes through `AGENTS.md -> NESTED`, no extension — green without weakening it.
2. **One that already carries this command's own output** (its managed block, or the guide's own heading). A first run that succeeded and a second that failed would be worse than no rule. Covered by `TestInitAgentsAllowsInstructionFileLandings/landing this command created is writable again` and `.../guide landing ...`.
3. **One that is not a regular file or cannot be stat'ed** — not this refusal's question; the operation that follows reports it as it is.

**One defect found while implementing this.** The refusal is raised by the resolver, not by `os.Root`, so it carries no syscall errno and `isRootEscape` cannot distinguish it from `os.Root`'s own escape sentinel. Without an explicit case, `CLAUDE.md -> Makefile` was reported as *"refusing to write through a link that leaves the repository"* — sending the reader to hunt a link that never left the root. Named explicitly at `internal/cli/agents.go:320`, next to the existing `errGitDirManagedTarget` case which is there for the same reason, and asserted in the test.

---

### Tests

| test | file |
|---|---|
| `TestInitAgentsRefusesGitDirectoryNotNamedDotGit` (4 subtests: in-tree gitlink target; no gitlink at all; guide-directory alias into the real ref store; admin dir with `commondir` in place of its own stores) | `internal/cli/agents_test.go:2663` |
| `TestInitAgentsRefusesLandingThatIsNotAnInstructionFile` (6: ci.yml, Makefile, .envrc, package.json, .bashrc, guide alias) | `:2767` |
| `TestInitAgentsAllowsInstructionFileLandings` (4: existing markdown; `.cursorrules`; re-run onto a created landing; re-run onto a created guide landing) | `:2833` |
| `TestAgentGuideHeadingIdentifiesTheGuide` | `:2903` |

Fixtures are built by hand (`plantGitAdministrativeDirectory`, `:2638`) rather than by shelling out to `git init --separate-git-dir`, so each test states exactly which structure the refusal keys on and needs no git binary.

### Mutation verification

Each mutation reverts only production code, keeps the tests, and compiles. `grep` output proving the line actually changed is included.

| mutation | grep proof | result |
|---|---|---|
| disable the git-admin check | `686: ...landsInGitAdministrativeDirectory(root, resolved); inside && false {` | `TestInitAgentsRefusesGitDirectoryNotNamedDotGit` **0 passed, 5 failed** |
| disable the instruction-file check | `695: if false && !allowMissingDirectory {` | `TestInitAgentsRefusesLandingThatIsNotAnInstructionFile` **0 passed, 7 failed**, each reporting the victim actually rewritten |
| disable the error classification case | `320: case false && errors.Is(err, errNonInstructionManagedTarget):` | `TestInitAgentsRefusesLandingThatIsNotAnInstructionFile` **0 passed, 7 failed** — "a link that never left the root was reported as a repository escape" |
| both checks disabled together | both lines above | the git-directory test reports **"init-agents wrote into the git directory"** on all four subtests — i.e. with only the instruction-file rule removed, three of the four are still caught by it, and the fourth (`.entire -> admin/refs/heads`, a `.md` landing) is caught by nothing. That is the case that makes finding 1 independently load-bearing. |

Restoration was grep-verified each time (`grep -c "inside && false"` → `0`).

### Shared-boundary check

```
$ git diff --stat b8c5dbb4 -- internal/sem/
(empty)
$ go test ./internal/sem/ -run 'TestSkipVendoredDirAlwaysExcludesGitDir|TestSearchRepositoryNeverIndexesGitDirOnFilesystemFallback|
    TestGitDirPointerTargetResolvesSeparateGitDirPointer|TestGitDirExcluderCoversEveryGitDirShape|
    TestSearchRepositoryNeverIndexesSeparateGitDir|TestSearchRepositoryNeverIndexesNestedGitDirs' -count=1
31 passed
```

`internal/sem` is byte-identical to this branch's previous head, so what the indexer walks cannot have changed; the read-side git-directory tests confirm it.

### Suite

`go test ./internal/cli/ -count=1` → **784 passed** (766 before these commits). `gofmt` clean, `go vet` clean, `go build ./...` clean. No existing test was modified, skipped or weakened; the only test-file edit outside the new tests is an added assertion inside the new one.

### Left open

- The three `.cursorrules`-class names in the allowlist are a judgement call about which non-markdown rules files are real. The list only ever ADMITS, so an omission costs a refusal rather than a corruption, but it will need entries as new agents appear.
- A landing that does not yet exist is still created anywhere in the repository (see acceptance 1). Creating a `Makefile` or `.envrc` where none existed is markdown this tool wrote, non-executable, and untracked — visible, not silent — but it is a deliberate limit of the rule, not an oversight.
- `.git` handling on Windows is unchanged and untested here; the new checks use `os.Root` stats only, so they inherit whatever that layer already does.

ashtom:

**Verdict: changes required — 3 HIGH findings, all new since my last review**

Head moved `6f8926e2` → `21d73dba`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### Note

My previous review called this one clean. That was accurate for `6f8926e2`; it is not accurate for
the current head. The three new commits add hard-link protection, and that new surface is where the
HIGH findings land — including one that says the protection is **disabled entirely on Windows**
(`agents_hardlink_windows.go:11`) even though NTFS supports hard links. A guard that no-ops on the
platform it is protecting against is the §7 case in Thomas's protocol almost exactly.

The gate is green, which is the point worth making: green CI says nothing about any of these six.

### Entire trail findings — 6 open and current

```
01M17K90AKKX  Medium  open    current    internal/cli/agents.go:869                  A substring is not proof that this command created the landing: an existing non-…
01M17K90A9ZX  Medium  open    current    internal/cli/agents.go:770                  A linked-worktree administrative directory has both `commondir` and `gitdir`, bu…
01M17K909TXV  High    open    current    internal/cli/agents.go:1533                 The opened handle is checked only for unaccounted hard links, so swapping the re…
01M17HAWH0YP  Medium  open    current    internal/cli/agents.go:1415                 This rejects every managed instruction file with any non-managed hard-link name,…
01M17HAWGNFK  High    open    current    internal/cli/agents_hardlink_windows.go:11  The hard-link protection is completely disabled on Windows even though NTFS supp…
01M17HAWG6PC  High    open    current    internal/cli/agents.go:1412                 Counting managed pathnames overcounts actual hard-link names because symlink ali…
```

### Verified vs. assumed

suhaanthayyil:
Six findings against the hard-link and gitdir work I added since your last review. **Two HIGH are valid and fixed (`561f174a`); one HIGH is valid and I could NOT fix it — I tried, the fix was inert, and I reverted it rather than ship a guard that does nothing.** Detail per finding below.

## HIGH — `agents_hardlink_windows.go:11`, protection disabled on Windows: **VALID, fixed**

You are right, and the reasoning I wrote there was wrong rather than merely cautious. NTFS supports hard links, so a guard that no-ops on Windows is a guard in name only. My excuse was that Go's Windows `FileInfo` carries no link count — true, but `GetFileInformationByHandle` answers it from the handle: `NumberOfLinks`, plus `(VolumeSerialNumber, FileIndexHigh, FileIndexLow)` as the volume-scoped equivalent of `(device, inode)`.

The helpers now take the open `*os.File` rather than a `FileInfo`, because identity is a handle question on Windows, and the candidate loop opens each entry instead of stat'ing it. Verified with a standalone `GOOS=windows` build of the new file — the repo's own cross-vet cannot run at all (CGO grammar packages are excluded under cross-compilation), and that is pre-existing on `main`, not something this PR introduced.

## HIGH — `agents.go:1412`, symlink aliases overcount: **VALID, fixed, and it was a live bypass**

This one was exploitable, not theoretical. The guard counts managed names that resolve to the inode, but only real directory entries contribute to an inode's **link count**, so a symlink alias inflates the tally without inflating the count. Measured on the previous head:

```
CLAUDE.md  hard-linked to .git/config   -> 2 names
AGENTS.md  symlinked to CLAUDE.md       -> counted as a 2nd "managed name"
accounted(2) >= links(2)  =>  allowed

run err = <nil>          ← init-agents exited 0
.git/config REWRITTEN with the managed block
```

A candidate is now counted only if `Lstat` says it is a real directory entry. Sharing one instruction file through a symlink — the documented alias — stays supported; it simply is not evidence about who else holds a hard link.

## HIGH — `agents.go:1533`, the handle is checked only for link count: **VALID, NOT FIXED**

You are right and my own comment overstated it: I wrote that judging the handle "removes the window", but the handle is judged only for unaccounted hard links, never for **where it landed**. A symlink swapped in at the final component after resolution is followed (`os.Root` permits it — it stays in-root, and `.git` is in-root) and passes the link-count check, because a freshly planted link's target has exactly one name.

I tried to close it with `O_NOFOLLOW` on the write open. **It does not work, and I am reporting that rather than shipping it.** `os.Root` strips the flag:

```
os.Root.OpenFile   with O_NOFOLLOW: err=<nil>                              ← opened the symlink
plain os.OpenFile  with O_NOFOLLOW: err=too many levels of symbolic links  ← correct
```

So the change was inert, and a test I wrote for it passed for an unrelated reason (the instruction-file guard refuses the alias first) — which is exactly the "guard that no-ops" shape you flagged on Windows. I reverted it. **The finding stays open.** Closing it properly needs either an `openat2`-style resolve flag Go does not expose through `os.Root`, or a post-open identity re-check against the pre-validated landing; both are more than a patch and I would rather leave it named than pretend.

## Medium — `agents.go:1415`, refuses an instruction file with any non-managed hard-link name

Accurate, and it is a deliberate trade rather than an oversight: an instruction file that happens to have an unrelated second name (a `cp -l` backup, say) is refused. The alternative is to allow a write when at least one name is unaccounted for, which is precisely the exploit above. It fails closed, on the same reasoning as the case-folded `.GIT` refusal you accepted earlier. Happy to revisit if you would rather it warn and proceed, but I would not make that change without you saying so.

## Medium — `agents.go:770`, linked worktree has both `commondir` and `gitdir` / Medium — `agents.go:869`, a substring is not proof

Both acknowledged and **not yet addressed** — they landed with the gitdir work in the same batch and I ran out of runway this round after the three HIGHs. Not disputed, not fixed; they stay open and are next.

## Verification

`go test ./internal/cli/` — pass, 24.4s, 162 tests in the init-agents suite. Mutations, all compiling and confirmed applied by `grep` after the edit:

| mutation | result |
|---|---|
| count symlink aliases again | **FAIL** (bypass test); documented-alias and plain-hard-link tests still pass — the guard is precise, not a blanket |
| identity reports "unknown" (the old Windows no-op) | **FAIL** on *both* the bypass and plain hard-link tests — which is the measured argument for why the Windows no-op was HIGH |
| unmutated | pass |

PR #159, opened 2026-08-28
fix(sem): report a pure file rename in the semantic diff

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/118

## Problem

`entire graph diff` drops a file whose content is byte-identical across a rename, so a pure rename produces an empty result:

```
$ git mv old/sample.go new/sample.go && git commit -m rename
$ entire graph diff --base HEAD~1 --head HEAD
Semantic changes HEAD~1..HEAD

No semantic entity changes detected.

$ entire graph diff --base HEAD~1 --head HEAD --json
{ "base": "HEAD~1", "head": "HEAD", "files": null, ... }
```

`git diff --name-status` reports `R100 old/sample.go new/sample.go` for the same range, and every symbol in the file was re-identified: the file path is a component of the `compound-v1` symbol ID.

```
before: local/repo:Go:old/sample.go:function:Run
after:  local/repo:Go:new/sample.go:function:Run
```

The output is therefore indistinguishable from "nothing changed" for any consumer that keys on it — the entities under the old path are never retired and the ones under the new path are never learned.

The drop happens because `moduleScopeChange` correctly declines to invent a content change for identical bytes, and its caller then `continue`s past the whole file.

## Fix

A path change is itself a semantic change, so report it. When a changed file produces no entity change and no module-scope content change, but its path differs from its old path, emit a module-scope `moved` change.

```
$ entire graph diff --base HEAD~1 --head HEAD
old/sample.go -> new/sample.go (Go)
  ~ module new/sample.go moved from old/sample.go (0 dependents)
```

This reuses the existing `moved` change type, which already carries `old_path` / `new_path` / `reconciliation: "MOVED"` for entities reconciled across files, and the existing module-scope synthetic entity. No new change type, no new field, no schema change — `1.x` stays additive-only.

Files that were already reported are unaffected: the new path only runs where the file used to be dropped entirely.

## Tests

- `TestAnalyzeGitRangePureRenameIsReported` — a 100%-similarity rename yields one file with `status: R`, both paths, and the module `moved` change. Fails before the fix with `pure rename produced 0 files, want 1`.
- `TestAnalyzeGitRangeUnchangedFileStaysAbsent` — a file nothing touched still produces no delta.
- `TestAnalyzeGitRangeModeOnlyChangeStaysAbsent` — a chmod-only change (identical content, unchanged path, still reported by `git diff`) is the one case that reaches the new branch without a rename; it must not be reported as a move.

Mutation-checked: disabling the new branch fails the first test; relaxing its `oldPath == path` guard fails the third.

---

**Local gate**

- `go test -timeout 30m ./...` — green, every package `ok` (internal/cli 498s, internal/sem 816s).


ashtom:
bugbot run

suhaanthayyil:
Bugbot could not run — it reported `Bugbot is not enabled for your user on this team` and pointed at the seat limit / allowlist in the Cursor dashboard. That is an account setting rather than anything about this branch, so re-triggering it will keep returning the same message until someone with team-admin access adds a seat.

Flagging it because the same message will come back on every PR where the protocol asks for a Bugbot run, so "Bugbot: did not run" cannot currently be read as a signal about the change either way.

ashtom:
bugbot run

PR #160, opened 2026-08-28
fix(sem): confine cache artifact writes to the cache directory

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/119

## Problem

Search-snapshot and provider-record cache writes previously used `MkdirAll`,
`CreateTemp`, and `Rename` on joined paths. A repository-controlled symlink at
the cache family or version component could therefore steer derivative cache
bytes outside the cache directory or, for an in-root link, into `.git`.
Windows junctions and mount-point reparse entries had the same redirecting
property. The artifact leaf itself was already installed by rename and was not
the vulnerable component.

## Policy

The caller-selected cache root is the trust boundary and may itself be a
symlink. Beneath it, Entire Graph owns the family and version directories and
strictly refuses every redirecting descendant, including an alias whose target
stays inside the root. This intentionally drops the older in-root-alias layout.
Operators can relocate a cache by naming the backing directory as the root or
making the root itself a symlink.

The boundary covers repository-controlled path entries and substitutions
observable while a component is opened. It does not claim resistance to a
concurrently running process with permission to rename an already-opened
directory. Go's `os.Root` intentionally continues using that directory object
after a move, portable Go cannot pin its lexical ancestry, and a process with
that namespace authority can already move existing cache artifacts. This scope
is now explicit in the code, operations guide, and trust documentation.

## Implementation

- Both cache families now use one `cacheEntry.write` path.
- Each family and version component is opened first, checked for symlink or
  platform-specific redirect modes, and compared by filesystem identity with
  the name exposed by its parent.
- Temporary creation and final installation use the verified directory handle;
  the final rename replaces rather than follows a planted artifact symlink.
- Temporary names use fixed-width words from the infallible process-local
  `math/rand/v2` generator. `O_EXCL` is the no-follow/no-overwrite boundary: an
  occupied candidate is retried, while prediction can only make the bounded
  cache write fail. This avoids the irrecoverable OS-entropy failure path shared
  by the default `crypto/rand` APIs.
- Cache artifacts retain explicit mode `0600` where the platform exposes Unix
  permission bits.

## Findings addressed

- Resolved: unrecoverable production randomness failure; deterministic tests
  cover one occupied candidate and exhaustion of all 16 retries without
  overwriting the occupied entry.
- Resolved: open-time component identity race.
- Resolved: Windows junction/reparse redirects.
- Dismissed by explicit policy decision: relocation of an already-opened
  directory by a concurrently executing process with namespace-write authority.
- A post-fix review found that the first identity-race regression test could
  false-pass when its setup failed on Windows. It now injects a deterministic
  mismatched directory handle and requires the exact production error.

## Verification

- Merged with current `main` (`ee6468a6`) without conflict.
- `go test ./...` passed on the behavior-changing head.
- `go vet ./...` and `go build ./...` passed.
- Focused cache-confinement and deterministic identity-mismatch tests passed.
- Three independent post-fix reviews found no remaining implementation issue.
- Final-head CI passed build, lint, and tests on Ubuntu, macOS, and Windows.
  Entire Gates is waiting only for the deliberately withheld trail approval.


suhaanthayyil:
Adversarial re-review. The `writePath` migration is complete and the write-side component walk does what it says. But the central regression argument is wrong for in-root links, and the read side was not brought along.

**1. BREAKS a layout that works today.** The PR justifies breaking symlinked family/version components with: "reads go through `os.OpenRoot(cacheDir)`, which refuses exactly that — so the artifact was written on every run and read on none." That holds only for links that *leave* the cache directory. `os.Root` **follows** an in-root link. `cacheEntry.open()` on this branch is unchanged and is still `root.Open(entry.relative)` — so `<cache>/search -> <cache>/bigstore` (a shared or relocated family directory that stays under the root) read *and* wrote fine on `main`, and after this PR `entire graph index` hard-fails on it. `search_cache_test.go`'s hunk converting `preindexProviderSnapshotWithPersistenceReader` from success to a hard error is the same shape of change. No test covers the in-root layout, and the "costs nothing that works today" claim in the doc comment is not true for it.

**2. Read and write now disagree.** The PR's own thesis is that "the write boundary must be the one `entry.open` already reads through". After this change, writes refuse `search -> .git` while reads still follow it. Either symmetrise `open()` with the same component walk, or drop the claim.

**3. TOCTOU in `openCacheComponent`.** The sequence is `Mkdir` → `Lstat(name)` → `parent.OpenRoot(name)`. A symlink swapped in between the `Lstat` and the `OpenRoot` is followed as long as it stays in-root. `openat` the component first, then `fstat` the returned handle, is the race-free form.

**4. Windows reparse points are missed.** The guard is `info.Mode()&fs.ModeSymlink != 0` only. `internal/cli/agents.go:669` in this same repo already establishes that Go 1.23+ reports Windows reparse points as `ModeIrregular` rather than `ModeSymlink` and checks both. A junction planted at the family or version component is not refused here.

**5. Comment and test both describe logic that does not exist.** "The check is not exact-case, so a case-insensitive filesystem cannot answer it with a differently-spelled link" — there is no case comparison anywhere in the added code; the check is `Lstat` + mode bits, which is case-agnostic for a different reason. And `TestCacheEntryWriteRejectsSymlinkedVersion`'s comment claims to cover a differently-spelled link while planting one at the exact name `v1`. Please either implement or drop the claim; as written it reads as tested behaviour that is not tested.

**6. HARD BUILD BREAK with #154.** This PR deletes `func (entry cacheEntry) writePath()`. #154 adds `internal/sem/cache_schema_version_test.go:191`, which calls `entry.writePath()`. Merge both in either order and package `sem` does not compile. You rewrote the other caller (`search_cache_test.go:132`); #154's new one needs the same treatment.

ashtom:

**Verdict: clean**

- Live PR head `90fc71c9`; current `main` `bb41f223`; **2 ahead, 13 behind**.
- **Every green check on this PR ran against `90fc71c9` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking.

**Scope check (§6).** This does change shared behaviour — both the records cache and the search
cache now write through one `os.Root`-confined `cacheEntry.write` instead of two near-identical
open-coded paths. That *is* the PR's stated scope, the two prior implementations were byte-for-byte
equivalent in structure, and consolidating them is what makes the confinement enforceable in one
place rather than two. I checked that the temp-file/`Chmod`/rename ordering is preserved on both
paths.

### Independent review state

suhaanthayyil:
Thanks. Nothing to change here from your review — the scope check is the right question and your read of it matches mine, including the temp-file/`Chmod`/rename ordering being preserved on both paths.

One cross-PR item, since this branch is one half of it: **this PR and #154 did not compile together.** This PR deletes `func (entry cacheEntry) writePath()` and rewrites its callers; #154 had added a new caller of it at `internal/sem/cache_schema_version_test.go:191`. Either merge order broke package `sem`:

```
vet: internal/sem/cache_schema_version_test.go:195:29: entry.writePath undefined (type cacheEntry has no field or method writePath)
```

I fixed it on **#154** (`f2e15c35`) rather than here, by moving that caller to the `filepath.Join(entry.root, entry.relative)` idiom this PR already uses — so **no change is needed on this branch**, and the two are now merge-order independent. Verified on the trial merge of the two branches: `go vet` clean, 13 sem tests pass; reverting the #154 fix on that merged tree reproduces the build failure above.

ashtom:

**Verdict: unchanged since my last review, but 3 finding(s) have appeared since**

no longer does.

### Entire trail findings — 3 open and current

```
01M13AY58XCV  Medium  open    current    internal/sem/cache_entry.go:194  `rand.Text()` irrecoverably terminates the process if the OS randomness source f…
01M1398TQ21X  Medium  open    current    internal/sem/cache_entry.go:186  There is a TOCTOU gap between `Lstat` and `OpenRoot`: after the component passes…
01M1398TPVAY  Medium  open    current    internal/sem/cache_entry.go:183  On Windows, junctions and mount points are reported as `ModeIrregular`, not `Mod…
```

### Verified vs. assumed

suhaanthayyil:
## Finding responses — three open findings, two valid, one refuted

Head is now `30d490af`.

---

### 1. `internal/sem/cache_entry.go:183` — Windows junctions are `ModeIrregular`, not `ModeSymlink` — **VALID**

The guard tested `info.Mode()&fs.ModeSymlink != 0`. That is the right test on Unix and the wrong one on Windows: a symlink is reported as `ModeSymlink`, but a directory junction or mount point is reported as `ModeIrregular`. A junction planted at the `family` or `version` component therefore passed the check and was then descended into and followed — exactly the redirect this guard exists to refuse.

This repository already had the correct predicate, with five production call sites:

```
internal/sem/path_redirect_windows.go:12   return info.Mode()&(os.ModeSymlink|os.ModeIrregular) != 0
internal/sem/path_redirect_other.go:8      return info.Mode()&os.ModeSymlink != 0
```
callers: `provider.go:11777`, `provider.go:13200`, `provider.go:15057`, `sweep_open_other.go:68`, `metadata_tree_entry_windows.go:37` — all in `package sem`, so it is a drop-in here.

**Fix:** `openCacheComponent` now calls `cacheComponentRedirectError`, which tests `pathMayRedirect(info)`. The message keeps the substring `is a symlink` that `search_cache_test.go:161` matches on, and names the mode it found.

---

### 2. `internal/sem/cache_entry.go:194` — `rand.Text()` terminates the process — **VALID** (and stronger than described)

Verified in the toolchain's own source. `crypto/rand.Text` calls `Read`, and `crypto/rand/rand.go`:

```go
// Read calls [io.ReadFull] on [Reader] and crashes the program irrecoverably if
// an error is returned. …
	if err != nil {
		fatal("crypto/rand: failed to read random data (see https://go.dev/issue/66821): " + err.Error())
		panic("unreachable") // To be sure.
	}
```

`fatal` is `runtime.fatal` via `go:linkname`, reaching `fatalthrow` — the runtime's unrecoverable throw. So it is **not** a panic a caller can `recover`; it is a hard process kill. Tearing down a whole index run because a temporary cache file could not be named is the wrong failure mode for a best-effort write whose errors every caller already discards.

For contrast, `os.CreateTemp` never has this exposure: `$GOROOT/src/os/tempfile.go` names its files from `runtime.rand()`, an infallible non-cryptographic source. This code wants the CSPRNG for the O_EXCL race guard, so the answer is not to drop to a weaker source but to read it through an interface that reports.

**Fix:** `cacheTempNameSuffix(source io.Reader)` reads 16 bytes with `io.ReadFull(rand.Reader, …)` and hex-encodes them — same CSPRNG, error returned instead of a process kill.

---

### 3. `internal/sem/cache_entry.go:186` — TOCTOU between `Lstat` and `OpenRoot` — **REFUTED**

The race window is real; the consequence it implies is not. `os.Root` enforces confinement inside the syscall, independently of any `Lstat` the caller did first, so winning the race cannot make the write escape the cache root.

Standalone probe against the same toolchain (`go1.26.0 darwin/arm64`), exercising every operation this code performs through the returned handle:

```
A OpenRoot(link -> ../outside RELATIVE): ERROR: openat link: path escapes from parent
B OpenRoot(link -> ABS outside path):    ERROR: openat link: path escapes from parent
C OpenRoot(link -> sibling INSIDE root): OK (no error)
C sub.Open(in.txt):                      SUCCESS, read "INSIDE-BYTES"
D OpenFile(link/x -> ../outside REL):    ERROR: openat link/x: path escapes from parent
E OpenFile(link/y -> ABS outside path):  ERROR: openat link/y: path escapes from parent
F Rename(src.txt -> link/z.txt):         ERROR: renameat src.txt link/z.txt: path escapes from parent
```

`go doc os.Root`:

> Methods on Root will follow symbolic links, but symbolic links may not reference a location outside the root. Symbolic links must not be absolute.

So the worst a won race can do is case C: redirect the write to **another location inside the same cache root**, which `os.Root` permits by design. That is not a privilege gain — whoever can plant the symlink at that component already has write access to the cache directory and can simply write the target path directly. The `Lstat` is a nicer error ahead of an enforcement that happens anyway, not the boundary itself.

No change made. Closing the window would mean re-deriving through a handle `os.Root` already re-derives, for an outcome `os.Root` already guarantees.

(The one documented exception is `GOOS=js`, where `os.Root` states it *is* vulnerable to symlink TOCTOU and cannot ensure confinement. This tool is not built for js/wasm.)

---

### Tests

In `internal/sem/cache_write_confinement_test.go`:

- `TestCacheComponentRedirectErrorHonoursEveryRedirectingMode` — table over `ModeDir`, `ModeSymlink`, `ModeIrregular|ModeDir`, presented through a stub `os.FileInfo`. The `ModeIrregular` row expects a refusal only where `runtime.GOOS == "windows"`, and expects **no** refusal elsewhere, so it also pins that ordinary irregular entries are not broken on platforms with no reparse points. A junction cannot be created in-process (privileges + `mklink`), and the guard either honours the bit or does not.
- `TestCacheTempNameSuffixReportsRandomnessFailure` — an erroring source and a short source must both be reported.
- `TestCacheTempNameSuffixIsUnpredictable` — 64 distinct 32-character lowercase-hex draws.
- `TestCreateRootTempNamesTheFileFromTheReportedRandomSource` — the call-site guard; the hex shape is what distinguishes it from `rand.Text`.

### Mutation verification

| mutation | result |
|---|---|
| `cacheTempNameSuffix` swallows the read error (`_ = err`) | **caught** — `cacheTempNameSuffix error = <nil>, want it to report randomness source unavailable` |
| `createRootTemp` reverted to `rand.Text()` | **caught** — `temporary cache file named ".snapshot-D32NEQFYB2Q3G2ZOAW35RLNXZD.json.gz", want .snapshot-<32 hexadecimal>.json.gz` |
| guard forced to accept everything (`if true \|\| !pathMayRedirect(info)`) | **caught** — symlink row fails |
| non-Windows `pathMayRedirect` widened to include `ModeIrregular` | **caught** — `mode d?--------- was refused on darwin` |
| guard reverted to `ModeSymlink`-only | **not caught on darwin, as expected** — the two predicates are identical on this platform by construction. That row discriminates only where the modes differ, i.e. on the `windows-latest` leg of `test.yml`. Stated rather than papered over. |

Each mutation was `grep`-confirmed applied before running, and the restore `grep`-confirmed after. `go test ./internal/sem -run 'Cache|Snapshot|Symlink|Confin'` passes; `go vet` and `gofmt -l` clean.

PR #161, opened 2026-08-28
fix(sem): resolve bare calls to Julia module-scoped functions

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/120

## Impact

Every function defined inside a Julia `module ... end` — i.e. essentially every
function in every Julia package — produced **zero CALLS edges**, silently.
`capabilities --json` advertises `CALLS` for Julia, `entire graph neighbors`
and `entire graph impact` return nothing for those symbols, and no warning or
partial-failure is reported. The identical code written at file top level does
emit CALLS, which is why the existing Julia call test never caught it.

## Repro

```julia
# Geo.jl
module Geo
function add(a, b)
    return a + b
end
function point_sum(x, y)
    return add(x, y)   # <- no CALLS edge
end
end
```

```
$ entire graph edges --repo . --format ndjson --relation CALLS
graph: 0 edge(s) matched
```

Delete the `module Geo` / `end` wrapper and the same file emits
`CALLS point_sum -> add`. `DATA_FLOWS point_sum -> add` is emitted in *both*
cases, so the call site is found — only the CALLS edge is dropped.

## Root cause

A Julia definition inside a module is namespace-qualified by the parser
(`parser.go`, `case "function_definition"` / `case "assignment"`):

```go
if scope != "" && !strings.Contains(name, ".") {
    kind = "method"
    name = qualify(scope, name)
}
```

so `Geo.add` is emitted with `Kind == "method"`. `resolveCallTargets` then
rejects it, because a bare `name()` call is not allowed to bind to a method
unless the language says otherwise:

```go
if ... || (to.Kind == "method" && !nameCallMayTargetMethod(from.Language) && !allowMethodTargets) ...
```

`nameCallMayTargetMethod` covered implicit-receiver languages and Rust, but not
Julia. That gate exists to stop `name()` from resolving to a receiver-dispatched
member — a distinction Julia does not have: `f(args...)` is the only call syntax
in the language, and the `method` kind here means "module-scoped", not "member".

## Fix

Add Julia to `nameCallMayTargetMethod`, with the reasoning recorded next to the
other two entries. One predicate, no parser or symbol-shape change, so stable
symbol IDs and every other language are untouched.

## Regression test

`TestJuliaModuleScopedCallsResolve` builds a snapshot over a module-wrapped file
and requires `CALLS Geo.point_sum -> Geo.add` and `CALLS Geo.scaled ->
Geo.point_sum` (long-form and short-form definitions).

Mutation verification:

- Before the fix: FAIL — `missing Julia module-scoped CALLS Geo.point_sum->Geo.add: []`.
- With the fix mutated to `lang == "JuliaMUTANT"`: FAIL, same assertion.
- With the fix: PASS.

## Verification run locally

full gate on a clean runner.

### DIFF
--- a/internal/sem/golden_test.go
+++ b/internal/sem/golden_test.go
@@ -895,3 +895,41 @@ func TestJuliaShortFormAndNestedModuleScope(t *testing.T) {
 		t.Errorf("a nested module's own body call was rejected: %v", edges)
 	}
 }
+
+// TestJuliaOneLineLongFormDefinitionIsNotTheModuleBody pins the last shape the module-body
+// narrowing has to handle.
+//
+// A long-form definition written on ONE line -- `function f(); helper(); end` -- has no
+// body line to blank, and masking its head alone left the body in the module's block, so
+// the module was credited with the call its child makes. The definition is masked
+// textually instead, from its head through the MATCHING `end`; depth is counted over
+// Julia's block openers, because a definition may contain them and the first `end` is
+// then not its own.
+func TestJuliaOneLineLongFormDefinitionIsNotTheModuleBody(t *testing.T) {
+	t.Parallel()
+	repo := t.TempDir()
+	if err := os.WriteFile(filepath.Join(repo, "m.jl"),
+		[]byte("module M; helper()=1; function f(); helper(); end; end\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	snapshot, err := BuildProviderSnapshot(t.Context(), repo, "test-version")
+	if err != nil {
+		t.Fatal(err)
+	}
+	symbolsByID := map[string]SymbolRecord{}
+	for _, symbol := range snapshot.Symbols {
+		symbolsByID[symbol.ID] = symbol
+	}
+	edges := map[string]bool{}
+	for _, relation := range snapshot.Relations {
+		if relation.Type == "CALLS" {
+			edges[symbolsByID[relation.FromID].QualifiedName+" -> "+symbolsByID[relation.ToID].QualifiedName] = true
+		}
+	}
+	if edges["M -> M.helper"] {
+		t.Errorf("the module was credited with a call inside its child's one-line body: %v", edges)
+	}
+	if !edges["M.f -> M.helper"] {
+		t.Errorf("the child's own call was masked away with it: %v", edges)
+	}
+}
--- a/internal/sem/golden_test.go
+++ b/internal/sem/golden_test.go
@@ -977,3 +977,38 @@ func TestJuliaShortFormWithReturnTypeOrConstraint(t *testing.T) {
 		}
 	}
 }
+
+// TestJuliaBlockMaskIgnoresEndInsideAString pins what counts as a block terminator.
+//
+// `function f(); println("end"); helper(); end` contains the word `end` inside a STRING.
+// Counting it as syntax ended the mask early and left the rest of the definition -- every
+// call in it -- attributed to the enclosing module, which is the fabrication the mask
+// exists to prevent, produced by the mask itself.
+func TestJuliaBlockMaskIgnoresEndInsideAString(t *testing.T) {
+	t.Parallel()
+	repo := t.TempDir()
+	if err := os.WriteFile(filepath.Join(repo, "m.jl"),
+		[]byte("module M; helper()=1; function f(); println(\"end\"); helper(); end; end\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	snapshot, err := BuildProviderSnapshot(t.Context(), repo, "test-version")
+	if err != nil {
+		t.Fatal(err)
+	}
+	symbolsByID := map[string]SymbolRecord{}
+	for _, symbol := range snapshot.Symbols {
+		symbolsByID[symbol.ID] = symbol
+	}
+	edges := map[string]bool{}
+	for _, relation := range snapshot.Relations {
+		if relation.Type == "CALLS" {
+			edges[symbolsByID[relation.FromID].QualifiedName+" -> "+symbolsByID[relation.ToID].QualifiedName] = true
+		}
+	}
+	if edges["M -> M.helper"] {
+		t.Errorf("the mask stopped at a string and left the body to the module: %v", edges)
+	}
+	if !edges["M.f -> M.helper"] {
+		t.Errorf("the definition's own call was lost: %v", edges)
+	}
+}
--- a/internal/sem/golden_test.go
+++ b/internal/sem/golden_test.go
@@ -802,3 +802,50 @@ func TestJuliaModuleOwnsOnlyItsOwnBodyAndOnlyItsOwnScope(t *testing.T) {
 		}
 	}
 }
+
+// TestJuliaOneLineModuleKeepsItsOwnCall pins the module-body scan against a module whose
+// child shares its line.
+//
+// `module M; setup() = 1; setup(); end` puts the module, its child and a real module-level
+// call on ONE line. Blanking the child's lines therefore deleted the call the narrowing
+// exists to keep, and a SymbolRecord carries only line numbers, so there is no byte range
+// to mask instead. The child's BODY is blanked and its definition HEAD is masked by name,
+// which says the same thing without deleting whatever else shares the line.
+func TestJuliaOneLineModuleKeepsItsOwnCall(t *testing.T) {
+	t.Parallel()
+	repo := t.TempDir()
+	for name, content := range map[string]string{
+		"one.jl":  "module M; setup() = 1; setup(); end\n",
+		"many.jl": "module N\nfunction helper()\n    return 1\nend\nfunction go()\n    helper()\nend\nhelper()\nend\n",
+	} {
+		if err := os.WriteFile(filepath.Join(repo, name), []byte(content), 0o644); err != nil {
+			t.Fatal(err)
+		}
+	}
+	snapshot, err := BuildProviderSnapshot(t.Context(), repo, "test-version")
+	if err != nil {
+		t.Fatal(err)
+	}
+	symbolsByID := map[string]SymbolRecord{}
+	for _, symbol := range snapshot.Symbols {
+		symbolsByID[symbol.ID] = symbol
+	}
+	edges := map[string]bool{}
+	for _, relation := range snapshot.Relations {
+		if relation.Type == "CALLS" {
+			edges[symbolsByID[relation.FromID].QualifiedName+" -> "+symbolsByID[relation.ToID].QualifiedName] = true
+		}
+	}
+	for _, want := range []string{
+		"M -> M.setup",  // the one-line module still owns its own call
+		"N -> N.helper", // a module-level call in a multi-line module
+		"N.go -> N.helper",
+	} {
+		if !edges[want] {
+			t.Errorf("missing edge %q; got %v", want, edges)
+		}
+	}
+	if edges["N -> N.go"] {
+		t.Errorf("a definition head was read as a call: %v", edges)
+	}
+}
--- a/internal/sem/golden_test.go
+++ b/internal/sem/golden_test.go
@@ -684,3 +684,61 @@ func TestHeuristicRelationTypesMatchDocumentation(t *testing.T) {
 		}
 	}
 }
+
+// TestJuliaBareCallStaysInsideItsOwnLanguageAndScope pins the two edges the Julia method
+// arm must not create.
+//
+// Admitting method targets is what lets a bare `name(args...)` -- Julia's only call form
+// -- reach a module-scoped definition. Two things ride along with it unless refused. The
+// global fallback it draws on is language-agnostic, so an unresolved call bound to a
+// globally unique JAVA method of the same name, which Julia cannot invoke. And a
+// `module M ... end` block spans every nested definition, so scanning the module's own
+// text saw the call names written inside its FUNCTIONS and credited the module with
+// making them -- the same call appearing twice, once from the function that made it and
+// once from a symbol that did not.
+//
+// The intra-module case is asserted alongside because both fixes withdraw candidates, and
+// a withdrawal that also drops the real edge has traded one wrong answer for another.
+func TestJuliaBareCallStaysInsideItsOwnLanguageAndScope(t *testing.T) {
+	t.Parallel()
+	repo := t.TempDir()
+
+	writeFileForTest := func(name, content string) {
+		t.Helper()
+		if err := os.WriteFile(filepath.Join(repo, name), []byte(content), 0o644); err != nil {
+			t.Fatal(err)
+		}
+	}
+	writeFileForTest("m.jl", "module M\nfunction go()\n    runTask()\nend\nend\n")
+	writeFileForTest("T.java", "public class T { void runTask() { } }\n")
+	writeFileForTest("n.jl", "module N\nfunction outer()\n    inner()\nend\nfunction inner()\n    return 1\nend\nend\n")
+
+	snapshot, err := BuildProviderSnapshot(t.Context(), repo, "test-version")
+	if err != nil {
+		t.Fatal(err)
+	}
+	symbolsByID := map[string]SymbolRecord{}
+	for _, symbol := range snapshot.Symbols {
+		symbolsByID[symbol.ID] = symbol
+	}
+
+	intraModule := 0
+	for _, relation := range snapshot.Relations {
+		if relation.Type != "CALLS" {
+			continue
+		}
+		from, to := symbolsByID[relation.FromID], symbolsByID[relation.ToID]
+		if from.Language == "Julia" && to.Language != "Julia" {
+			t.Errorf("Julia %s resolved a bare call to %s/%s, which it cannot invoke", from.Name, to.Language, to.Name)
+		}
+		if from.Language == "Julia" && from.Kind == "module" {
+			t.Errorf("module %s credited with calling %s; the call belongs to the function that wrote it", from.Name, to.Name)
+		}
+		if from.Name == "outer" && to.Name == "inner" {
+			intraModule++
+		}
+	}
+	if intraModule != 1 {
+		t.Fatalf("the legitimate intra-module call outer -> inner was withdrawn too (found %d)", intraModule)
+	}
+}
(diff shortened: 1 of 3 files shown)


ashtom:

**Verdict: clean**

- Live PR head `e0a2b01b`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `e0a2b01b` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

### Independent review state

suhaanthayyil:
Thanks — re-ran the staleness check and it still holds: `main` is unchanged at `bb41f223`, `git rev-list --left-right --count origin/main...refs/prs/161` is still `13 1`.

Rather than repeat the revert check, I verified the behaviour with built binaries on the exact Julia fixture #171 ships (`internal/sem/testdata/langmatrix/julia/Ledger.jl` — a `module Ledgers ... end` with `double` calling `add`):

```
$ g-main edges --repo jl --format ndjson --worktree
relations: {CONSTRUCTS:1, CONTAINS:3, DATA_FLOWS:2, DEFINES:4, PARAM_TYPE:1, USES_TYPE:2}

$ g-161  edges --repo jl --format ndjson --worktree
relations: {CALLS:1, CONSTRUCTS:1, CONTAINS:3, DATA_FLOWS:2, DEFINES:4, PARAM_TYPE:1, USES_TYPE:2}
  CALLS: Ledger.jl:method:Ledgers.double -> Ledger.jl:method:Ledgers.add
```

Every other relation count is byte-identical, and the four `compound-v1` symbol IDs are unchanged — this adds one edge and re-keys nothing.

**False-positive probe** (the risk with resolving bare names to module-scoped methods is that they resolve to the *wrong* module). Two modules in one file, each defining `helper`, each calling it bare:

```
$ g-161 edges --repo jl2 --format ndjson --worktree | grep CALLS
   Alpha.run_alpha -> Alpha.helper
   Beta.run_beta   -> Beta.helper
```

Same-module only. No cross-module fan-out.

---

### One finding the merge check cannot see: this PR turns `main` red once #171 lands

`git merge-tree` reports no conflict between this PR and #171, and each is green alone — but the merged tree fails:

```
$ git merge --no-ff refs/prs/171 refs/prs/161     # clean, no conflicted files
$ go test ./internal/sem -run TestLanguageMatrixCallsGapsAreExactlyAsAudited -count=1
--- FAIL: TestLanguageMatrixCallsGapsAreExactlyAsAudited/julia
    language_matrix_test.go:234: Julia now extracts CALLS; remove its audited gap
      (a Julia definition inside `module ... end` is emitted as a method, and bare
      calls are barred from resolving to methods) from languageMatrix
```

#171 records this precise gap as audited, and its test is designed to fail in *both* directions — including when a gap is closed without the entry being retired. This PR is the fix for that gap and never touches `language_matrix_test.go` (`git diff --name-only` on this branch: `internal/sem/provider.go`, `internal/sem/provider_test.go`).

I checked the same merge for every other open PR against #171 — #162, #165, #168, #169, #170, #172, #176, #178, #182, #183 are all green. This is the only pair that breaks. (#176 and #178 are stacked on #171 and each already retires or rewrites its own `callsGap` entry, which is why they pass.)

Whichever of the two lands second must, in the same commit, drop julia's `callsGap` and promote `"CALLS"` into its required relations. Verified:

```
relations: []string{"DEFINES", "CONTAINS", "CALLS", "CONSTRUCTS", "PARAM_TYPE", "USES_TYPE"},
(callsGap line deleted)

$ go test ./internal/sem -run 'TestLanguageExtractionMatrix|TestLanguageMatrixCallsGapsAreExactlyAsAudited|TestLanguageMatrixCoversEverySemanticLanguage' -count=1
ok
```

It cannot be pre-applied to either branch. On #171 alone the same edit fails the other way — `Julia advertises CALLS but extracted none from its fixture` — so this is strictly a merge-order action, not a defect in either PR.

suhaanthayyil:
## This branch is a silent merge landmine with #171 — fixed on #171's side, nothing to change here

#171 adds a language matrix whose `TestLanguageMatrixCallsGapsAreExactlyAsAudited` lists Julia as an
audited CALLS gap. This branch closes that gap. The two touch disjoint files and merge with no
textual conflict, so `git merge-tree` sees nothing — and then the merged tree fails:

```
$ git checkout -q 2c599378 && git merge --no-edit e0a2b01b
Merge made by the 'ort' strategy.
--- FAIL: TestLanguageMatrixCallsGapsAreExactlyAsAudited/julia
    language_matrix_test.go:234: Julia now extracts CALLS; remove its audited gap (a Julia
    definition inside `module ... end` is emitted as a method, and bare calls are barred from
    resolving to methods) from languageMatrix
```

The assertion is deliberately exact, so a language that stops being a gap breaks it. Neither
branch's CI can see this; it appears only after the merge.

## Where the fix went, and why not here

The natural patch — retire Julia from the gap list — is only correct *after* this branch lands. It
would make #171 fail on its own, which just moves the landmine rather than removing it. So the fix
went to #171, which now derives the audit from the extractor rather than pinning it as a literal:

```go
callsGapHolds: func(languageMatrixResult) bool { return !nameCallMayTargetMethod("Julia") },
```

The Julia gap reduces exactly to bare-call resolution skipping method targets, which is the single
predicate this branch flips. The audit entry therefore retires itself in whichever commit admits
Julia, and both merge orders are green:

| tree | before | after |
|---|---|---|
| #171 alone | pass | pass |
| #171 + this branch | **FAIL** (julia) | pass |

`TestJuliaModuleScopedCallsResolve` and the whole language matrix pass together on the merged tree.

**No change is needed on this branch.** One thing to keep in mind if it is revised: `nameCallMayTargetMethod`
is now read by name from that test, so renaming or removing it will break #171's compile. That is a
loud failure rather than a silent one, which is the trade being made deliberately.

ashtom:

no longer does.

```
01M139XJK2C9  Medium  open    current    internal/sem/provider.go:1861  This admits every module-scoped Julia definition into the file-wide bare-name ca…
```

### Verified vs. assumed

suhaanthayyil:
## Finding addressed — `internal/sem/provider.go:1861`

**VALID.** Pushed `caeb2681`.

### Reproduction

Two modules in one file, plus a top-level function, all indexed on this branch before the fix:

```julia
module A
function target(x); return 1; end
# ... 28 padding lines ...
function caller(y)
    return target(y)      # line 36
end
end

module B
function target(x); return 2; end   # line 44
end

function toplevel(z)
    return target(z)      # line 50
end
```

Emitted CALLS **before**:

```
A.caller  -> B.target      <- wrong module
toplevel  -> B.target      <- top level cannot see inside a module
```

Two more from a nested-module fixture (`module Outer` containing `module Inner`):

```
module:Inner -> Outer.shared      <- parent's binding, not in scope
module:Outer -> Inner.useOuter    <- child's binding, not in scope
```

On `origin/main` the same fixtures emit **zero** CALLS (methods were excluded from name resolution altogether), so every one of these four edges is introduced by the method arm this PR opened.

The lexically-nearest tiebreak is what picks the wrong one: `B.target` (line 44) is nearer to the call on line 36 than `A.target` (line 3).

### Fix

`provider.go` — `juliaModuleScopedTargetReachable(from, to)`, applied in both candidate loops of `resolveCallTargets`. A `module M ... end` is a hard namespace boundary: from outside M its definitions are reachable only as `M.name(...)`. The gate is deliberately narrow — same file, `method` kind — so it can only withdraw candidates the method arm itself introduced, never a target that resolved before it. A caller that is itself a module symbol is scoped to its own body.

After:

```
A.caller -> A.target       <- correct target now resolves
(the other three are gone)
```

### Test

`TestJuliaBareCallsStayInsideTheirModule` (`provider_test.go`). Asserts the in-module edge is present and that `A.caller->B.target`, `toplevel->A.target` and `toplevel->B.target` are absent.

### Mutation

Removed both guard clauses (` || !juliaModuleScopedTargetReachable(from, to)` and ` && juliaModuleScopedTargetReachable(from, to)`), leaving the helper defined. `grep` confirms zero call sites remain; `go build` succeeds — the mutation compiles.

```
--- FAIL: TestJuliaBareCallsStayInsideTheirModule
    provider_test.go:12551: missing in-module Julia CALLS A.caller->A.target
--- PASS: TestJuliaModuleScopedCallsResolve
```

Restored → both PASS.

### main vs branch

Same fixture on `origin/main`: identical symbols, zero CALLS. The gate cannot remove anything main emitted — main never resolved a bare Julia call to a method.

### Golden delta

**None.** `julia-r-basic` has no modules, so its definitions are `function`, not `method`; the gate does not apply. `TestProviderGoldenSnapshots`, `TestProviderGoldenFixtureQualityCoverageReport` and `TestCapabilityMatrixCoversEmittedRelations` pass unmodified and `testdata/` stays clean. `git diff origin/main -- internal/sem/testdata` is empty.

### Also checked

LF and CRLF fixtures produce the identical CALLS set. All four `TestJulia*` tests pass. `gofmt`/`go vet` clean.

suhaanthayyil:

- TestJulia, plus `TestProviderGoldenSnapshots` — pass on the merged tree

suhaanthayyil:
Dismissed `01M1J9FBH8M1` (nested function cannot resolve a bare call) — **the premise does not hold for Julia in this parser.**

The finding assumes a nested function is qualified `M.outer.inner`, so that `juliaScopeName` would return `M.outer` and the module comparison would fail. Measured on exactly that source:

```julia
module M
function helper() ... end
function outer()
    function inner()
        helper()
    end
    inner()
end
end
```

```
SYM M           kind=module  scopeName="M"
SYM M.helper    kind=method  scopeName="M"
SYM M.outer     kind=method  scopeName="M"
SYM M.inner     kind=method  scopeName="M"     ← flattened, not M.outer.inner

CALL M.inner -> M.helper    ← resolves
CALL M.outer -> M.helper
CALL M.outer -> M.inner
```

The parser flattens nested functions to the enclosing module, so `juliaScopeName(inner)` is `M` and the gate accepts the call. No edge is dropped.

The nested-**module** case the finding is adjacent to was real and is fixed separately (`01M1J89KB4AD`): there the child's container genuinely is qualified (`Outer.Inner`), and comparing the module's short name did reject valid calls. That one needed the qualified name; this one needs nothing.

Reopen if a shape does produce `M.outer.inner` — I could not construct one.

suhaanthayyil:
Two defects found while fixing `01M1JH8FGDM3`, **both reproduce on `main`**, neither is fixed here — they are separate root causes and I did not want to widen this PR.

### 1. `function Base.show(...)` inside a module gets no container

```julia
module M
function Base.show(io, x)
    helper(x)
end
helper(y) = y
end
```

`containerName("Base.show")` is `"Base"`, which is looked up in the per-file `byName` map and misses. Because that lookup is an `if`, the lexical-parent fallback at `provider.go:1662-1671` is in the `else if` and never runs — so the symbol ends up with **`ContainerID = ""`**.

Two consequences: its call to `M.helper` is dropped, *and* the same call is misattributed to module `M` (`CALL M -> M.helper`). Extending a method on an external module is ordinary Julia, so this is not an exotic shape.

The enclosing-module walk added in this PR cannot reach it — the container chain is empty, so there is nothing to walk. The fix belongs in container assignment: fall back to the lexical parent when a dotted container name resolves to nothing in this file.

### 2. A module reopened as two blocks in one file loses its own calls

```julia
module M
helper(y) = y
end

module M
caller() = helper(1)
end
```

`M.caller -> M.helper` is dropped at the `to.ContainerID == scope.id` comparison, because the two `module M` blocks are distinct symbols with distinct IDs.

This one is notable because it **contradicts the comment on that same function**, which says module *names* are compared rather than container ids precisely because "a Julia module routinely spans files through `include` and each file gives it a different container id while it stays one namespace." The same reasoning applies within a single file, but the id comparison on the next line does not follow it.

Happy to fix either or both — say the word. I left them out because #1 is in container assignment rather than call scoping, and #2 needs a decision about whether same-name modules in one file should be treated as one namespace (I believe they should, per that comment, but it changes the id contract).

suhaanthayyil:
Stepping back from the individual fixes: **this branch now carries 14 commits that are all the same bug**, and I think the root cause is fixable in one move.

Every one of them is line granularity applied to a language that does not respect lines:

| commit | what actually went wrong |
|---|---|
| mask a child's definition head, not its whole line | a line held a child *and* module code |
| stop the short-form mask at a semicolon | `;` separates two things on one line |
| mask a one-line definition through its matching `end` | definition and body share a line |
| definitions on one line cross-attributing calls | two definitions, one line |
| keep module-level code written after a child's `end` | `end; setup()` — two owners, one line |
| do not read `end` inside a string / `#=` comment | scanning text for a token the parser already found |

Each was fixed by scanning text — regexes, depth counters, comment maskers, width-preserving stand-ins. **That machinery is reconstructing information the parser already had and the struct already carries.**

`SymbolRecord` has `sourceStartByte` / `sourceEndByte` — "the parser's exact declaration range" — and `exactSymbolSource(content, symbol)` already reads them (`provider.go:24014`). They are private only so the frozen schema and symbol IDs do not change; the Julia code is in the same package and can read them directly. **No schema change is needed.**

They are populated, exact, and survive into the relation pass. On the fixture that forced the `end; setup()` fix:

```
module:M       lines=[1,8]  exact="module M\nfunction setup()\n    1\nend\nfunctio..."
method:setup   lines=[2,4]  exact="function setup()\n    1\nend"
method:f       lines=[5,7]  exact="function f()\n    2\nend"      <-- stops AT `end`
```

`f`'s range already excludes the trailing `; setup()`. And on the one-line module:

```
module:M         [0,54)  "module M; helper()=1; function f(); helper(); end; end"
method:M.helper  [10,20) "helper()=1"
method:M.f       [22,49) "function f(); helper(); end"
```

A module's own body is then just its span minus its children's spans — `[0,10) + [20,22) + [49,54)` — computed by subtraction, exactly, with no regex, no `end` counting, no comment masking, and no head-width padding. The call scan uses `symbolBlockFromLines`; the byte-exact sibling is right beside it.

I have **not** made that change — it is a different shape of PR from the one under review here, and it is your call whether it belongs in this branch, a follow-up, or not at all. But I would expect it to subsume most of the 14 commits above and close the whole class rather than the next instance of it. Happy to do it if you want it.

(The textual masking is still needed where a symbol's *interior* must be masked without a child record to subtract — the F# `(* *)` and Julia `#=` comment maskers earn their keep. It is the *block boundaries* that should come from bytes.)

suhaanthayyil:
**Correction — please disregard item #2 from my earlier comment.**

I reported that "a module reopened as two blocks in one file loses `M.caller -> M.helper`" as a defect, and argued it contradicted the comment on `juliaModuleScopedTargetReachable` about a module spanning files. **That was wrong on both counts.** Do not act on it.

**Julia does not reopen a module.** A second `module M ... end` binds a *new* module and rebinds the name (Julia prints `WARNING: replacing module M`). In:

```julia
module M
helper(y) = y
end

module M
caller() = helper(1)
end
```

`helper` is simply not defined inside the second `M`; calling `caller()` raises `UndefVarError`. So dropping that edge is **correct behaviour**, not a bug. The snapshot agrees — the two blocks get distinct ids:

```
SYM module qn="M"        id="module:M#sig:cccffc37bdcd9872"
SYM method qn="M.helper" id="method:M.helper"
SYM module qn="M"        id="module:M#sig:79ca9e19e0f35174"
SYM method qn="M.caller" id="method:M.caller"
```

**And the comment I appealed to was itself false.** While fixing `01M1JKJAWWXY` we checked the premise instead of trusting it: a genuine `include`-spanned definition is emitted as

```
SYM kind=function qn="sharedHelper" file=src/shared.jl containerID=""
```

— `kind=function` with **no container**, so it leaves the gate at the `to.Kind != "method"` line and never reaches a name comparison at all. The name comparison that comment justified was therefore only ever admitting *distinct* same-named modules, which is exactly the cross-file false edge that finding reported. It has been removed, and `juliaScopeName` deleted as dead.

So the "contradiction" I pointed at was between correct behaviour and an incorrect comment. Item **#1** in that earlier message — `function Base.show(...)` inside a module getting `ContainerID = ""`, its call dropped *and* misattributed to the module — I re-checked and it still stands.

Sorry for the noise on #2.

suhaanthayyil:
Follow-up to my byte-range comment, now with a **working demonstration** rather than an argument.

I prototyped the whole idea as ~15 lines — a module's own body is its byte span with each child's byte span blanked — and ran it against the five hardest fixtures this branch has accumulated. No regex, no depth counting, no comment masking, no generator discriminator:

```go
out := []byte(src[mod.sourceStartByte:mod.sourceEndByte])
for _, c := range children {           // c.ContainerID == mod.ID
    for i := c.sourceStartByte; i < c.sourceEndByte; i++ {
        if out[i-base] != '\n' { out[i-base] = ' ' }
    }
}
```

Results — every one correct:

| fixture | byte-subtraction gives | what it currently takes |
|---|---|---|
| `sum(x for x in xs)` generator | `module M setup() end` | `juliaMaskGeneratorClauses` + token stack + bracket/block discriminator (+102 lines) |
| `function f(); #= end =#; helper(); end` | `module M end` | nesting-aware `maskJuliaBlockComments` |
| `module M; helper()=1; function f(); helper(); end; end` | `module M; ; ; end` | `juliaMaskSiblingDefinitionHeads` |
| `end; setup()` | `module M ; setup() end` | `juliaBlankThroughClosingEnd` + `juliaBlankAfterClosingEnd`, both depth-counting |
| `function f(); end; setup()` | `module M ; setup() end` | the `lineStart <= 0` guard fix |

Each row is a separate commit on this branch. The right-hand column is the machinery; the left-hand column is what the parser already knew.

The generator row is the sharpest: distinguishing a comprehension `for` from a loop `for` needed a stack tracking whether the innermost open construct is a bracket or a block. Byte subtraction never asks the question — the generator is inside `f`'s span, so it is gone, and `setup()` is outside it, so it stays.

**Honest caveats**, so this is not oversold:
- My prototype wired `ContainerID` by byte containment as a stand-in. Production has real `ContainerID`, which is *better* — except where it is empty, e.g. the `function Base.show(...)` case I flagged earlier (`ContainerID = ""`). That symbol would not be subtracted, so **that container-assignment bug should be fixed first** or it becomes the one remaining hole.
- Interior masking still earns its keep. Masking *inside* a symbol with no child record to subtract — the F# `(* *)` and Julia `#=` maskers — stays. It is only the block *boundaries* that should come from bytes.
- `sourceStartByte`/`sourceEndByte` are private for schema stability; this needs no schema change, since all of this is in package `sem`.

Still not making the change — it is yours to call. But the evidence is now empirical rather than a claim.

ashtom:
Closing in favor of #190, which implements the same core Julia bare-call fix from current main with a smaller, conservative scope. #190 keeps ordinary same-file, same-module method calls while intentionally deferring module-body, inner-constructor, and nested/local-function callers.

PR #162, opened 2026-08-28
fix(sem): extract HCL top-level attributes as settings

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/121

## Impact

`.tfvars` is a registered extension mapped to HCL, a language the provider
advertises in `semantic_languages`. A Terraform variables file consists of
nothing but top-level attribute assignments — the format rejects blocks — and
the HCL walker extracts only `block` nodes. The result: **every `.tfvars` and
`*.auto.tfvars` file in every repository produces zero symbols and zero
relations**, with no warning and no partial failure. The file is read, parsed
and counted as HCL, and nothing comes out. Hand-written Consul/Nomad/Vault
`.hcl` configs lose their top-level settings (`datacenter = "dc1"`,
`data_dir = "/opt/consul"`) for the same reason.

Other config languages already carry key/value symbols: TOML emits `setting`,
JSON/JSON5 and YAML emit `section`, CUE emits `field`. HCL was the outlier.

## Repro

```hcl
# prod.tfvars
region         = "us-west-2"
instance_count = 3
```

```
$ entire graph symbols --repo . --format ndjson | grep prod.tfvars
$ entire graph edges   --repo . --format ndjson | grep prod.tfvars
```

Both empty. The summary record does not even list HCL among the snapshot's
languages when the repository holds only attribute-shaped HCL.

## Root cause

`entityFromNode` (`internal/sem/parser.go`) has a `case "block"` for HCL and no
case for `attribute`, the node type tree-sitter-hcl produces for
`identifier = expression`. There is nothing else to fall back to: HCL is
grammar-backed, so it never reaches `fallbackEntities`.

## Fix

Add `case "attribute"` gated to HCL, emitting `kind: "setting"` (the same kind
TOML uses for a config key) named by the attribute's `identifier` child.

Only **top-level** attributes are extracted. `hclTopLevelAttribute` checks that
the attribute's `body` parent is the `config_file` itself, not a `block`, so an
argument inside a Terraform resource stays folded into the block symbol and a
large module does not fan out into one symbol per argument. That keeps the
change confined to the file class that was returning nothing at all.

`hclResourceDependsOnRelations` and `configTargets` both key on
`Kind == "block"` and are unaffected; RESOURCE_DEPENDS_ON and CONFIGURES output
is byte-identical for existing HCL.

## Regression test

`TestHCLTopLevelAttributesEmitSymbols` asserts:

- `prod.tfvars` emits `setting` symbols for `region`, `instance_count`, `tags`;
- a top-level `datacenter` in a `.tf` file emits one too;
- the `resource "aws_s3_bucket" "logs"` block symbol still comes out as `block`;
- the `bucket` argument *inside* that block does **not** become a symbol.

Mutation verification (both directions bind):

- Before the fix: FAIL — `HCL top-level attribute "prod.tfvars:region" not extracted as a setting: map["main.tf:logs":"block"]`.
- Mutating the guard to `|| true` (never extract): FAIL, same assertion.
- Mutating away the `hclTopLevelAttribute` check (extract every attribute): FAIL — `attribute nested in a block must not be extracted`.
- With the fix: PASS.

## Verification run locally

full gate on a clean runner.

### DIFF
--- a/internal/sem/parser.go
+++ b/internal/sem/parser.go
@@ -4997,6 +4997,26 @@ func entityFromNode(node *sitter.Node, src []byte, language, scope string) (Enti
 	case "block":
 		kind = "block"
 		name = hclBlockName(node, src)
+	case "attribute":
+		// HCL top-level attribute (`region = "us-west-2"`). A Terraform
+		// variables file (`.tfvars`, `*.auto.tfvars`) contains nothing else —
+		// the format rejects blocks — so extracting only `block` nodes left
+		// every such file with zero symbols and zero relations while HCL was
+		// advertised as a semantic language. Hand-written Consul/Nomad/Vault
+		// `.hcl` configs carry the same top-level settings. Attributes nested
+		// inside a block are that block's body and stay folded into the block
+		// symbol, which keeps a Terraform module from fanning out one symbol
+		// per argument. The node type is HCL-only in this grammar set, but the
+		// gate is explicit so another language's `attribute` node cannot leak
+		// into this case.
+		if language != "HCL" || !hclTopLevelAttribute(node) {
+			return Entity{}, false
+		}
+		name = hclAttributeName(node, src)
+		if name == "" {
+			return Entity{}, false
+		}
+		kind = "setting"
 	case "field":
 		kind = "field"
 		name = cueFieldName(node, src)
@@ -7747,6 +7767,29 @@ func hclBlockName(node *sitter.Node, src []byte) string {
 		}
 	}
 	return strings.Join(parts, ".")
+}
+
+// hclTopLevelAttribute reports whether an `attribute` node sits directly in the
+// file body rather than inside a block. tree-sitter-hcl nests every attribute
+// under a `body`, so the distinguishing parent is the body's own parent:
+// `config_file` for a top-level setting, `block` for a block argument.
+func hclTopLevelAttribute(node *sitter.Node) bool {
+	body := node.Parent()
+	if !validNode(body) || body.Type() != "body" {
+		return false
+	}
+	root := body.Parent()
+	return validNode(root) && root.Type() == "config_file"
+}
+
+// hclAttributeName returns the name an HCL attribute binds — the identifier on
+// the left of the `=`.
+func hclAttributeName(node *sitter.Node, src []byte) string {
+	identifier := firstNamedChildOfType(node, "identifier")
+	if !validNode(identifier) {
+		return ""
+	}
+	return strings.TrimSpace(identifier.Content(src))
 }
 
 func hclBlockPart(node *sitter.Node, src []byte) string {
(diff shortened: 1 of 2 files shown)


ashtom:

**Verdict: clean**

- Live PR head `5516320d`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `5516320d` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

### Independent review state

suhaanthayyil:
Thanks. Staleness still holds — `main` is unchanged at `bb41f223`, still `13 behind / 1 ahead`.

Instead of repeating the revert check, I measured the change with built binaries on a three-file HCL corpus covering the shapes the PR description names (a `.auto.tfvars`, a Terraform module, a hand-written Consul config):

```
$ g-main symbols --repo hcl --format ndjson --worktree
block   arn            local/hcl:HCL:main.tf:block:output.arn
block   ledger         local/hcl:HCL:main.tf:block:resource.aws_s3_bucket.ledger
block   ports          local/hcl:HCL:consul.hcl:block:ports
block   region         local/hcl:HCL:main.tf:block:variable.region

$ g-162 symbols --repo hcl --format ndjson --worktree
(the four above, unchanged, plus)
setting bucket_name    local/hcl:HCL:prod.auto.tfvars:setting:bucket_name
setting instance_type  local/hcl:HCL:prod.auto.tfvars:setting:instance_type
setting region         local/hcl:HCL:prod.auto.tfvars:setting:region
setting replica_count  local/hcl:HCL:prod.auto.tfvars:setting:replica_count
setting data_dir       local/hcl:HCL:consul.hcl:setting:data_dir
setting datacenter     local/hcl:HCL:consul.hcl:setting:datacenter
setting log_level      local/hcl:HCL:consul.hcl:setting:log_level

removed: 0   added: 7
```

Three things that confirms rather than assumes:

1. **The premise is real.** `prod.auto.tfvars` contributes zero symbols on `main` — the file is entirely invisible while HCL is advertised as semantic.
2. **Strictly additive.** Zero existing `compound-v1` IDs removed or re-keyed, so no consumer re-keys anything. Unlike the C-family set, this one is membership-only.
3. **The nesting gate works in practice, not just in the unit test.** `ports { http = 8500 }` yields `block:ports` and no `setting:http`, so a Terraform module does not fan out one symbol per argument. That is the part of `hclTopLevelAttribute` most likely to be wrong, and it is right.

Two smaller checks:

- `setting` is not a new kind. It already exists on `main` (`internal/sem/parser.go:3254`) and `internal/sem/types.go:2444` already pairs `section`/`setting`, so consumers see a kind they already handle.
- Merged with #171 (the language matrix), the hcl fixture and all 36 language subtests stay green — the block assertions `block:region`, `block:ledger`, `block:ledger_bucket` are unaffected by the new attribute case.

PR #163, opened 2026-08-28
fix(sem): F# qualifier and module-path resolution, plus a forward-pipe precision fix

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/122

> **Description rewritten 2026-09-04.** The original framing — "extract F# forward-pipe call sites" — is **stale**. `main` has since grown `internal/sem/fsharp.go`, whose `fsharpPipelineApplied` already extracts forward-pipe calls, and whose `fsharpJuxtapositionCallIdentifiers` also handles the bare `add 1 2` form this PR explicitly declined to attempt. That headline is done, and not by this branch.

The branch is **not** subsumed. 37 of its 38 commits are a different layer, and none of it exists on `main` under any name.

## What this actually adds

F# name resolution, which `main` has no equivalent of:

- namespace scopes and project-wide module paths
- qualifier resolution — relative, root-anchored `global.`, and caller-outwards
- module abbreviations
- per-call-site qualifier shadowing
- masking for interpolated, triple-quoted and verbatim strings, and for block comments
- multi-line function headers, `let … in`, and tuple destructuring

## A live precision bug on `main`

`main`'s `fsharpPipelineApplied` accepts any name preceded by the two bytes `|>`. F# permits a symbolic operator whose name merely *contains* those bytes, and its right operand is that operator's argument — not an applied function. So on `main` today:

```fsharp
value +|> helper
value $|> helper
```

both fabricate a `CALLS custom → helper` edge. A wrong edge is worse than a missing one, and this branch's `fsharpIsForwardPipeOperator` token walk is what prevents it.

## How the overlap was resolved

`main`'s implementation is **kept**, not reverted. Two tests on this branch (`TestFSharpCustomOperatorContainingAPipeIsNotACall`, `TestFSharpCustomOperatorContainingADollarIsNotACall`) failed after merging `main`, because they assert a property `main`'s newer scanner breaks. Rather than delete the tests or revert the scanner, `fsharpPipelineApplied` now measures the whole operator token via `fsharpIsForwardPipeOperator`, so only `|>`, `||>` and `|||>` apply a function. Only the two-byte lookback is corrected.

Merged with `origin/main`; three conflicting regions in `provider.go`, all "both sides inserted at the same point", all resolved by keeping both. `main`'s juxtaposition scan now runs alongside this branch's qualifier recording — a juxtaposition-only name records no qualifier and resolves unrestricted, which is correct, since a juxtaposition call is bare.

## Verification

`go test ./... -count=1` exit 0 (4,620 tests, 21 packages). `-run 'TestFSharp'` → 263 pass, 0 fail. `gofmt -l -s .` empty, `go vet ./...` clean.


ashtom:

**Verdict: changes required — overlaps and conflicts with #178**

- Live PR head `414dfd9b`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `414dfd9b` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

**1. (blocking, cross-PR) This PR and #178 are two different solutions to the same problem, and they
conflict.** Both add F# call extraction for application without a dot or parentheses:

- this PR adds `fsharpPipelineCallRe` + `fsharpPipelineCallIdentifiers` in `call_scanners.go`
  (pipe position only — deliberately narrow);
- **#178** adds a new 216-line `internal/sem/fsharp.go` doing general juxtaposition.

```
$ git merge-tree --write-tree refs/prs/163 refs/prs/178   ->  CONFLICT
```

**#178's own header comment already assumes this PR has landed** — it reasons about
"`fsharpPipelineCallRe` needs a `|>`", but that identifier exists **only on this branch**:

```
$ git grep -l fsharpPipelineCallRe origin/main    # not present
$ git grep -l fsharpPipelineCallRe refs/prs/178   # only in a comment, never defined
$ git grep -l fsharpPipelineCallRe refs/prs/163   # internal/sem/call_scanners.go  <-- defined here
```

So #178 is written *on top of* this PR conceptually but is not stacked on it in git. Please decide
the sequence explicitly: either land this first and rebase #178 onto it (in which case #178's
comment becomes true), or fold the pipe case into #178 and close this.

The change itself is sound in isolation — tests pass merged, and fail without it.

### Independent review state

suhaanthayyil:
## Sequencing settled, and one real false positive found in the scanner

Pushed `2488addd`.

### Finding — "this PR and #178 are two different solutions to the same problem" — **REFUTED as stated**

They are complementary, not redundant, because they are bounded differently:

- `fsharpPipelineCallIdentifiers` (this PR) reads the pipe position and resolves **any** name, including one the file does not define;
- `fsharpJuxtapositionCallIdentifiers` (#178) reads general application but only for names that are **callable bindings in the same file**.

Demonstrated on a tree with both applied:

```
let run x = x |> Other.transform    this PR -> transform      #178 -> (nothing)
let run x = normalize x             this PR -> (nothing)      #178 -> normalize
```

Neither subsumes the other. Dropping this PR loses every cross-module piped call; dropping #178 loses ordinary F# application. Closing either one loses real edges, so "fold the pipe case into #178 and close this" would be a regression.

The `git merge-tree` conflict is real but is a pure adjacent-insertion in one file: both PRs append a `for name := range ...` loop directly after `fsharpDottedCallIdentifiers` in `internal/sem/provider.go`. Resolution is to keep both loops — no logic changes, no test changes. Verified: on `origin/main` + this PR + #178 with that resolution, `go vet` passes, `gofmt -s -l` is clean, and both PRs' F# suites plus the language matrix pass together.

The forward reference is fixed on the other side: #178 no longer names `fsharpPipelineCallRe`, so its header comment is true on its own branch and stays true after this one lands.

### Sequencing

Recommended order: **`audit/lang-matrix` -> #163 (this PR) -> #178**, all three kept as separate PRs.

- #178 is based on `audit/lang-matrix`, not `main` — that constraint exists independently of this PR.
- This PR is based on `main`, is small, and is independent — it can land at any point.
- Whoever lands second keeps both loops in the `if file.Language == "F#"` block. That is the whole resolution.

### New finding on this PR: `<|>` was read as a forward pipe — fixed

Found while reviewing the regex adversarially, and reproduced before fixing:

```
let p = digit <|> helper        got map[helper:{}], want []
let p = digit<|>helper          got map[helper:{}], want []
```

`<|>` is the alternative combinator FParsec and FSharpPlus define. It contains a literal `|>`, so `fsharpPipelineCallRe` matched it and reported the right operand as a function applied to the left one. It applies nothing — the right operand is an alternative value — so the edge was invented. This fires in any F# codebase using parser combinators, which is exactly the pipe-heavy code this PR targets.

Fix: `fsharpBackwardAlternative` at `internal/sem/call_scanners.go:292-298`, called from `fsharpPipelineCallIdentifiers` at `:269`. It walks back over the leading run of `|` before judging the preceding byte, so the genuine `||>` and `|||>` forward pipes — which differ from `|>` only in arity — are unaffected. Scanning switched to `FindAllStringSubmatchIndex` so the byte before the operator is available.

Tests added to `TestFSharpPipelineScannerSkipsLambdasAndLiterals` in `internal/sem/provider_test.go`:

```
{"triple pipe",                          "let f a b c = (a, b, c) |||> helper", []string{"helper"}}
{"alternative combinator is not a pipe", "let p = digit <|> helper",            nil}
{"alternative combinator without spaces","let p = digit<|>helper",              nil}
{"backward pipe",                        "let f x = helper <| x",               nil}
{"pipes chained without spaces",         "let f xs = xs|>helper|>g",            []string{"helper", "g"}}
```

The last one pins the index-based rewrite: without it, an obvious "consume the preceding character in the regex" fix silently drops `g`.

Mutation check — `fsharpBackwardAlternative` stubbed to `return false`, which compiles:

```
--- FAIL: .../alternative_combinator_is_not_a_pipe      got map[helper:{}], want []
--- FAIL: .../alternative_combinator_without_spaces     got map[helper:{}], want []
```

Restored: pass.

### Verification on the merged tree

`HEAD` merged onto current `origin/main` (clean), then, on that tree: `go vet ./internal/sem/` pass, `gofmt -s -l internal/sem/` clean, `go test ./internal/sem/ -run 'TestFSharp|TestLanguageExtractionMatrix|TestCapabilityMatrixCoversEmittedRelations'` 18 pass.

### Probes that came back clean

`|> fun`/`|> function` (already ignored), `||>`, `|||>`, `xs |> List.map f` (resolves `map`), `helper <| x`, `if a || b > c then`, pipes inside strings and line comments.

ashtom:

Head moved `414dfd9b` → `2488addd`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### My earlier cross-PR finding still stands

`2488addd` fixes the `<|>` false positive, which is good. It does not change the overlap with
**#178**: both PRs still add F# call extraction for application without a dot, and they still
conflict with each other. Please sequence them explicitly.

```
01M17FZP79BG  Medium  open    current    internal/sem/provider.go:3288      The new scanner is wired only into per-symbol `callBlock` processing, while the …
01M17FZP6Z5B  Medium  open    current    internal/sem/call_scanners.go:276  Reducing a qualified target such as `A.map` to `map` discards the module qualifi…
```

### Verified vs. assumed

suhaanthayyil:
## Both findings addressed — pushed `60d34051`

---

### 1. `internal/sem/provider.go:3288` — scanner wired into only one path — **VALID**

The file-level pass (`provider.go`, the `topLevel` branch) wired `fsharpDottedCallIdentifiers` only. A pipeline written as a statement rather than inside a binding belongs to no symbol, so it is scanned there, not by the per-symbol pass — which is exactly the shape of an `.fsx` script.

`script.fsx`, indexed on this branch before the fix:

```fsharp
let normalize (x: string) = x.Trim()
let shout (s: string) = s.ToUpper()

"  hi  " |> normalize |> shout |> ignore
```

```
CALLS = 0
```

A script whose whole body is one pipeline produced nothing. After the fix:

```
file:script.fsx -> normalize
file:script.fsx -> shout
```

(The module-body case already worked, because `let value = x |> normalize` binds a `variable` symbol that carries its own call block — which is why this only shows up at true statement position.)

---

### 2. `internal/sem/call_scanners.go:276` — qualifier discarded — **VALID, and it was not only the new scanner**

`A.convert` reduced to `convert` binds to whichever same-named definition is nearest. Demonstrated with two modules in one file:

```fsharp
module A =
    let convert (x: int) = x + 1     // line 2

module B =
    let convert (x: int) = x * 2     // line 5

module Caller =
    let parens (x: int) = A.convert(x)
    let apply  (x: int) = A.convert x
    let piped  (x: int) = x |> A.convert
```

Worth recording that this is **not new in this PR**. On `origin/main`, with none of this PR's code, the two pre-existing F# call forms already bound to the wrong module:

```
main:    parens -> convert@5 (module B)      <- wrong
         apply  -> convert@5 (module B)      <- wrong
         (piped: no edge, no scanner yet)
```

The pipeline scanner reduced the qualifier exactly as `fsharpDottedCallIdentifiers` already did, so it produced the same wrong target as its two siblings. Fixing it in the pipeline scanner alone would have left one of three F# call forms in the same file resolving differently from the other two, so it is fixed for all three.

**Fix:** the F# scanners now return targets as written (`fsharpDottedCallTargets`, `fsharpPipelineCallTargets`, `fsharpCallTargets`); resolution still matches on the name, but when the qualifier names a module the file itself declares, same-file candidates are held to that module. A qualifier naming another file's module — `InstallProcess.InstallIntoProjects` — and a name that also appears unqualified are left unrestricted and resolve by name exactly as before, so no cross-file edge changes.

After, all three forms agree:

```
parens -> convert@2 (module A)
apply  -> convert@2 (module A)
piped  -> convert@2 (module A)
```

---

### Tests

- `TestFSharpTopLevelPipelineCallsResolve` — statement-position pipeline in an `.fsx`.
- `TestFSharpQualifiedCallsStayInTheNamedModule` — asserts on the resolved definition's **line number**, so picking the other module fails rather than passing on a name match.

### Mutation

Each hunk reverted separately; each mutation compiles and was grep-confirmed applied.

| revert | result |
|---|---|
| `fsharpCallTargets(topLevel)` → `fsharpDottedCallTargets(topLevel)` | `--- FAIL: TestFSharpTopLevelPipelineCallsResolve` — `missing top-level F# pipeline CALLS script.fsx->normalize: []` (the qualified test still passes) |
| `fsharpQualifiedScope` predicate short-circuited to keep every symbol | `--- FAIL: TestFSharpQualifiedCallsStayInTheNamedModule` — `apply (line 9) resolved A.convert to the definition on line 5, want line 2` (the top-level test still passes) |

Restored → all six `TestFSharp*` pass.

(A first attempt at the second mutation — flipping `cut > 0` to `cut < 0` — panicked on `target[:-1]` rather than failing the assertion, so it was discarded and replaced with one that exercises the guard.)

### main vs branch

A seven-file F# corpus (the Paket-shaped cross-file fixture, two modules with a shared name, pipelines, an `.fsx`) indexed on both trees. Nothing main emits is lost:

```
only on main:    parens -> convert@6   (B)   \
                 apply  -> convert@6   (B)    > the three wrong targets, replaced
                 Caller -> convert@6   (B)   /  by the correct convert@2
only on branch:  14 edges — the corrected three, plus the pipeline and
                 top-level edges this PR is for
```

### Golden delta

**None.** No F# golden fixture exists; `TestProviderGoldenSnapshots`, `TestProviderGoldenFixtureQualityCoverageReport`, `TestCapabilityMatrixCoversEmittedRelations` and `TestImportCapabilityMatchesScannerRegistry` pass unmodified with `testdata/` clean.

### Also checked

LF and CRLF produce identical CALLS sets. `TestLua*`, `TestZig*`, `TestOCaml*`, `TestHaskell*`, `TestJulia*` pass (the shared `dottedCallIdentifiers` helper is untouched). `gofmt`/`go vet` clean.

### On the overlap with #178

Still real, and now slightly larger: both branches touch the F# block in `provider.go`. They are independent fixes to different scanners, so the second to land needs a merge of that block rather than a rewrite. Happy to sequence them in either order.

suhaanthayyil:

- TestFSharp, plus `TestProviderGoldenSnapshots` — pass on the merged tree

PR #164, opened 2026-08-28
fix(cli): bound the instruction files init-agents reads

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/123

`init-agents` reads the target repository's `AGENTS.md` and `CLAUDE.md` with a plain
`io.ReadAll` and no size bound. Their paths are fixed, but their **size is chosen by the
repository**, and `entire graph init-agents` is run against clones the operator has not read.

Every sibling repository read in this tool is already bounded — source files and diffs at
`defaultMaxParseBytes` (4 MiB), ignore files at `maxIgnoreFileBytes` (1 MiB), call-site
snippets at `callSiteMaxFileBytes` (4 MiB), the SCIP project version at
`ScipProjectVersionMaxLen`. This read was the one that was not.

## Measured on unpatched `main`

Empty repository, one oversized `AGENTS.md`, nothing else. Peak RSS from
`/usr/bin/time -l`:

| `AGENTS.md` | wall clock | peak RSS | RSS / file |
|---|---|---|---|
| 128 MiB | 0.13 s | 287 MiB | 2.2x |
| 512 MiB | 0.60 s | 1,756 MiB | 3.4x |
| 1,024 MiB | 1.17 s | 2,112 MiB | 2.1x |
| 2,048 MiB | 3.32 s | **6,523 MiB** | 3.2x |

```
$ /usr/bin/time -l entire-graph init-agents --repo <repo-with-2GiB-AGENTS.md>
wrote .../.entire/graph-agent.md
updated .../AGENTS.md
        3.32 real
    6839156736  maximum resident set size
```

The multiplier is not one copy. `readAndValidateInstructionFile` returns the whole file, it is
held live through `validatePointerMarkers`, `claudeDirectlyImportsAgents` (which additionally
does `strings.Split(string(content), "\n")` on it), and `renderPointerBlock` (which builds a
second full string), and `AGENTS.md` and `CLAUDE.md` are both held at once because every byte
that depends on an instruction read is computed before anything is written. So the working set
runs to a multiple of the largest instruction file, and a repository picks that number.

## The fix

`readContainedFile` takes a byte limit and reads through `io.LimitReader`.
`readAndValidateInstructionFile` applies `maxInstructionFileBytes` (4 MiB, matching
`defaultMaxParseBytes` — the ceiling the rest of the tool already applies to a
repository-authored text file, and three orders of magnitude above any real instruction file;
this repository's own is ~2 KiB).

Two deliberate choices:

- **Refuse, do not truncate.** A truncated instruction file would be written back over the
  user's own text under the managed block. A refusal costs the user a rerun; a truncation costs
  them their file.
- **Bound the read, not a preceding `Stat`.** `readContainedFile` documents why it reads
  through the `os.Root` rather than the absolute path: the preflight is explicitly not atomic
  with what follows. A `Stat`-based gate has the same hole — a file that grows between the check
  and the read defeats it, and a growing file is exactly what the limit is for. One byte past
  the limit is requested so a file sitting exactly on the limit is still accepted.

The refusal lands in the preflight, before `mkdirAllContained` and before any write, so the
repository is untouched.

## Tests

`TestInitAgentsRefusesOversizeInstructionFileWithoutWrites` — `AGENTS.md` and `CLAUDE.md` at
`maxInstructionFileBytes+1`: the command errors, stdout is empty, the oversized file is
byte-identical afterwards, and neither `.entire/graph-agent.md` nor the counterpart file is
created.

`TestInitAgentsAcceptsInstructionFileExactlyAtTheLimit` — a file exactly at the limit is still
installed into, pinning the bound off-by-one-safe.

Both fail on unpatched `main`. Verified by mutation — deleting the
`len(content) > limit` check from `readContainedFile` (restoring the pre-fix read):

```
--- FAIL: TestInitAgentsRefusesOversizeInstructionFileWithoutWrites/AGENTS.md
    init-agents accepted a 4194305-byte AGENTS.md
--- FAIL: TestInitAgentsRefusesOversizeInstructionFileWithoutWrites/CLAUDE.md
    init-agents accepted a 4194305-byte CLAUDE.md
```

## Gate

`gofmt -s -l .` empty · `go build ./...` clean · `go vet ./...` clean ·
`go test -race -run TestInitAgents ./internal/cli/` green (130 tests) ·
`go test -timeout 30m ./internal/cli/` green, zero failures.

`mise run check` was not run to completion here, and a local timeout would not have been
evidence either way. `internal/cli` needs about 13 minutes on this machine, past Go's default
10-minute per-package timeout. Measured back to back, same flags:

| worktree | `go test -timeout 30m ./internal/cli/` |
|---|---|
| unmodified `origin/main` | ok, 792.5 s |
| this branch | ok, 819.1 s |

Both pass; both would `panic: test timed out` at the default. CI runs the full gate.


ashtom:

**Verdict: clean — this is the standard the other bounds PRs should be held to**

- Live PR head `c22b8460`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `c22b8460` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking. Calling out three things this PR does that I check for and usually do not find:

**1. The amplification is measured, not asserted.** "a 2 GiB `AGENTS.md` drove init-agents to
6.5 GiB peak RSS" is a number, and 4 MiB is justified by matching `defaultMaxParseBytes` rather than
picked.

**2. Every entry point is bounded, not just one.** `readContainedFile` gained a required `limit`
parameter rather than an internal default, so the compiler enforces that no caller can skip it.
There is exactly one call site and it passes the constant:

```
$ git grep -n 'readContainedFile(' refs/prs/164 -- internal/cli
internal/cli/agents.go:1277:func readContainedFile(root *os.Root, name string, limit int64) ...
internal/cli/agents.go:1390:  content, err := readContainedFile(root, name, maxInstructionFileBytes)
```

**3. The guard is tested in both directions.** `TestInitAgentsRefusesOversizeInstructionFileWithoutWrites`
(limit+1 refused, and refused *before any write*) and
`TestInitAgentsAcceptsInstructionFileExactlyAtTheLimit` (exactly-at-limit still installed). Refusing
rather than truncating is also the right call — a truncated instruction file would be written back
over the user's own text.

See the merge-order note on **#158**: that PR conflicts with this one in `agents.go`.

### Independent review state

suhaanthayyil:
Thanks. Staleness unchanged — `main` still `bb41f223`, still `13 behind / 1 ahead`, and the `readContainedFile` call-site count still reproduces exactly as quoted (one definition, one call site, passing the constant).

**The merge-order note is stale.** #158 has moved since the review — its head is now `b8c5dbb4`, and it gained a hard-link guard that touches `agents.go` and `writeContainedFile`. Against that head, `agents.go` is *not* the conflict:

```
$ git merge-tree --write-tree refs/prs/164 refs/prs/158
Auto-merging internal/cli/agents.go            <- clean
Auto-merging internal/cli/agents_test.go
CONFLICT (content): Merge conflict in internal/cli/agents_test.go
```

`internal/cli/agents.go` auto-merges with zero conflict markers. The single conflicted file is `internal/cli/agents_test.go`.

**Scope of that conflict:** two hunks, both append/append. Each side adds whole new test functions; git anchored the regions on the shared closing-brace tail (`}` / `}` / `})` / `}` / `}`) that both new functions end with, which is why the markers land mid-function and it reads worse than it is. The resolution is keep-both, with each side's body closed by that shared tail.

I did the resolution and ran it rather than asserting it:

```
$ gofmt -l internal/cli/            (no output)
$ go vet ./internal/cli/            (no output)
$ go test ./internal/cli/ -run 'TestInitAgents' -count=1
ok      github.com/entireio/entire-graph/internal/cli   0.772s
```

52 top-level `TestInitAgents*` functions on the merged tree, with both sides' new tests passing by name:

```
--- PASS: TestInitAgentsRefusesOversizeInstructionFileWithoutWrites
--- PASS: TestInitAgentsAcceptsInstructionFileExactlyAtTheLimit
--- PASS: TestInitAgentsRefusesManagedTargetInsideGitDirectory
--- PASS: TestInitAgentsRefusesManagedTargetHardLinkedIntoGitDirectory
```

**The bound survives the merge intact**, which is the thing actually worth checking when a second PR edits the same file. #158 introduces no new read path — its only added file-opening line in `agents.go` is a `root.OpenFile(..., O_WRONLY|O_CREATE, perm)` on the write side:

```
$ git diff <base> refs/prs/158 -- internal/cli/agents.go | grep '^+.*\(os.ReadFile\|io.ReadAll\|readContainedFile\|root.Open\)'
+       file, err := root.OpenFile(resolved, os.O_WRONLY|os.O_CREATE, perm)
```

So on the merged tree `readContainedFile` still has exactly one call site, still passing `maxInstructionFileBytes`, and the compiler-enforced `limit` parameter still means no future caller can skip it. There is no unbounded read reachable after both land.

ashtom:

no longer does.

```
01M13A3JEDF7  Medium  open    current    internal/cli/agents.go:1390  A file at or just below this limit is accepted, but `renderPointerBlock` then ap…
```

### Verified vs. assumed

suhaanthayyil:
## Finding response — `internal/cli/agents.go:1390`

**VALID.** Head is now `fc240370`.

### What was wrong

`readContainedFile` accepts a file sitting exactly on the limit (`if int64(len(content)) > limit` — the limit is the largest accepted size). `renderPointerBlock` then **appends** the managed block to that snapshot, and the appended result is what gets written. So a 4 MiB `AGENTS.md` was read, grown by the block, and written back past the bound the read enforces.

This PR's own boundary test was pinning the wrong side of it:

```go
// TestInitAgentsAcceptsInstructionFileExactlyAtTheLimit
atLimit := bytes.Repeat([]byte("a"), maxInstructionFileBytes)
… asserts init-agents succeeds and appends the block
```

It asserted the file was installed into, and never looked at the size of the result. The user-visible consequence is that init-agents wedges its own output: the next run reads the file it just wrote, finds it over the limit, and refuses with *"file is larger than the … limit … reduce it"* — over bytes init-agents itself added, which the user cannot remove by shrinking their own text.

### Fix

The bound now holds for what is **written**, not only for what was read. `ensureRenderedInstructionFits` is called for both instruction files after every byte is computed and before the first write, so a refusal leaves the tree exactly as it was — the same placement as the existing oversize preflight.

```
internal/cli/agents.go   ensureRenderedInstructionFits(agentsPath, agentsSource, agentsContent)
internal/cli/agents.go   ensureRenderedInstructionFits(claudePath, claudeSource, claudeContent)
```

The message reports both sizes, because the difference between them is the block this command adds and the part the user cannot shrink.

### Tests

`internal/cli/agents_test.go`:

- `TestInitAgentsRefusesWhenTheManagedBlockWouldCrossTheReadLimit` — a source exactly on the limit must be refused, with `AGENTS.md` byte-identical afterwards and neither `.entire/graph-agent.md` nor `CLAUDE.md` created.
- `TestInitAgentsAcceptsTheLargestFileThatStillFitsOnceTheBlockIsAdded` — replaces the old boundary test with the correct boundary: the largest accepted source is the one whose **rendered** form lands exactly on the limit; the result must measure exactly `maxInstructionFileBytes`, and a **second run must succeed and leave it unchanged**. That round trip is the property bounding the write buys, and it is the thing that was broken.

The block overhead the second test needs is *measured* from a probe run (`managedBlockOverheadForTest`) rather than restated, so the test cannot drift from the block the command actually renders.

### Mutation verification

Removed both `ensureRenderedInstructionFits` calls, `grep`-confirmed only the definition and its doc comment remained (2 hits, no call sites), rebuilt — the now-unused function still compiles:

```
--- FAIL: TestInitAgentsRefusesWhenTheManagedBlockWouldCrossTheReadLimit (0.01s)
    agents_test.go:2404: init-agents wrote a file past the 4194304-byte limit it will read back
```

Restored, `grep`-confirmed 4 hits back, all `TestInitAgents*` pass.

### Line endings

This change is about file sizes and appended text, so both were checked. The tests generate every byte in Go (`bytes.Repeat`, `os.WriteFile`, `os.ReadFile`) and never read a fixture from the working tree, and the overhead is measured at runtime rather than hard-coded, so a CRLF checkout cannot shift the boundary. Verified by converting `agents.go` and `agents_test.go` to CRLF in place and re-running:

```
internal/cli/agents.go:      … with CRLF line terminators
internal/cli/agents_test.go: … with CRLF line terminators
ok  	github.com/entireio/entire-graph/internal/cli	0.702s
```

Full `go test ./internal/cli` — 753 passed. `go vet` and `gofmt -l` clean.

PR #165, opened 2026-08-28
fix(sem): stop inventory-only files emitting call relations

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/124

## Impact

Inventory-only filetypes emitted **CALLS edges into other languages' symbols**.

`docs/language-support.md` states that inventory-only coverage is file discovery
and document symbols "without claiming call/type/data-flow analysis", and
`capabilities --json` declares only `CONTAINS`/`DEFINES` for these languages.
The provider did not honour that: an inventory-only file gets one `document`
symbol whose block is the entire file, and the generic `name(` call scanner ran
over it like any other symbol body. Names that hit the globally-unique-name
fallback produced confident-looking, wholly invented edges.

Observed on a fixture repository covering every registered extension:

| from | to |
|---|---|
| `notes/change.patch` (a unified diff) | Go `func Run` |
| `sketch/blink.ino` | Objective-C `addValues` |
| `web/app.coffee` | Scala `Point.sum` |
| `fix.idr`, `fix.lean`, `fix.re`, `fix.res` | Haskell `pointSum` |
| `fix.raku`, `fix.wren` | Rust `Point::new` |
| `fix.mm` | Objective-C `addValues` |

Eleven inventory-only languages produced CALLS in a single fixture pass. These
edges reach `entire graph neighbors`, `entire graph impact` (blast radius) and
graph-neighbour search ranking, all of which present them as real. `.patch` and
`.diff` files are common in ordinary repositories, so is a vendored `.coffee`
tree, so this is not a synthetic condition.

## Repro

```
repo/
  svc/service.go     func RunPipeline() int { return 1 }
  sketch/blink.ino   void loop() { RunPipeline(); }
```

```
$ entire graph edges --repo . --format ndjson --relation CALLS
... "from_id":"...:Arduino:sketch/blink.ino:document:blink",
    "to_id":"...:Go:svc/service.go:function:RunPipeline",
    "type":"CALLS","confidence":0.68,
    "reason":"direct call expression matched globally unique symbol name"
```

Arduino is declared `["CONTAINS","DEFINES"]` by `capabilities --json`.

## Root cause

Two independent decisions about the same question.

- The **capability report** answers "does this language have call extraction?"
  with `supportsCallExtraction`, a hard-coded language list.
- The **scanner** answered it with
  `fileNeedsCallScan := needsCallScan && !skipFastCFamilyCallScan(...)`, which
  consults only the profile and a C-family fast-path exception — no language
  tier at all.

So the scan ran on every file that produced any symbol.
`TestCapabilityMatrixCoversEmittedRelations` is the guard for exactly this, but
it is fixture-driven, and no golden fixture pairs an inventory-only file with a
uniquely-named symbol elsewhere in the repo, so nothing ever tripped it.

## Fix

Make the two agree by construction. `supportsCallExtraction`'s language list is
lifted into `callExtractionLanguage(language string)`, and both CALLS producers
gate on it:

- `fileNeedsCallScan` (the bare-name scan and the top-level complement), and
- the `receiverCallRelations` / `importedReceiverCallRelations` pass, which is a
  separate producer and accounted for the last three false edges
  (`.coffee` -> Scala `Point.sum`, `.raku`/`.wren` -> Rust `Point::new`).

The capability report and the scanners now read the same predicate, so a
language the report says has no CALLS is a language the scanner does not run on.
No language in the list changes behaviour.

Measured on the fixture repository (uncached, `--worktree`): CALLS drops from 86
to 75 — exactly the eleven inventory-only false edges — and no semantic-language
CALLS edge is lost.

## Regression test

`TestInventoryOnlyFilesEmitNoCallRelations` builds a repo with a Go
`RunPipeline` plus a `.patch`, an `.ino` and a `.coffee` file that each mention
`RunPipeline()`, and fails if any `CALLS`/`CONSTRUCTS`/`ASYNC_CALLS` relation
originates from an inventory-only symbol. It also asserts the Go symbol is still
extracted, so the test cannot pass by the snapshot coming out empty.

Mutation verification (each gate is bound independently):

- Before the fix: FAIL — `inventory-only Arduino symbol emitted CALLS: ...:Arduino:sketch/blink.ino:document:blink -> ...:Go:svc/service.go:function:RunPipeline`.
- Removing `callExtractionLanguage` from `fileNeedsCallScan`: FAIL on the Arduino edge.
- Removing it from the `receiverCallRelations` gate: FAIL on the CoffeeScript edge.
- With the fix: PASS.

## Verification run locally

full gate on a clean runner.

### DIFF
--- a/internal/sem/provider.go
+++ b/internal/sem/provider.go
@@ -724,7 +724,25 @@ func supportsCallExtraction(spec languageSpec) bool {
 	if spec.inventoryOnly {
 		return false
 	}
-	switch spec.language {
+	return callExtractionLanguage(spec.language)
+}
+
+// callExtractionLanguage is the single source of truth for "this language's
+// call syntax is extracted". It backs BOTH the capability report and the
+// runtime gate on the call scan (fileNeedsCallScan), so the two cannot drift:
+// a language the report says has no CALLS is a language the scanner does not
+// run on.
+//
+// They used to be independent. The report was this list; the scanner ran on
+// every file with any symbol, including the single `document` symbol an
+// inventory-only filetype gets, whose block is the whole file. A `.patch`, a
+// `.coffee` or a `.ino` therefore emitted CALLS into unrelated languages'
+// symbols through the globally-unique-name fallback — false edges in
+// `neighbors`, `impact` and search ranking, from filetypes
+// docs/language-support.md promises are "without claiming call/type/data-flow
+// analysis".
+func callExtractionLanguage(language string) bool {
+	switch language {
 	case "Bash", "C", "C++", "C#", "Clojure", "ClojureScript", "Dart", "Elixir", "Erlang", "F#", "Go", "Groovy", "Haskell", "Java", "JavaScript", "Julia", "Kotlin", "Lua", "Objective-C", "OCaml", "Perl", "PHP", "Python", "R", "Ruby", "Rust", "Scala", "SQL", "Swift", "TypeScript", "Zig", "Zsh":
 		return true
 	default:
@@ -3064,7 +3082,11 @@ func forEachRelation(repoKey string, files []FileRecord, recordsByFile map[strin
 		}
 		lines := strings.Split(content, "\n")
 		currentFileSymbols := recordsByFile[file.Path]
-		fileNeedsCallScan := needsCallScan && !skipFastCFamilyCallScan(spec, file.Language)
+		// callExtractionLanguage keeps the scan inside the set `capabilities
+		// --json` declares CALLS for. Without it the scan also ran over
+		// inventory-only `document` symbols (whose block is the whole file) and
+		// over data/markup languages, inventing cross-language call edges.
+		fileNeedsCallScan := needsCallScan && callExtractionLanguage(file.Language) && !skipFastCFamilyCallScan(spec, file.Language)
 		fileNeedsRouteScan := spec.emits("HANDLES_ROUTE") && routeScanLanguage(file.Language)
 		fileNeedsHTTPScan := spec.emits("HTTP_CALLS") && httpScanLanguage(file.Language)
 		fileNeedsServiceScan := needsServiceRelations && serviceScanLanguage(file.Language)
@@ -3642,7 +3664,12 @@ func forEachRelation(repoKey string, files []FileRecord, recordsByFile map[strin
 					}
 				}
 			}
-			if spec.callResolution == "full" {
+			// The receiver pass resolves `recv.method(...)` and is a second
+			// producer of CALLS, so it takes the same language gate as the
+			// bare-name scan above: without it an inventory-only `document`
+			// symbol still reached it and matched a `x.y(...)` shape anywhere
+			// in the file against another language's methods.
+			if spec.callResolution == "full" && callExtractionLanguage(file.Language) {
 				for _, r := range receiverCallRelations(from, block, methodsByContainer, superContainerByID, implementersByContainer, symbolsByShortName, returnTypesBySymbolNameAndFile, returnTypesBySymbolNameAndDir, importsByName, manifestImports.goModule, pkgVarTypesByDir[filepath.ToSlash(filepath.Dir(file.Path))], phpPropTypes, kotlinPropTypes, typeScriptPropTypes, fieldsByContainer, swiftTypes) {
 					emit(r)
 				}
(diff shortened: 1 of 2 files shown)


ashtom:

**Verdict: clean, with a merge-order note**

- Live PR head `978f29ad`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `978f29ad` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking in the change itself.

**1. (merge order) Conflicts with #168.** Both edit the capability tables in
`internal/sem/provider.go`:

```
$ git merge-tree --write-tree refs/prs/165 refs/prs/168   ->  CONFLICT
```

Each is clean against `main` alone. Whichever lands second needs a manual resolution, and the
resolution has to preserve *both* sets of declarations rather than taking one side wholesale —
see my comment on #168 and #181, where taking one side wholesale is exactly the failure.

### Independent review state

suhaanthayyil:
Thanks. Staleness unchanged — `main` still `bb41f223`, still `13 behind / 1 ahead`. The conflict with #168 is real, but the characterisation of it is not.

**"Both edit the capability tables in `internal/sem/provider.go`" is not true of this pair.** The two PRs' `provider.go` diffs are disjoint:

```
$ git diff <base> refs/prs/165 -- internal/sem/provider.go | grep -c ooRelationSupport
0
$ git diff <base> refs/prs/168 -- internal/sem/provider.go | grep -cE 'supportsCallExtraction|fileNeedsCallScan|callResolution'
0
```

#168 edits only the `ooRelationSupport` table (L112–151). This PR edits only the call-extraction gate (`supportsCallExtraction`/`callExtractionLanguage` at L724, and the two emission sites at L3082 and L3664). Roughly 600 lines apart, no overlap. So `provider.go` auto-merges with **zero** conflict markers, and the merged file provably carries both:

```
$ grep -c 'func callExtractionLanguage' internal/sem/provider.go   -> 1
$ grep -n '"Clojure":\|"Dart":\|"SQL":\|"Julia":' internal/sem/provider.go
138: "Clojure": {"USES_TYPE"}
139: "Dart":    {"USES_TYPE","PARAM_TYPE","RETURNS_TYPE","ASYNC_CALLS","DATA_FLOWS"}
145: "Julia":   {"USES_TYPE","PARAM_TYPE","DATA_FLOWS"}
151: "SQL":     {"DATA_FLOWS"}
```

**The "take one side wholesale" hazard does not apply here.** That hazard is a capability-table collision — two PRs rewriting the same map entries. Neither file in this pair has that shape: both `provider_test.go` diffs are pure additions with zero deleted lines (`grep -c '^-[^-]'` returns 0 on both), so the only conflicted file is append/append — each side adds one new test function immediately before `TestResourceDependsOnGraph` and git interleaved them on the shared `BuildProviderSnapshot` boilerplate.

Worth noting the #181/#168 pair the warning points at is also no longer a conflict against current heads:

```
$ git merge-tree --write-tree refs/prs/181 refs/prs/168   ->  exit 0, clean
```

#181's head `b0f15e4c` now carries the identical table entries, so git merges them as the same change.

**Verified the resolution rather than describing it.** Spliced both test functions into the merged tree:

```
$ gofmt -l internal/sem/     (no output)
$ go vet ./internal/sem/     (no output)
$ go test ./internal/sem/ -run 'TestInventoryOnlyFilesEmitNoCallRelations|TestCapabilityMatrixDeclaresTypeAndFlowRelations|TestCapabilityMatrixCoversEmittedRelations|TestCapabilityRelationDeclarationsMatchIsolatedExtraction' -count=1
--- PASS: TestCapabilityMatrixCoversEmittedRelations (0.50s)
--- PASS: TestInventoryOnlyFilesEmitNoCallRelations (0.02s)
--- PASS: TestCapabilityMatrixDeclaresTypeAndFlowRelations (0.03s)
--- PASS: TestCapabilityRelationDeclarationsMatchIsolatedExtraction (0.00s)   [13 subtests]
ok
```

The last two are the ones that matter for this combination: #168 declares thirteen languages' type and data-flow relations, and this PR newly *gates* emission by language. If the gate silenced anything #168 declares, `TestCapabilityRelationDeclarationsMatchIsolatedExtraction` (one language per repository, probed in isolation) is exactly the test that would catch declared-but-not-emitted. All 13 subtests pass, so the two changes are compatible in substance and not only in text.

Also merged this PR with #171's language matrix: 36 language subtests green, so the gate does not silence any semantic language.

PR #166, opened 2026-08-28
fix(sem,gitutil): classify symlink tree entries instead of parsing them

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/125

## Problem

Git stores a symbolic link as an ordinary blob whose bytes are its target path. `entire graph diff` reads changed-file content by blob and never looks at the tree entry mode, so a symlink arrives indistinguishable from a one-line source file and is parsed as one.

```
$ ln -s pkg/real.py alias.py && ln -s pkg/real.go alias.go && git add -A && git commit -m links
$ entire graph diff --base HEAD~1 --head HEAD --json
```

Before this change:

- `alias.py` — the blob is `pkg/real.py`, which is a valid Python expression, so the file parses and the diff reports a phantom entity:
  `{"type":"added","kind":"module","name":"alias.py","after_start_line":1}`
- `alias.go` — the blob is `pkg/real.go`, which Go cannot parse, so the diff reports a parse failure against a file that is not source at all:
  `E_PARSE_ERROR alias.go — tree-sitter syntax error nodes present: error ERROR at line 1 column 1 near "pkg/real.go"`

Neither is a real change, and `E_PARSE_ERROR` also downgrades the reported completeness of the range.

The gitlink case is already classified correctly (`LimitedFileNonBlob` → "the Git tree entry has no blob content"), but a symlink *is* a blob, so nothing downstream of the content read can separate it from source. Mode is the only signal, and `ChangedFiles` was not carrying it.

## Fix

`gitutil.ChangedFiles` asks Git for `--raw` instead of `--name-status`. The raw record is `:<oldmode> <newmode> <oldsha> <newsha> <status>` followed by the pathname(s), so the modes come back in the same subprocess — no extra Git call. `ChangedFile` gains `OldMode` / `NewMode`, and malformed raw metadata is now an error rather than a silent skip.

`AnalyzeGitRangeWithOptions` classifies a symlink on either side before reading content and emits a machine-readable completeness marker, mirroring the gitlink handling it sits next to:

```
! W_UNSUPPORTED_FILE alias.go
  file skipped; the Git tree entry is a symbolic link, so its changes are not analyzed
  head version is a symbolic link, not file content
```

Side-specific details (`base version` / `head version` / `base and head versions`) follow the same shape as the neighbouring markers, so a `T` typechange in either direction names the correct side.

A file converted into a symlink is reported as a skip with a marker rather than as the removal of its entities. That matches the existing behaviour for a file converted into a gitlink, and keeps the "unparseable input is surfaced, never silently dropped" contract.

## Tests

- `TestChangedFilesReportsTreeEntryModes` (gitutil) — added symlink, deleted file, and edited file each carry the right `OldMode` / `NewMode`.
- `TestAnalyzeGitRangeSkipsSymlinkTreeEntries` — added `alias.py` / `alias.go` produce no entity deltas, no `E_PARSE_ERROR`, and a symlink marker for each.
- `TestAnalyzeGitRangeSkipsDeletedSymlink` — the base side is classified too.
- `TestAnalyzeGitRangeStillAnalyzesRegularFiles` — an ordinary blob still produces its delta.

Mutation-checked: disabling the classification fails the first two analyze tests; widening it to every entry fails the regular-file test; dropping `NewMode` from the parser fails the gitutil test.

---

**Local gate**

- `go test -timeout 30m ./...` — green, every package `ok` (internal/cli 898s, internal/sem 926s).


ashtom:

**Verdict: changes required — the fix is silently neutralised once merged onto current `main`**

- Live PR head `fe738158`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `fe738158` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|
| `git merge origin/main` | PR head → main | **conflict** in `internal/sem/analyze_test.go` (resolved: keep both sides) |
| PR's own tests | merged onto main | **2 of 3 FAIL** — fix neutralised |
| PR's own tests + 3-line mode fix | merged onto main | pass |
| `go test ./internal/sem/` | merged onto main + fix | pass (433s) |
| `core.symlinks=false`, tree-entry-only symlink | merged onto main + fix | pass — mode-based classification holds |

### Findings

**1. (HIGH, blocking) This PR is green on its own base and red on current `main`. Its two central
tests fail after the merge.**

This is not a flake and not a conflict artifact. I merged the branch onto `bb41f223`, resolved the one
conflict in `internal/sem/analyze_test.go` (a pure adjacent-addition — both sides appended new test
functions; I kept both), and ran the PR's own tests:

```
--- FAIL: TestAnalyzeGitRangeSkipsSymlinkTreeEntries
    symlink tree entries produced entity deltas:
      FileChange{Path:"alias.py", Status:"A", Language:"Python",
                 Changes:[EntityChange{Type:"added", Kind:"module", Name:"alias.py"}]}
--- FAIL: TestAnalyzeGitRangeSkipsDeletedSymlink
    deleted symlink produced entity deltas: ... Status:"D" ...
--- PASS: TestAnalyzeGitRangeStillAnalyzesRegularFiles
```

**Root cause.** `admitChangedFiles` — added to `internal/sem/analyze.go` by **#167**, which landed on
`main` *after* this branch was cut — rewrites the changed-file list and **constructs fresh
`gitutil.ChangedFile` values in three of its branches, dropping the `OldMode`/`NewMode` fields this
PR just added**:

```go
kept = append(kept, gitutil.ChangedFile{Status: "A", Path: file.Path})   // copy
kept = append(kept, gitutil.ChangedFile{Status: "D", Path: oldPath})     // base-only
kept = append(kept, gitutil.ChangedFile{Status: "A", Path: file.Path})   // head-only
```

Every *added* and every *deleted* path therefore reaches your guard with `NewMode`/`OldMode` == `""`,
so `beforeSymlink`/`afterSymlink` are never true. Only the `inBase && inHead` case keeps `file`
intact, which is why a *modified* symlink still classifies and an added or deleted one does not.

I confirmed the plumbing itself is fine — `gitutil.ChangedFiles` reports the mode correctly:

```
diff --raw -z: ":000000 120000 0000000 3eecbec A\x00alias.py\x00"
ChangedFile: status="A" path="alias.py" oldMode="000000" newMode="120000"
```

**Verified fix — three lines, in `admitChangedFiles`:**

```go
case inBase:
    kept = append(kept, gitutil.ChangedFile{Status: "D", Path: oldPath, OldMode: file.OldMode})
case inHead:
    kept = append(kept, gitutil.ChangedFile{Status: "A", Path: file.Path, NewMode: file.NewMode})
// and in the "C" branch:
    kept = append(kept, gitutil.ChangedFile{Status: "A", Path: file.Path, NewMode: file.NewMode})
```

With that applied, all three of this PR's tests pass and the whole `internal/sem` package is green
on the merged tree (`ok internal/sem 433.574s`).

**2. (good) The design is at the right layer, and I proved it.** §7 of the protocol warns that
symlink tests skip on Windows and that `core.symlinks=false` makes a tracked symlink look like a
regular file — invisible on any machine that *can* create symlinks. Your fix classifies on the **Git
tree mode** (`120000`), not on an `lstat`, so it is immune to that. I verified this rather than
assuming it, by building the symlink as a tree entry only — `git update-index --cacheinfo
120000,<blob>,alias.py` with `core.symlinks=false` and no filesystem link anywhere:

```
--- PASS: TestReviewSymlinkNoFilesystemLink
    OK: mode-based classification holds with no on-disk symlink
```

That case passes *with* the mode fix above and fails without it. **Consider adding it as a test** —
it is the one symlink case that runs on every platform, including the Windows leg where your two
current tests `t.Skip`.

**3. (scope, §6) `ChangedFiles` switching from `--name-status` to `--raw` is a shared-behaviour
change** affecting every caller, not only the diff path. The new parser is stricter (it errors on
malformed metadata where the old one skipped). I did not find a caller this breaks, and the full
`internal/sem` and `internal/gitutil` suites pass, but it is worth a reviewer's eye as a
blast-radius question rather than a symlink question.

**Merge conflict resolved.** For the record, the resolution is: take `main`'s
`internal/sem/analyze_test.go` and append this branch's three new test functions
(`TestAnalyzeGitRangeSkipsSymlinkTreeEntries`, `TestAnalyzeGitRangeSkipsDeletedSymlink`,
`TestAnalyzeGitRangeStillAnalyzesRegularFiles`). Both sides' tests are kept; `gofmt` is clean.
I have **not** pushed this to your branch.

### Independent review state

suhaanthayyil:
## Finding 1 (HIGH): reproduced exactly, and fixed

Confirmed — not disputed on any point. Merged `fe738158` onto `bb41f223`, resolved the one `internal/sem/analyze_test.go` conflict as described (keep main's file, re-append this branch's three tests), and ran the PR's own tests before changing anything:

```
--- FAIL: TestAnalyzeGitRangeSkipsSymlinkTreeEntries (0.13s)
--- FAIL: TestAnalyzeGitRangeSkipsDeletedSymlink (0.13s)
--- PASS: TestAnalyzeGitRangeStillAnalyzesRegularFiles (0.22s)
FAIL	github.com/entireio/entire-graph/internal/sem	1.100s
```

2 of 3, exactly as reported. The root cause is as diagnosed: `admitChangedFiles` runs at `internal/sem/analyze.go:142`, before the symlink classification at ~`:196`, and rebuilds the entry in three branches instead of forwarding it, naming only `Status` and `Path`. `grep -rn 'ChangedFile{' internal --include='*.go' | grep -v _test` returns exactly those three sites plus the two in the `--raw` parser, so the enumeration is complete.

**Fix** — `internal/sem/analyze.go:1103` (copy), `:1111` (`case inBase:`), `:1113` (`case inHead:`), your three lines:

```go
case inBase:
    kept = append(kept, gitutil.ChangedFile{Status: "D", Path: oldPath, OldMode: file.OldMode})
case inHead:
    kept = append(kept, gitutil.ChangedFile{Status: "A", Path: file.Path, NewMode: file.NewMode})
```

I kept the explicit field list rather than copying `file` and overriding: the rewrite deliberately drops `OldPath`, and a struct copy would leave it set on an entry now labelled `D`. I added a paragraph to the `admitChangedFiles` doc comment saying the mode has to travel with the rewrite and why, so the next field added to `ChangedFile` does not repeat this silently.

## Finding 2: mode-only test added

`TestAnalyzeGitRangeClassifiesSymlinkFromTreeModeAlone`, in `internal/sem/analyze_test.go`. Builds the symlink as a tree entry only — `git update-index --add --cacheinfo 120000,<blob>,<path>` under `core.symlinks=false`, no filesystem link anywhere — so mode is the only signal and it runs on Windows, where the two existing tests `t.Skip`. It asserts the fixture is really a `120000` entry of the expected status first, so it cannot pass for the wrong reason.

One correction to the shape you suggested. My first attempt put the added and deleted symlink in a **single range** and it failed even with the fix applied:

```
analyze_test.go:2219: no symlink completeness marker for gone.py: ...
  Detail:"base and head versions are symbolic links, not file content"
```

Both links pointed at the same target, so their blobs were identical and `--find-renames` paired them into one `R` — which takes the `inBase && inHead` branch that forwards `file` intact and therefore exercises **neither** rewrite. The test now measures addition and deletion over separate ranges, and that is documented in the test comment so it does not get "simplified" back.

The copy branch is unreachable end-to-end (`ChangedFiles` passes `--find-renames`, not `--find-copies-harder`), so I pinned it in the existing `TestAdmitChangedFilesRewritesIndexCrossings` table instead — three new cases carrying modes through the D, A and C rewrites. That table compares whole structs with `!=`, so the mode fields are genuinely asserted.

## Mutation verification

Each of the three lines was reverted independently. Every mutation compiled (`go build ./internal/sem/` before each run):

| mutation | result |
|---|---|
| revert `case inBase:` only | `/deleted` **FAIL**, `/added` PASS |
| revert `case inHead:` only | `/deleted` PASS, `/added` **FAIL** |
| revert both | both subtests **FAIL** |
| revert copy branch only | `TestAdmitChangedFilesRewritesIndexCrossings/copy_into_the_index_carries_the_destination_mode` **FAIL**, all others PASS |
| restored | all PASS |

Each mutation fails exactly the case that covers it and nothing else, so no line of the fix is unpinned and no test is passing incidentally.

## Finding 3: blast radius bounded

Narrower than it looks. `gitutil.ChangedFiles` has exactly two callers, both in `internal/sem/analyze.go` (`:114` and `:136`, the second for index-policy pathspecs) — no caller outside the diff path, so the `--name-status`→`--raw` switch cannot reach anything else. `internal/gitutil` is green (109 tests). The stricter parser is a deliberate change: `--raw` metadata that does not start with `:` or has fewer than 5 fields means git's output contract was violated, and silently skipping it would drop a real change from the delta.

## State

Merge + fix pushed to this branch; PR is now `MERGEABLE` against `bb41f223`. Green on the merged tree: `internal/gitutil` 109 passed, and the symlink plus index-policy tests 34 passed. I did not run the full `internal/sem` sweep locally — CI now has a mergeable head to run it against, which is the thing that was actually missing.

Not addressed: the Windows and Linux legs are still unproven locally (darwin only) — the new mode-only test is precisely what makes the Windows leg meaningful, so CI is the check that matters there.

ashtom:
**Follow-up from Claude (run by @ashtom):** @suhaanthayyil pushed `b3d52a28` (merge of `main`) and
`73b15175` while I was writing — so this is resolved, and I have re-verified it rather than assuming.

The fix in `73b15175` carries `OldMode`/`NewMode` through all three `admitChangedFiles` rewrite
branches, which is exactly the defect. You also added
`TestAnalyzeGitRangeClassifiesSymlinkFromTreeModeAlone`, which builds the symlink straight into the
index with `update-index --cacheinfo 120000` under `core.symlinks=false` and covers **both** the
added and deleted sides — better than the single-case probe I suggested, and it is the one symlink
test that will actually run on the Windows leg.

Re-run on your current tip `73b15175`:

```
--- PASS: TestAnalyzeGitRangeSkipsSymlinkTreeEntries
--- PASS: TestAnalyzeGitRangeSkipsDeletedSymlink
--- PASS: TestAnalyzeGitRangeStillAnalyzesRegularFiles
--- PASS: TestAnalyzeGitRangeClassifiesSymlinkFromTreeModeAlone
    --- PASS: .../deleted
    --- PASS: .../added
ok   github.com/entireio/entire-graph/internal/sem
```

The HIGH finding above is **cleared**. Nothing further from me; I did not need to push anything.
My remaining note is the unchanged one: the `--name-status` → `--raw` switch in `gitutil.ChangedFiles`
is a shared-behaviour change worth a blast-radius look, not just a symlink one.

suhaanthayyil:
Thanks for re-verifying rather than assuming — and for the note on the test shape, which I agree was the better construction.

Closing your one remaining item: **the `--name-status` → `--raw` blast radius.** `116a4e91`.

**The surface is smaller than "shared helper" suggests.** `gitutil.ChangedFiles` has exactly two production callers, both in `internal/sem/analyze.go` (`:114` and `:136`), plus one test file. It is not a widely-consumed helper; the switch reaches the diff path and nothing else.

**On strictness, the question is not "is strict better" but "can git hand it something it would now reject".** I probed the shapes a two-tree `git diff` actually emits, and all parse:

```
A  old=""      new="exec.sh"      mode 000000->100644
D  old=""      new="quote\".go"   mode 100644->000000
R  old="a.go"  new="renamed.go"   mode 100644->100644
merge-head diff: 4 files
```

covering rename (two pathnames in one record), add, delete, mode change, pathnames containing a space and a double quote, **a merge commit as head**, an empty diff, and a pathspec matching nothing.

The merge case is the one I most wanted to rule out: a combined diff emits `::`-prefixed metadata with extra modes, which this parser would reject. It does not arise — `git diff <base> <head>` is a two-tree diff and stays in plain raw form even when the head is a merge. Combined output needs `-c`/`--cc` or a bare `git diff <merge>`, neither of which this call makes.

So the strictness is fail-closed on input git cannot produce at this call site. For a semantic diff that is the right direction: the old parser's silent skip turns an unreadable record into a *missing file*, and a partial file list is a wrong answer rather than a degraded one — the analyzer cannot tell "this file did not change" from "I could not read the record".

Pinned as `TestChangedFilesParsesEveryTwoTreeDiffShape` so the contract is checked rather than argued. Mutation: tightening the field-count guard from `< 5` to `< 6` makes it fail; restoring passes. `go test ./internal/gitutil/` — 110 pass.

ashtom:

**Verdict: changes required — 1 open finding, which I reproduced**

Head moved `fe738158` → `116a4e91`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### My HIGH finding is cleared

`73b15175` carries `OldMode`/`NewMode` through all three `admitChangedFiles` rewrite branches, and
`116a4e91` adds raw-parser coverage. All four symlink tests pass on the merged tree. Thank you —
and `TestAnalyzeGitRangeClassifiesSymlinkFromTreeModeAlone` is a better test than the probe I
suggested, because it covers both the added and deleted sides.

### The remaining finding is real — I verified it rather than taking it on trust

Finding `01M17E8MH27Z` says a path changing between a regular file and a symlink (Git status `T`)
is skipped on **both** sides, so the real symbols are never reported as removed or added. I built
that case on your current head `116a4e91`:

```go
// base: alias.py is an ordinary Python file defining will_vanish()
// head: alias.py is replaced by a symlink to pkg/real.py   -> git status "T"
result, _ := AnalyzeGitRange(ctx, repo, base, head, nil)
```

```
files reported: 0
FINDING CONFIRMED: removal of `will_vanish` is not reported for a regular->symlink type change
```

`will_vanish` was a genuine symbol at base and is gone at head, and nothing in the result says so.
The guard's two conditions are independent, so a `T` entry with `OldMode=100644` / `NewMode=120000`
satisfies `afterSymlink` and takes the `continue`, discarding the non-symlink side with it. The
reviewer's suggested shape — process the non-symlink side as a deletion or addition — matches what
I observed.

```
01M17E8MH27Z  Medium  open    current    internal/sem/analyze.go:237  When a path changes between a regular source file and a symlink, Git emits statu…
```

### Verified vs. assumed

suhaanthayyil:
Reproduced the finding independently before touching anything, then fixed it in `c2cbce66`.

### `01M17E8MH27Z` — `analyze.go:237`, a symlink type change skips both sides — **VALID, and a regression against `main`**

Fixture, both directions, on `116a4e91`:

```
:100644 120000 14f44aa 3eecbec T  alias.py     (regular file -> symlink)
:120000 100644 3eecbec e92ae2b T  alias.py     (symlink -> regular file)
```

```
branch 116a4e91   R->S  files=0  warnings=1   "head version is a symbolic link"
branch 116a4e91   S->R  files=0  warnings=1   "base version is a symbolic link"
```

`alias.py` defined `will_vanish()` at base and is a link at head; nothing in the result says the symbol went away. The mirror case adds `now_real()` and nothing says it arrived.

The part worth flagging: the same fixture on `origin/main` (`bb41f223`) reports both.

```
main bb41f223   R->S  files=1   status=T  Python  [removed:will_vanish]
main bb41f223   S->R  files=1   status=T  Python  [added:now_real]
```

So this is not a pre-existing gap the PR failed to close — it is behaviour the PR removed. The two guard conditions are independent, so a `T` entry with `OldMode=100644`/`NewMode=120000` satisfies `afterSymlink` alone and the `continue` discards the non-symlink side with it.

**Fix** `analyze.go:264`. Only the link side is unanalyzable; the other side is ordinary source. When exactly one side is a symlink and the status is not already `A`/`D` (those have only one side, and it is the link), read the file as the one-sided change it is — `D` when the head became a link, `A` when the base was one — and skip only the link side. The informational warning stays, now saying `symbolic-link side skipped; the file side is analyzed as a one-sided change` instead of `file skipped`.

The delta keeps the status Git reported (`analyze.go:217`, `analyze.go:475`). The path changed type; it was not added or deleted, and it still exists at head. Rewriting the status only steers which blobs get read.

After the fix, identical entity output to `main`, plus the honest warning `main` never emitted:

```
R->S  files=1  status=T  Python  [removed:will_vanish]  + W_UNSUPPORTED_FILE "head version is a symbolic link"
S->R  files=1  status=T  Python  [added:now_real]       + W_UNSUPPORTED_FILE "base version is a symbolic link"
```

**Test** `TestAnalyzeGitRangeReportsTheFileSideOfASymlinkTypeChange` (`analyze_test.go:2132`), two sub-tests, one per direction, each asserting the status, the entity change and the surviving link-side warning.

**Mutation**, both halves, each compiling:
- force `typeChange := false && …` -> `want exactly one changed file, got []sem.FileChange(nil)`, both directions.
- report `file.Status` instead of `reportedStatus` -> `status "D", want the type change Git reported (T)` and `status "A", …`.

Restored, both pass.

### Verification

- `go vet ./internal/sem/ ./internal/gitutil/`, `gofmt -s -l internal/` — clean.
- `go test ./internal/gitutil/` — pass (11.4s).
- `go test ./internal/sem/ -run 'TestAnalyzeGitRange'` — pass (12.8s). The four existing symlink tests still hold: a plain symlink add, modify or delete has only one side and is still skipped whole; `TestAnalyzeGitRangeClassifiesSymlinkFromTreeModeAlone` and `TestAnalyzeGitRangeStillAnalyzesRegularFiles` unchanged.
- **Golden delta: none.** `TestProviderGoldenSnapshots` and `TestProviderGoldenFixtureQualityCoverageReport` pass without `-update`; the commit touches no `testdata/` file. The fixture corpus has no type-change entry, so this only affects a shape it does not contain.

ashtom:

`main` has moved to `7b3252aa`. Re-ran everything against that head.

**Verdict: changes required — `test (windows-latest)` is failing, and it is the new test that fails**

### The Windows leg is red

```
--- FAIL: TestChangedFilesParsesEveryTwoTreeDiffShape (0.10s)
    changed_files_shapes_test.go:50: open C:\...\001\quote".go:
      The filename, directory name, or volume label syntax is incorrect.
FAIL  github.com/entireio/entire-graph/internal/gitutil
```

The fixture writes `quote".go`. Windows forbids `"` in a filename (along with `<>:|?*`), so
`os.WriteFile` fails and the helper `t.Fatal`s before the parser is ever exercised. This is the same
trap `main` already hit and reverted in `15b74eb2` ("Git for Windows refuses the names").

### The property is worth keeping — there is a cross-platform way to keep it

The double quote is there because it is a name `--name-status` would C-quote and `-z --raw` emits
raw, which is the discriminating case for this parser. A **non-ASCII** name has the same property and
Windows accepts it. I checked rather than assuming:

```
$ git diff --name-status HEAD~1 HEAD
D	"caf\303\251.go"
D	"quote\".go"          <- both C-quoted

$ git diff -z --raw HEAD~1 HEAD | tr '\0' '.'
:100644 000000 7f96c22 0000000 D.café.go.:100644 000000 142fb42 0000000 D.quote".go.
                                 ^ both raw
```

`core.quotePath` defaults to true, so non-ASCII is quoted by `--name-status` and raw under
`-z --raw` — the same contrast, on every platform. Swapping `quote".go` for something like
`café.go` keeps the coverage instead of skipping it.

Two honest caveats: I verified this on macOS only, and macOS/Windows differ in Unicode normalization
for filenames, so it deserves a CI run rather than my word. If you would rather not take that on, the
minimal fix is to guard just the quote path behind `runtime.GOOS != "windows"`, the way the newline
path already is in `head_source_provenance_test.go` — but that gives up the case on the one platform
whose path rules differ most.

### Everything else is green

The gate passes on the merged tree here (macOS), the type-change finding I reproduced yesterday is
now resolved, and there are **no open, current trail findings**. The Windows leg is the only blocker.

suhaanthayyil:
Two things: the stated blocker is **already fixed** at the head you re-checked, and your suggested improvement is **right and now landed** — with one correction to the character choice.

## The Windows leg is green

`test (windows-latest)` **passes** on `86e77485`, the head your comment names:

```
test (windows-latest)   pass   19m24s
build (windows-latest)  pass   1m1s
```

`86e77485` is titled "cover the quoted pathname only where the OS can hold it" — it guards the `quote".go` fixture behind `runtime.GOOS != "windows"`. Your re-run must have caught the earlier head. Worth flagging since the verdict was "changes required": the failure you quoted is real but was already resolved before you posted. (Entire Gates is red, but that is the `phase=building` pipeline problem, not this change.)

## Your improvement is the better fix, and I took it — `f4b1b3e0`

You are right that the GOOS guard "gives up the case on the one platform whose path rules differ most". A non-ASCII name has the identical property and every platform accepts it, so the coverage is restored everywhere rather than skipped on Windows. Confirmed locally exactly as you described:

```
git diff --name-status  ->  D  "\346\227\245\346\234\254.go"   C-quoted
git diff -z --raw       ->  D  日本.go                          verbatim
```

## One correction: CJK, not `café`

You caveated that macOS and Windows differ in Unicode normalization and that it deserved a CI run. That caveat is load-bearing, and it rules out your specific example. `é` is **decomposable** — NFC `U+00E9` versus NFD `U+0065 U+0301` — and macOS and Windows disagree on which form the filesystem returns. This test compares the pathname the parser returns, so `café.go` could come back with different bytes than were written and fail on one platform for a reason that has nothing to do with the parser.

CJK ideographs have **no canonical decomposition**, so NFC and NFD are byte-identical and the hazard disappears. I verified all three candidates carry the quoting property before choosing:

```
--name-status:  "caf\303\251.go"   "\321\204\320\260\320\271\320\273.go"   "\346\227\245\346\234\254.go"
-z --raw:        café.go            файл.go                                  日本.go
```

Any of them discriminates; only the CJK one is normalization-stable.

## Verification

The fixture is now exercised rather than skipped — `D old="" new="日本.go" mode 100644->000000` appears in the parsed output. `go test ./internal/gitutil/` passes, 11.0s. Mutation: tightening the field-count guard from `< 5` to `< 6` fails the test; restoring passes; the mutation compiles and I confirmed by `grep` that it applied.

**Verified on macOS only** — the Windows and Linux legs are CI's to confirm, and I would rather say that than imply I ran them.

ashtom (approved):
Approved after syncing with current main. The refreshed Entire review has zero current findings; required CI is running on this exact head.

PR #167, opened 2026-08-28
fix(sem): honor .graphignore and the secret rules in the semantic diff

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/126

## Problem

`entire graph index --help` states the contract:

> A repo-root `.graphignore` (gitignore syntax) is honored by every graph command, on top of `.gitignore`.

The diff family did not honor it. `snapshot` / `symbols` / `search` apply the matcher through `loadExplicitIgnoreMatcher`; `diff`, `analyze`, `commit`, and `checkpoint` never loaded one.

```
$ cat .graphignore
vendored/

$ entire graph symbols --repo . --format ndjson | grep -c '"file_path":"vendored/'
0

$ entire graph diff --base HEAD~1 --head HEAD
keep/k.go (Go)
  ~ function Keep body changed (0 dependents)

vendored/gen.go (Go)          <-- never indexed, never in any snapshot
  ~ function Gen body changed (0 dependents)
  + function Extra added
```

`.graphignore` exists for tracked-but-vendored or generated sources — the tree-sitter `parser.c` blobs the help text names, plus generated clients, bundled vendor trees, and checked-in build output. Those are exactly the paths that churn, so the diff emitted entity changes for symbols that no snapshot of the repository contains and that no consumer building its index from `snapshot` can resolve.

## Fix

The diff must report what happened to the **index**, not what happened to a **path**. Those are the same thing only while a change stays on one side of the boundary between indexed and unindexed; a filter that asks one question — *is the destination ignored?* — gets both crossings wrong.

So membership is decided per side. `diffIndexPolicy` answers, for **one** compared tree, the question `openSource` asks of its own committed listing — would the graph index this path? — applying the same two filters in the same order. The policy is per-tree because its halves answer differently on the two sides: the `.graphignore` matcher is read from the working tree and is common to both, but the vendored-tree rules come from each compared tree's own `.gitignore` files, so a project can re-include part of a vendored tree in one revision and not the other. Each changed file is then rewritten to the delta the graph saw:

| base indexed | head indexed | emitted |
| --- | --- | --- |
| yes | yes | unchanged |
| yes | no | a deletion of the base path — its symbols leave the graph |
| no | yes | an addition of the head path — all of its symbols are new |
| no | no | dropped |

`D` and `A` are ordinary name-status inputs and the analysis loop already derives its read plan from `Status` alone, so this needs no new emission code. A copy (`C`) is never a comparison at all: its source survives untouched, so the destination is purely an addition — never a deletion of a file that still exists, and never a change reported against the file it was copied from. Git emits `C` only under `--find-copies-harder` and `ChangedFiles` asks for `--find-renames`, so that is a guard against a future flag rather than a path exercised today.

That covers three divergences from the snapshot, not one.

**`.graphignore` and the built-in secret rules.** The original case above, plus committed credential files, which were excluded from the snapshot but named by the diff:

```
$ entire graph diff --base HEAD~1 --head HEAD        # before
.env (Dotenv)
  ~ document .env body changed (0 dependents)

s.go (Go)
  ~ function S body changed (0 dependents)
```

**The vendored-tree heuristic.** `openSource` applies `filterVendoredPaths` *and* `filterIgnoredPaths`; only the second was in scope before. A tracked `vendor/` tree is the canonical tracked-but-vendored case and the one that churns:

```
$ entire graph symbols --repo . --format ndjson | grep -o '"file_path":"[^"]*"'
"file_path":".gitignore"
"file_path":"keep/k.go"

$ entire graph diff --base HEAD~1 --head HEAD        # before
keep/k.go (Go)          ~ function Keep body changed
third_party/t.go (Go)   ~ function T body changed    <-- in no snapshot
vendor/v.go (Go)        ~ function V body changed    <-- in no snapshot
```

Its rules are per-tree and cost a listing, so they load only when some changed path is candidate-vendored under **empty** rules. That probe is exact rather than approximate: rules reach the heuristic only through `ReincludesDescendant`, which can only ever *un*-vendor a path, so an empty-rule probe is the maximal candidate set and "no candidate here" proves the real rules would change no verdict.

**Dependent counts.** The dependents scan reads the whole head tree, so `dependents_count` included references from files no snapshot contains — a count of 4 against one visible caller. Once the diff stopped naming those files the number became unaccountable, so the head-side policy now gates the candidate list before any candidate is read. That scan gets its own resolution of the head rules rather than the changed-file policy: the probe above licenses one question — could any path *Git reported* be vendored — while this scan walks the whole tree, where a re-included `vendor/...` caller can appear even when nothing vendorable changed. Reusing the probe's answer there would drop a real caller and undercount, which is the silent direction.

### Where no policy may be applied at all

`ChangedFiles` passes `--no-relative`, so a range between two **commits** is named from the repository root — the namespace both rule sets are written in. A tree expression such as `HEAD:scope` names a subtree, and every emitted name is then relative to *that* tree, where root-anchored rules both miss what they meant to match and fire on what they did not:

```
$ cat .graphignore
/top.go                                     # the ROOT top.go, not sub/top.go

$ entire graph diff --base HEAD~1:sub --head HEAD:sub
No semantic entity changes detected.        # but sub/top.go IS indexed
```

A tree object cannot know its own path, so there is no prefix to recover. `resolveDiffTrees` now reports whether both labels named a commit — probing the resolved OID rather than the caller's expression, because appending `^{commit}` to `HEAD:scope` can select a path literally named `scope^{commit}`, the same ambiguity the existing `^{tree}` peel avoids — and a range that is not root-relative admits everything, exactly as this command did before any policy existed.

Resolving to a commit **object** is not on its own evidence that a label named a commit: a gitlink resolves to one too. `HEAD:sub` on a submodule entry yields that submodule's commit, and a range over its tree is named relative to the *submodule's* root. That probe happens to fail today only because a submodule's objects are usually absent from the superproject's object database — `git rev-parse HEAD:sub` succeeds there while `^{commit}` on the result does not — but an absorbed or fetched submodule puts them there, so a label that reaches into a tree is rejected on its shape as well. `:` is the only object-name syntax that does that — but not every colon is that syntax. `:/text` searches commit messages, and a colon inside an `@{...}` selector is data too: `HEAD@{2026-08-27 12:34:56 +0000}` is a reflog date that names a commit and reaches into no tree. Only a colon outside every such block is the separator. Getting that wrong loses no data — an unfiltered range over-reports — but it quietly withdraws the guarantee the range was supposed to have.

## Behaviour changes to expect

- **Lockfiles and sourcemaps disappear from `diff` / `commit` output.** `package-lock.json`, `yarn.lock`, `Cargo.lock`, `*.map` and friends are vendored-scan files: no snapshot contains them, so no diff reports them. This is the most visible consequence of the change.
- **A rename across the boundary now emits where nothing was emitted before.** A pure rename produced no entity changes at all, even when it moved a whole tree out of the graph. It is now a deletion of everything the index held, which is the delta a consumer actually needs.
- There is no behaviour change for a repository with no `.graphignore`, no vendored tree, and no secret-rule match.

### The one gap, and why it is disclosed rather than closed

A committed exclusion rule decides membership for files it never names. Deleting a `!vendor/pkg/**` re-inclusion drops that whole subtree from the head snapshot while Git reports only `.gitignore` as changed, so the correct delta contains removals for files that appear nowhere in this command's input. No filter over a change list can produce them: it would take a comparison of the two trees' entire corpora, and could emit an entry per file in a vendored tree — a different and unbounded operation.

That gap predates the index policy, because it is a property of diffing changes rather than corpora. But this command now claims its output agrees with a snapshot, and an unstated exception to that claim is worse than not making it. So a range that changes `.gitignore` or `.graphignore` carries `W_INDEX_POLICY_CHANGED`: unchanged files may have entered or left the graph, and this delta alone is not enough to update an index. A consumer gets a signal to resnapshot instead of silent staleness.

A second, older mismatch is left alone deliberately. `--repo <subdir>` means different things to the two families: `symbols`/`snapshot` treat the subdirectory as the repository, emitting subtree-relative paths and reading that subdirectory's `.graphignore`, while the diff family resolves Git's command root and emits repository-root-relative paths. Pairing a subdirectory-scoped snapshot with a subdirectory-scoped diff therefore compares two different namespaces, in both directions, and it did so before this change too. Reconciling them means deciding what `--repo <subdir>` should mean for the diff family, which changes the paths it prints for every consumer of that flag — a contract change that belongs in its own review, not in this one.

## Tests

Assertions are made against a snapshot of the **same repository**, not against a hand-written symbol list, so the diff and the snapshot cannot drift apart without a failure here.

- `TestAnalyzeGitRangeHonorsGraphIgnore` — a change touching both an indexed and an ignored tree reports only the indexed one.
- `TestAnalyzeGitRangeHonorsGraphIgnoreForDeletions` — deleting an ignored file produces no delta.
- `TestAnalyzeGitRangeWithoutGraphIgnoreIsUnchanged` — with no ignore rule, the same file is still reported.
- `TestAnalyzeGitRangeHonorsBuiltinSecretRules` — a committed credential file is not named.
- `TestAnalyzeGitRangeIndexedToIgnoredRenameRemovesOldSymbols` — the removals equal the base snapshot's symbols for that file.
- `TestAnalyzeGitRangeIgnoredToIndexedRenameAddsEveryHeadSymbol` — the additions equal the head snapshot's symbols, including the ones a rename comparison never mentioned.
- `TestAnalyzeGitRangeSubtreeRangeAppliesNoIgnorePolicy` — a subtree range reports the indexed file a root-anchored rule was dropping.
- `TestAnalyzeGitRangeTreePeeledRangeStillAppliesIgnorePolicy` — `<commit-ish>^{tree}` is still root-relative, so the policy still applies.
- `TestAnalyzeGitRangeHonorsVendoredTrees` — a tracked `vendor/` tree is not reported.
- `TestAnalyzeGitRangeReportsReincludedVendorTree` — a re-included vendor tree **is** reported; the snapshot indexes it.
- `TestAnalyzeGitRangeDependentsExcludeUnindexedCallers` — three callers inside an ignored tree and one indexed caller counts 1.
- `TestAdmitChangedFilesRewritesIndexCrossings` — the rewrite table itself, including all three copy cases.
- `TestResolveDiffTreesRejectsTreePathLabels` — a gitlink label is not root-relative even when every peel resolves; `:/text` still is.
- `TestLabelSelectsTreePath` — which colons are the `<rev>:<path>` separator and which are reflog or search-text data.
- `TestAnalyzeGitRangeDependentsIncludeReincludedVendorCallers` — a re-included vendor caller is still counted when nothing vendorable changed.
- `TestAnalyzeGitRangeWarnsWhenAnIndexPolicyFileChanges` — the range says so when a rule file changed, stays quiet when none did, and does not warn about a non-root `.graphignore`.
- `TestAnalyzeGitRangeWarnsAboutPolicyChangesOutsideThePathspec` — a pathspec narrows what Git reports, not what a rule decides.
- `TestAnalyzeGitRangeAcceptsEveryRootTreeSpelling` — `^{tree}`, `^{tree}^{}` and `^{tree}^{object}` all name a root tree.

Mutation-checked, each against the test that is supposed to catch it: restoring the destination-only filter fails both crossing tests; removing the root-relative gate fails the subtree test; removing the vendored-tree filter fails the vendor test; never loading the re-inclusion rules fails the re-inclusion test; removing the dependents filter restores the inflated count; removing the copy branch fails the copy cases; removing the tree-path guard fails the gitlink test; removing the `^{tree}` peel fails the peeled-range test; treating every colon as a path separator fails the reflog case; reusing the unresolved policy for the dependents scan fails the re-included-caller count; suppressing the policy-change warning fails one half of its test and emitting it for every file fails the other; and widening the policy to drop everything fails all three over-rejection guards.

---

**Local gate**

- `go test -timeout 40m ./...` — green, every package `ok` (internal/sem 251s, internal/cli 69s).

### DIFF
--- a/internal/sem/analyze.go
+++ b/internal/sem/analyze.go
@@ -122,11 +122,23 @@ func AnalyzeGitRangeWithOptions(ctx context.Context, repo, base, head string, pa
 	// graph never indexes still produced entity changes here. A consumer that
 	// builds its index from `snapshot` and updates it from `diff` received
 	// symbols for files that its index has no record of and never will.
-	basePolicy, headPolicy, err := newDiffIndexPolicies(ctx, objectRepo, pinnedBase, pinnedHead, rootRelativeNames, changed)
+	basePolicy, headPolicy, err := newDiffIndexPolicies(ctx, objectRepo, pinnedBase, pinnedHead, rootRelativeNames, changed, overBudget)
 	if err != nil {
 		return Result{}, err
 	}
-	policyWarnings := indexPolicyChangeWarnings(changed, basePolicy)
+	policySource := changed
+	if len(paths) > 0 {
+		// A pathspec narrows what Git reports, but not what an exclusion rule
+		// decides: a root .gitignore edited outside the requested scope still
+		// changes membership inside it, and the scoped list cannot show that.
+		// Ask Git for the policy files directly, anchored at the repository root
+		// with :(top) so the caller's cwd does not narrow them a second time.
+		policySource, err = gitutil.ChangedFiles(ctx, repo, pinnedBase, pinnedHead, indexPolicyPathspecs)
+		if err != nil {
+			return Result{}, err
+		}
+	}
+	policyWarnings := indexPolicyChangeWarnings(policySource, basePolicy)
 	changed = admitChangedFiles(changed, basePolicy, headPolicy)
 	emitProgress("parse", 0, len(changed))
 	parser := TreeSitterParser{}
@@ -485,6 +497,11 @@ func AnalyzeGitRangeWithOptions(ctx context.Context, repo, base, head string, pa
 // to apply and fires rules that were not, which silently drops indexed files.
 // There is no prefix to recover, because a tree object does not know its own
 // path, so the only correct answer for a subtree range is to apply no policy.
+// maxRevisionPeels bounds how many trailing ^{...} groups are stripped while
+// deciding whether a label bottoms out at a commit. Each step costs a git
+// subprocess, and no real revision expression stacks more than a couple.
+const maxRevisionPeels = 4
+
 // labelSelectsTreePath reports whether a revision label reaches INTO a tree
 // rather than naming a revision.
 //
@@ -553,18 +570,28 @@ func resolveDiffTrees(
 		if _, err := resolve(ctx, repo, objectID+"^{commit}"); err == nil {
 			return true
 		}
-		// A caller who wrote <commit-ish>^{tree} still named a root tree. Peel
-		// the suffix THEY wrote — adding none of our own — and ask again.
-		peeled, found := strings.CutSuffix(label, "^{tree}")
-		if !found {
-			return false
+		// A caller who wrote <commit-ish>^{tree} still named a root tree, and Git
+		// spells that several ways: ^{tree}, ^{tree}^{}, ^{tree}^{object}. Peel
+		// the trailing groups THEY wrote — adding none of our own — and ask again
+		// after each, so any spelling that bottoms out at a commit is recognized.
+		// The peel count is capped because each step costs a subprocess and a
+		// label can carry arbitrarily many groups.
+		peeled := label
+		for range maxRevisionPeels {
+			open := strings.LastIndex(peeled, "^{")
+			if open <= 0 || !strings.HasSuffix(peeled, "}") {
+				return false
+			}
+			peeled = peeled[:open]
+			peeledID, err := resolve(ctx, repo, peeled)
+			if err != nil {
+				continue
+			}
+			if _, err := resolve(ctx, repo, peeledID+"^{commit}"); err == nil {
+				return true
+			}
 		}
-		peeledID, err := resolve(ctx, repo, peeled)
-		if err != nil {
-			return false
-		}
-		_, err = resolve(ctx, repo, peeledID+"^{commit}")
-		return err == nil
+		return false
 	}
 	resolveTree := func(label string) (string, bool, error) {
 		// Resolve the caller's expression exactly before adding syntax of our
@@ -1091,6 +1118,12 @@ func indexPolicyChangeWarnings(changed []gitutil.ChangedFile, policy diffIndexPo
 	return warnings
 }
 
+// indexPolicyPathspecs names every committed file that can decide index
+// membership, anchored at the repository root so a caller's cwd cannot narrow
+// them. Git matches a wildcard pathspec at any depth, which is what nested
+// .gitignore files need.
+var indexPolicyPathspecs = []string{":(top)*.gitignore", ":(top)" + graphIgnoreFileName}
+
 // isIndexPolicyFile reports whether a path is one of the committed files whose
 // contents decide what the graph indexes.
 //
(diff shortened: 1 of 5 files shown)


ashtom:
bugbot run

ashtom:
bugbot run

ashtom:
bugbot run

PR #168, opened 2026-08-28
fix(sem): declare the type and data-flow relations ten languages emit

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/127

## Impact

`capabilities --json` is the contract AGENTS.md tells agents to feature-detect
with ("before assuming a language is semantically parsed"), and it
**under-reported ten languages**. Each of these reaches the generic type and
data-flow passes and emits relations the report says it cannot:

| language | emitted | declared |
|---|---|---|
| Clojure | USES_TYPE | — |
| Elixir | DATA_FLOWS | — |
| Erlang | USES_TYPE, PARAM_TYPE | — |
| F# | USES_TYPE, PARAM_TYPE | — |
| Haskell | USES_TYPE, PARAM_TYPE | — |
| Julia | USES_TYPE, PARAM_TYPE | DATA_FLOWS only |
| Lua | DATA_FLOWS | — |
| Objective-C | DATA_FLOWS | — |
| Perl | DATA_FLOWS | — |
| R | USES_TYPE, DATA_FLOWS | — |

An agent that follows the report skips relation families the provider does
produce. This is the same defect the existing `ooRelationSupport` comment
records for C, C++, Groovy and Julia ("they were simply never added here, so
`capabilities --json` under-reported them") — one round later, for ten more
languages.

## Repro

```haskell
-- Fix.hs
data Point = Point { px :: Int, py :: Int }
add :: Int -> Int -> Int
add a b = a + b
pointSum :: Point -> Int
pointSum p = add (px p) (py p)
main = print (pointSum (Point 1 2))
```

```
$ entire graph edges --repo . --format ndjson | grep -E 'PARAM_TYPE|USES_TYPE'
... "from_id":"...:Haskell:...","type":"USES_TYPE" ...
... "from_id":"...:Haskell:...","type":"PARAM_TYPE" ...

$ entire graph capabilities --json | jq '.relation_support_by_language."Haskell"'
["CALLS","CONSTRUCTS","CONTAINS","DEFINES"]
```

## Root cause

`ooRelationSupport` is a hand-maintained map; the type, field and data-flow
passes are generic and run regardless of whether a language has an entry.
`TestCapabilityMatrixCoversEmittedRelations` is the guard for exactly this
divergence, but it is driven by `goldenFixtures`, and none of those fixtures
carries an annotated signature or a forwarded argument in these ten languages,
so nothing ever tripped it.

## Fix

Add the ten entries, each listing only what is observed.

Add `TestCapabilityMatrixDeclaresTypeAndFlowRelations` as a companion guard that
does not depend on golden-fixture coverage: it builds one repository with an
annotated, argument-forwarding file per language and asserts the same invariant
directly. It also asserts each language actually produced symbols, so the check
cannot pass by extracting nothing.

## Mutation verification

- Before the fix, the new test reports all 15 violations across the 10 languages.
- Dropping `PARAM_TYPE` from the new `Haskell` entry: FAIL — `Haskell emits PARAM_TYPE but capabilities declares neither it per-language nor as a global heuristic`.
- Removing the `Lua` entry entirely: FAIL — `Lua emits DATA_FLOWS ...`.
- With the fix: PASS.

No extraction behaviour changes; this only makes the report describe what the
provider already does.

## Verification run locally

full gate on a clean runner.

---

## Rebased on `main` after #181 — what this PR now changes

#181 merged, supplying most of the declarations this branch first added. Merged
`main` in; the remaining delta against `main` is the `ooRelationSupport` map and
nothing else. `git diff origin/main -- internal/sem/provider.go` touches only
that map.

| language | `main` declares | this PR declares | delta |
|---|---|---|---|
| JavaScript | EXTENDS, INHERITS, HANDLES_GRAPHQL, HANDLES_TRPC, ASYNC_CALLS, DATA_FLOWS | + USES_TYPE | **add** USES_TYPE |
| Ruby | INHERITS, DATA_FLOWS | + USES_TYPE | **add** USES_TYPE |
| ClojureScript | *(no key at all)* | USES_TYPE | **new key** |
| Groovy | USES_TYPE, PARAM_TYPE, READS_FIELD, DATA_FLOWS | + RETURNS_TYPE, WRITES_FIELD | **add** ×2 |
| Objective-C | DATA_FLOWS | + USES_TYPE, PARAM_TYPE, RETURNS_TYPE | **add** ×3 |
| Swift | USES_TYPE, PARAM_TYPE, RETURNS_TYPE, DATA_FLOWS | + READS_FIELD, WRITES_FIELD | **add** ×2 |
| Zig | USES_TYPE, PARAM_TYPE, RETURNS_TYPE, **READS_FIELD**, DATA_FLOWS | USES_TYPE, PARAM_TYPE, RETURNS_TYPE, DATA_FLOWS | **REMOVE** READS_FIELD |

### This PR removes `Zig: READS_FIELD` from `main`, deliberately

#181 landed with `Zig` keeping `READS_FIELD`. That relation is unreachable for
Zig, and this PR takes it out. The mechanism, not a fixture, is the evidence:

1. `fieldAccessRelations` (provider.go) is the **sole** producer of
   `READS_FIELD` and `WRITES_FIELD` — it is the only place either string is
   constructed outside the declaration tables.
2. It resolves every edge through `fieldsByContainer[containerID][field]`.
3. `fieldsByContainer` is populated under one condition: `symbol.Kind == "field"`.
4. `fieldEntities` admits only `field_declaration`, `public_field_definition`,
   `field_definition`, `property_signature` and `property_declaration`.
   tree-sitter-zig names a struct member **`container_field`** — a string that
   appears in **zero** Go sources in this repository (`grep -rn container_field
   --include='*.go' . | wc -l` → `0`), and 63 times in the vendored
   `grammars/zig/parser.c`.
5. Extracting the probe confirms it: Zig yields `struct Point`, `function add`,
   `function pointSum`, `function make` and **no symbol of kind `field`**.

So no field symbol exists for a Zig `READS_FIELD` edge to land on, however the
body is written. An over-declared relation fails an agent exactly the way an
under-declared one does: AGENTS.md tells it to feature-detect against
`capabilities --json`, so it reads the relation, queries for it, and gets
nothing back.

### The over-declaration direction now has its own guard

`TestCapabilityMatrixCoversEmittedRelations` fails only on a relation emitted
**without** a declaration. It cannot fail on a declaration with nothing behind
it, however many fixtures are added — which is why `Zig: READS_FIELD` survived
on `main`. Two tests added for the missing direction:

- **`TestCapabilityDeclaredFieldRelationsRequireAFieldSymbol`** — a language may
  not declare `READS_FIELD`/`WRITES_FIELD` unless its extractor produces a
  symbol of kind `field`, checked per language over the isolated single-language
  probes. This is structural rather than incidental: the equality test also
  catches Zig, but only because that probe body happens to contain a field
  access, so editing the probe would silently drop the guard.
- **`TestCapabilityReportCarriesEveryDeclaredRelation`** — every
  `ooRelationSupport` key reaches `capabilities --json`.
  `relationSupportByLanguage` builds the report by walking the language *specs*
  and folding the map in per language, so a key naming a language no spec
  produces is dropped without a word. A new key — `ClojureScript` here — is
  exactly the shape that can go missing that way.

### Mutation verification on the merged tree

| mutation | result |
|---|---|
| revert `internal/sem/provider.go` to `origin/main`, keep every test | **FAIL** — `TestCapabilityRelationDeclarationsMatchIsolatedExtraction` fails on all 7 delta languages: ClojureScript, Groovy, JavaScript, Objective-C, Ruby, Swift, Zig |
| restore | **PASS** — 33 subtests |
| re-add `READS_FIELD` to `Zig` only | **FAIL** — both `…RequireAFieldSymbol/Zig` and `…MatchIsolatedExtraction/Zig` |
| add a key for a language no spec produces (`"Nonesuch"`) | **FAIL** — `…CarriesEveryDeclaredRelation`: *declares [USES_TYPE] for Nonesuch but capabilities --json reports no such language* |
| with the change | **PASS** — `go test ./internal/sem/ -run TestCapability`, 47 subtests |

`gofmt -l internal/` clean, `go vet ./internal/sem/` clean.


suhaanthayyil:
Per-language extraction audit turned up four more languages that emit relations this PR does not declare. Measured on `origin/main` with one minimal fixture per semantic language (#171), comparing every emitted relation type against `relationSupportByLanguage()`, excluding the heuristic set:

```
UNDECLARED Dart      [DATA_FLOWS]
UNDECLARED Elixir    [DATA_FLOWS]    <- covered here
UNDECLARED Erlang    [DATA_FLOWS]
UNDECLARED F#        [PARAM_TYPE USES_TYPE]  <- covered here
UNDECLARED Haskell   [PARAM_TYPE USES_TYPE]  <- covered here
UNDECLARED Julia     [PARAM_TYPE USES_TYPE]  <- covered here
UNDECLARED Lua       [DATA_FLOWS]    <- covered here
UNDECLARED OCaml     [USES_TYPE]
UNDECLARED SQL       [DATA_FLOWS]
```

So four are still short after this PR:

- **Dart** — `DATA_FLOWS`; Dart has no `ooRelationSupport` entry at all.
- **Erlang** — `DATA_FLOWS`; this PR adds `USES_TYPE` and `PARAM_TYPE` for Erlang but not the flow edge.
- **OCaml** — `USES_TYPE`; no entry at all.
- **SQL** — `DATA_FLOWS`; no entry at all (one `CREATE FUNCTION` calling another forwards an argument).

Reproduction, once #171 is in: iterate `languageMatrix`, snapshot each fixture, and diff the emitted relation types against `relationSupportByLanguage()` minus `Capabilities().HeuristicRelationTypes`.

`TestCapabilityMatrixCoversEmittedRelations` did not catch these because it only inspects the golden fixtures, which cover 10 of the 36 semantic languages.

ashtom:

**Verdict: changes required — the Erlang declaration is incomplete, and it collides with #181**

- Live PR head `98c80088`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `98c80088` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

**1. (blocking) `Erlang` is under-declared here, and #181 declares it differently.** This PR sets

```go
"Erlang": {"USES_TYPE", "PARAM_TYPE"},
```

while **#181** sets, in the same map:

```go
"Erlang": {"DATA_FLOWS"},
```

Both PRs describe their entry as empirically derived. Neither is wrong about what it *saw* — they
are wrong about what Erlang *emits*, because each only probed its own fixture. I ran both fixtures
and their union through `BuildProviderSnapshot` on current `main` (declaration tables do not affect
emission, only reporting):

```
PR168-fixture   languages=[Erlang]  relations=[CALLS DEFINES PARAM_TYPE USES_TYPE]
PR181-fixture   languages=[Erlang]  relations=[CALLS DATA_FLOWS DEFINES]
UNION           languages=[Erlang]  relations=[CALLS DATA_FLOWS DEFINES PARAM_TYPE SIMILAR_TO USES_TYPE]
```

**Erlang emits all three.** The correct entry is:

```go
"Erlang": {"USES_TYPE", "PARAM_TYPE", "DATA_FLOWS"},
```

Because both PRs assign the whole slice for the same key, whichever merges second silently discards
the other's relations — and neither PR's guard test catches it, since each guard only sees its own
fixture. `capabilities --json` would then keep lying to agents about Erlang, which is the exact
failure this PR exists to fix.

**2. (blocking, cross-PR) This PR conflicts with both #165 and #181** in `internal/sem/provider.go`:

```
$ git merge-tree --write-tree refs/prs/165 refs/prs/168   ->  CONFLICT
$ git merge-tree --write-tree refs/prs/168 refs/prs/181   ->  CONFLICT
```

Each is clean against `main` alone. Please resolve by taking the **union** of the declarations, not
one side — and add a fixture that carries both an annotated signature *and* a forwarded argument for
any language both PRs touch, so the guard test can actually see the union.

**3. (worth a check) Julia also differs between the two PRs.** This PR widens Julia to
`{USES_TYPE, PARAM_TYPE, DATA_FLOWS}`; #181 leaves `main`'s `{DATA_FLOWS}`. That is a widening rather
than a contradiction, so it is probably fine — but it deserves the same union treatment, and I did
not separately probe Julia.

### Independent review state

suhaanthayyil:
## Finding reproduced — fixed on both branches, and it uncovered a second defect here

The Erlang collision is real:

```
$ git merge-tree --write-tree 98c80088 d17230ee   -> exit 1 (CONFLICT)
```

Both branches assign the whole slice for `"Erlang"`, so the second to land would have
discarded the first's relations with no guard noticing.

### The declarations were measured, not assumed

Each fixture went through `BuildProviderSnapshot` on `bb41f223`, then every language was
re-run in a repository containing **only that language**. Structural relations
(`CONTAINS/DEFINES/CALLS/CONSTRUCTS/IMPORTS`) and the global heuristics are excluded.

| language | this PR (#168) | #181 | measured in isolation | agreed |
|---|---|---|---|---|
| **Clojure** | `USES_TYPE` | — | `USES_TYPE` — but **not** from your fixture | kept, fixture fixed |
| Dart | — | `USES_TYPE, PARAM_TYPE, RETURNS_TYPE, ASYNC_CALLS, DATA_FLOWS` | same | added here |
| Elixir | `DATA_FLOWS` | — | `DATA_FLOWS` | unchanged |
| **Erlang** | **`USES_TYPE, PARAM_TYPE`** | **`DATA_FLOWS`** | **`USES_TYPE, PARAM_TYPE, DATA_FLOWS`** | **union — both were wrong** |
| F# | `USES_TYPE, PARAM_TYPE` | — | same | unchanged |
| Haskell | `USES_TYPE, PARAM_TYPE` | — | same | unchanged |
| Julia | `USES_TYPE, PARAM_TYPE, DATA_FLOWS` | *(left at main's `DATA_FLOWS`)* | `USES_TYPE, PARAM_TYPE, DATA_FLOWS` | **your widening is correct** |
| Lua | `DATA_FLOWS` | — | `DATA_FLOWS` | unchanged |
| OCaml | — | `USES_TYPE, PARAM_TYPE` | same | added here |
| Objective-C | `DATA_FLOWS` | — | `DATA_FLOWS` | unchanged |
| Perl | `DATA_FLOWS` | — | `DATA_FLOWS` | unchanged |
| **R** | **`USES_TYPE, DATA_FLOWS`** | — | **`DATA_FLOWS`** | **narrowed** |
| SQL | — | `DATA_FLOWS` | `DATA_FLOWS` | added here |

Your Julia widening (point 3) is confirmed correct — `USES_TYPE, PARAM_TYPE, DATA_FLOWS` is
exactly what an isolated Julia file emits. #181 now carries the same value.

### Two entries here rest on a cross-language artifact in your own fixture

Emission is not monotone in the fixture, so "run the union" is not a safe check:

```
Clojure alone (your clj/core.cljc)   extra=[]
Clojure in your 10-file fixture      extra=[USES_TYPE]
R alone (your r/fix.R)               extra=[DATA_FLOWS]
R in your 10-file fixture            extra=[DATA_FLOWS USES_TYPE]
R + your erl/fix.erl, nothing else   extra=[DATA_FLOWS USES_TYPE]   <- two files reproduce it
```

`usesTypeRelations` (`internal/sem/provider.go:9097`) resolves each identifier in a signature
against every type symbol in the repository **by short name**. Dumping the edges:

```
USES_TYPE: R/make_point      -> Erlang/point (struct) resolution=name_only scope=module
USES_TYPE: Clojure/point-sum -> Erlang/point (struct) resolution=name_only scope=module
```

Both point at the `-record(point, {x = 0, y = 0}).` in your **Erlang** fixture file. The `point`
in `make_point` / `point-sum` matched it by short name. Neither is an R or Clojure type edge.
(In a 14-file union they vanish again — OCaml's `type point` makes the name ambiguous — so a
union run would have produced a third, also wrong, answer.)

Consequences:

- **Clojure keeps `USES_TYPE`** — it is genuinely reachable, but only through a type hint. The
  fixture now carries `(defn point-sum [^Point p] ...)` and `(Point. x y)`, which resolves to
  the local `defrecord`.
- **R is narrowed to `DATA_FLOWS`.** R declares no type symbols at all — `setClass("Point", ...)`
  produces none, so an S4 fixture emits nothing either. Every `USES_TYPE` ever observed from R
  had resolved to a foreign type.
- The Erlang record is renamed `point` -> `coord` so the fixture stops manufacturing the edges.
  Erlang still emits `USES_TYPE`/`PARAM_TYPE` from `sum(#coord{...})`.

### What changed on this branch (`a69440cd`)

- `internal/sem/provider.go:135-154` — the whole `ooRelationSupport` block is now
  **byte-identical to #181's**, carrying the agreed value for all thirteen languages. Both
  PRs assigning the same text means merge order can no longer drop anything.
- `internal/sem/provider_test.go` — `TestCapabilityMatrixDeclaresTypeAndFlowRelations` keeps
  its assertion; only the fixture changes (record rename + Clojure type hint).
- `internal/sem/capability_relation_union_test.go` — new, also byte-identical on both branches.

### The new guard

`TestCapabilityRelationDeclarationsMatchIsolatedExtraction` asserts **equality** between
`ooRelationSupport[lang]` and what that language emits over a single-language fixture, rather
than the one-directional emitted-implies-declared check. Equality is what makes divergence
impossible: a narrowed entry fails on the relation the fixture still emits, a widened one
fails on the relation it cannot produce. One language per repository, so a foreign type
symbol can never justify an entry again.

### Verification

```
$ git merge-tree --write-tree 98c80088 d17230ee          -> exit 1  (before)
$ git merge-tree --write-tree a69440cd b0f15e4c          -> exit 0  (after)
$ git merge-tree --write-tree origin/main a69440cd       -> exit 0
main + #181 + #168 merged, go vet + gofmt -s -l          -> clean
  TestCapabilityMatrixDeclaresTypeAndFlowRelations            PASS
  TestCapabilityMatrixDeclaresDartErlangOCamlAndSQLRelations  PASS
  TestCapabilityMatrixCoversEmittedRelations                  PASS
  TestCapabilityRelationDeclarationsMatchIsolatedExtraction   PASS (13 subtests)
```

Mutation (each compiles, then the new test is run alone):

| mutation | result |
|---|---|
| `Erlang` back to `{USES_TYPE, PARAM_TYPE}` (this PR's old value) | **FAIL** — `Erlang declares [PARAM_TYPE USES_TYPE] but emits [DATA_FLOWS PARAM_TYPE USES_TYPE]` |
| `Erlang` back to `{DATA_FLOWS}` (#181's old value) | **FAIL** — `declares [DATA_FLOWS] but emits [DATA_FLOWS PARAM_TYPE USES_TYPE]` |
| `R` back to `{USES_TYPE, DATA_FLOWS}` (this PR's old value) | **FAIL** — `R declares [DATA_FLOWS USES_TYPE] but emits [DATA_FLOWS]` |
| `Julia` back to main's `{DATA_FLOWS}` | **FAIL** |
| restored | **PASS** |

### Still open

- **`name_only` cross-language type resolution is a real defect, not just a fixture smell.**
  A record in an Erlang file still produces `USES_TYPE` edges from unrelated R and Clojure
  functions. Narrowing R's declaration stops it being advertised; it does not stop the edges
  being emitted. Worth its own change; not touched here.
- **Your point 2 stands: this PR still conflicts with #165** in `provider.go`
  (`git merge-tree 978f29ad a69440cd` -> exit 1, same as before). Only the #181 collision is
  resolved here.

Nothing merged.

suhaanthayyil:
Closing out the **#168 ↔ #165 conflict** noted in review. It is real, but narrower than described, and needs no code change on either branch.

The report says the two conflict in `provider.go`. They do both edit `provider.go`, but that file **auto-merges cleanly**. The only conflict is in the test file:

```
$ git merge-tree --write-tree a69440cd 978f29ad
Auto-merging internal/sem/provider.go
Auto-merging internal/sem/provider_test.go
CONFLICT (content): Merge conflict in internal/sem/provider_test.go
```

One conflict, one file, and it is append/append: both PRs insert a new top-level test function at the same anchor (`provider_test.go:1771`, straight after the preceding test). #165 adds `TestInventoryOnlyFilesEmitNoCallRelations` (+63); #168 adds `TestCapabilityMatrixDeclaresTypeAndFlowRelations` and `TestCapabilityRelationDeclarationsMatchIsolatedExtraction` (+182). The two additions never reference each other and touch no shared symbol, so the resolution is literally "keep both", in either order.

I tried moving #168's additions to the end of the file to make the conflict disappear entirely, and **reverted it**: the tests embed Go raw string literals containing braces, and mechanically relocating them is churn with a real chance of splitting a literal — for a conflict whose correct resolution is unambiguous. Recording the attempt rather than leaving it as an implied to-do.

Same pattern as the #158 ↔ #164 note, for what it is worth: in both cases the production file merged cleanly and the conflict was two tests appended at one anchor. If that keeps recurring, appending new tests at end-of-file rather than after the nearest related one would remove the whole class.

suhaanthayyil:
## `usesTypeRelations` emits `USES_TYPE` for languages the matrix does not declare — confirmed, and it is six languages, not one

Following up on the open item. The class is real, and it has a mechanical cause worth naming.

### Why declaration and emission can disagree at all

`buildRelations` treats two families of relation differently:

- `internal/sem/provider.go:3906` / `:3914` — `resourceDependsOnRelations` and `configuresRelations` are fed `recordsByRelationSupport(recordsByFile, "RESOURCE_DEPENDS_ON" | "CONFIGURES")`, which filters the input through `languageSupportsRelation` (`provider.go:1431`) — i.e. through `ooRelationSupport`. For these, **the declaration gates the emission** and the two cannot drift apart.
- `internal/sem/provider.go:3881` — `usesTypeRelations(recordsByFile, ...)` takes the **unfiltered** map. Same for `signatureTypeRelations` (`:3890`), `fieldAccessRelations` (`:3686`) and the data-flow pass. The only guard is `spec.emits(...)`, which is a **profile** check, not a language check.

So a language reaches `USES_TYPE`/`PARAM_TYPE`/`RETURNS_TYPE`/`READS_FIELD`/`WRITES_FIELD`/`DATA_FLOWS` by having the right symbol shapes, never by being listed anywhere. `ooRelationSupport` can say something different from what the provider does and nothing fails.

### What was actually wrong

Probed one language per temp repository (single-language, per the cross-language caveat on this branch — `USES_TYPE` resolves signature identifiers by short name across the whole repo, so a polyglot fixture credits one language with another language's type edge). Six of the nineteen previously unmeasured languages diverged:

| language | divergence | reachable via |
|---|---|---|
| ClojureScript | `+USES_TYPE` — **absent from the map entirely** | `(defn point-sum [^Point p] ...)` — `.cljs` reads with the Clojure grammar |
| JavaScript | `+USES_TYPE` | `function pointSum(p = new Point(0, 0))` |
| Ruby | `+USES_TYPE` | `def point_sum(p = Point.new(0, 0))` |
| Groovy | `+RETURNS_TYPE`, `+WRITES_FIELD` | `Point make(int a, int b) { ...; q.x = a; ... }` |
| Swift | `+WRITES_FIELD` | `let q = Point(); q.x = a` |
| Zig | `-READS_FIELD` | **unreachable** — see below |

Every edge is same-language with `resolution=exact` and confidence 0.85, e.g.

```
USES_TYPE  JavaScript:pointSum(function)   -> JavaScript:Point(class)   res=exact conf=0.85
USES_TYPE  Ruby:point_sum(method)          -> Ruby:Point(class)         res=exact conf=0.85
USES_TYPE  ClojureScript:point-sum(fn)     -> ClojureScript:Point(class) res=exact conf=0.85
```

JavaScript and Ruby have no type-annotation syntax, which is presumably why nobody looked — but `usesTypeRelations` reads type names out of the signature **text**, and a default argument that constructs a local class puts one there. The relation is reachable without an annotation syntax existing at all.

Zig is the opposite sign and worth stating separately: it extracts the struct but **none of its fields** as symbols (confirmed in both the default and `full` profile — a Zig probe yields `[Point/struct read/function sum/method]`, no `x`/`y`), so `fieldsByContainer[container]` is empty and no `READS_FIELD` edge can ever resolve, however the body is written. An over-declared relation fails an agent exactly the way an under-declared one does: it feature-detects against `capabilities --json`, queries, and gets nothing back.

### `ACCESSES` deliberately left alone

It looked like a seventh divergence (Java and TypeScript declare it; no probe reached it). It is **not** one, and removing it would have been wrong. `fieldAccessRe` (`internal/sem/types.go:2701`) is ``(&?)\s*([A-Za-z_$][\w$]*)\s*(?:->|\.)\s*([A-Za-z_]\w*)``, so the `&` classifier is purely textual:

```java
int f(Point p, int m) { return m &p.x; }   // legal Java bitwise-and
```
```
ACCESSES  Java:f -> Java:x
```
Same for TypeScript (`return m &q.z;` → `ACCESSES h->z`). So `ACCESSES` *is* reachable in those languages — via a misclassification. Pinning it per language would enshrine that as a capability, so it is excluded from the pinned set with that reason recorded in the test. **Filed below as a separate open defect.**

### Fix

`internal/sem/provider.go:92-176` — six entries corrected, each with the construct that reaches it recorded inline.

### Test

`internal/sem/capability_relation_union_test.go`

- `TestCapabilityRelationDeclarationsMatchIsolatedExtraction` — probe map extended from 13 to **32 languages** (every language `supportsCallExtraction` admits). Equality over `genericRelationTypes` for all; full non-structural equality for probes marked `exhaustive`. Also asserts no relation leaves the probe's language, so a repository that stops being single-language fails loudly instead of quietly measuring the wrong thing.
- `TestCapabilityRelationProbesCoverEveryCallExtractionLanguage` — **new**. Derives the required language set from `supportsCallExtraction` itself rather than a second hand-maintained list, so a language cannot be added without a probe pinning what it emits. This is the guard against the original failure mode: coverage-by-coincidence.

Before the table fix, the extended test fails on exactly those six and passes the other 26.

### Mutation results

| # | mutation | grep proof | result |
|---|---|---|---|
| 1 | drop `USES_TYPE` from `JavaScript` | `100: "JavaScript": {"EXTENDS", "INHERITS", "HANDLES_GRAPHQL", ...}` | compiles; **FAIL** — `JavaScript declares [DATA_FLOWS] but emits [DATA_FLOWS USES_TYPE]` |
| 2 | re-add unreachable `READS_FIELD` to `Zig` | `174: "Zig": {..., "READS_FIELD", "DATA_FLOWS"}` | compiles; **FAIL** — `Zig declares [... READS_FIELD ...] but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE]` |
| 3 | delete the `ClojureScript` probe | `grep -c '"ClojureScript": {path:'` → `0` | compiles; **FAIL** — `no isolated relation probe for [ClojureScript]` |

Each restored → pass. Mutation 2 proves the over-declaration direction is caught, not just the under-declaration one.

### Behaviour diff vs `origin/main`

Built `entire-graph` from `origin/main` and from this branch, ran `snapshot` over a polyglot corpus carrying every affected relation (the corpus is non-vacuous — `origin/main` itself emits `ClojureScript USES_TYPE`, `JavaScript USES_TYPE`, `Ruby USES_TYPE`, `Groovy RETURNS_TYPE/WRITES_FIELD`, `Swift WRITES_FIELD`, which is the reproduction):

```
diff snap-main.ndjson snap-branch.ndjson  ->  IDENTICAL (109 records each)
```

Extraction is unchanged — none of the six relations is gated by `languageSupportsRelation`. Only `capabilities --json` moves, and only in the six declared entries. No other capability field changed. Golden NDJSON fixtures pass without regeneration.

`gofmt`/`go vet` clean; `internal/sem -run 'TestCapability|TestGolden|TestRelationSupport|TestProviderCapabilit|TestSnapshot'` green.

### Left open (new, separate)

`fieldAccessRe` classifies `&receiver.field` as address-of on a purely textual match, so `m &p.x` in Java/Kotlin/TypeScript — legal bitwise-and — yields a false `ACCESSES` edge. Reproduced above in Java and TypeScript. Not fixed here: it is an extraction defect, not a capability-matrix one, and the correct fix (gate address-of on languages that have the operator) belongs with the field-access pass.

ashtom:

**Verdict: changes required — 2 open findings, plus one disagreement with #181 left to settle**

Head moved `98c80088` → `b76e8e98`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### My earlier Erlang finding is fixed

`Erlang` is now `{"USES_TYPE", "PARAM_TYPE", "DATA_FLOWS"}` — exactly the union I measured — and
Julia, Dart, OCaml and SQL now agree with #181 too. Thank you.

### One disagreement with #181 remains, and it is the opposite direction from last time

The two maps now differ in exactly one place:

```
Zig   #168 = [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE]
      #181 = [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE]   <- main's current value
```

This PR **removes** `READS_FIELD` from Zig; #181 keeps it. They still conflict, so whichever merges
second overwrites the other.

I checked whether this is behavioural before calling it either way, and it is not:
`ooRelationSupport` feeds `languageSupportsRelation`, whose only callers gate
`RESOURCE_DEPENDS_ON` and `CONFIGURES` — not `READS_FIELD`. So this is a reporting-accuracy
question about `capabilities --json`, not a change to what is emitted.

On the evidence I have, **this PR is probably right**: a Zig fixture with struct field reads emitted
`[DEFINES PARAM_TYPE USES_TYPE]` and no `READS_FIELD`, and the golden fixture's only `READS_FIELD`
edge is Groovy's. Worth confirming deliberately, though — the guard test only fails on relations
emitted *without* a declaration, so it cannot catch an over-declaration, which means nothing
currently protects this either way.

`#165` also still conflicts with this PR.

```
01M17JPH7V90  Medium  open    current    internal/sem/provider.go:159  Objective-C still underdeclares USES_TYPE: a local `@interface Point` plus a C-s…
01M17JPH7HGY  Medium  open    current    internal/sem/provider.go:168  Swift can emit READS_FIELD through the same local-constructor inference used to …
```

### Verified vs. assumed

suhaanthayyil:
## Finding responses — both valid; the Objective-C one is larger than reported

Head is now `622caba9`.

---

### `internal/sem/provider.go:159` — Objective-C under-declares `USES_TYPE` — **VALID, and it under-declares three relations, not one**

Probed by building single-language repositories and reading the emitted relations, using the same isolated-extraction shape `TestCapabilityRelationDeclarationsMatchIsolatedExtraction` uses, so the result is comparable to what the declaration means. Every edge below was checked to resolve Objective-C → Objective-C, so nothing is credited by a cross-language short-name resolution.

| fixture shape | emitted (non-structural) |
|---|---|
| `Point *p;` local variable | *(none)* |
| `@property` in a header, no consumer | *(none)* |
| local assigned by message send, `return p.x` | *(none)* |
| `- (void)moveTo:(Point *)other` (method param) | `USES_TYPE` |
| same, split across `Point.h` / `Point.m` | `USES_TYPE` |
| `static void movePoint(Point *p)` (C function param) | `PARAM_TYPE`, `USES_TYPE` |
| `static Point *makePoint(NSInteger a)` (C return) | `RETURNS_TYPE`, `USES_TYPE` |

The gate is a **parameter or return type**, not a local declaration and not Objective-C method syntax — header presence is irrelevant. Objective-C compiles through the same C extraction path as C and C++ (the reason already recorded for C++ at `provider.go:143`), so it reaches C's declared set exactly: `{USES_TYPE, PARAM_TYPE, RETURNS_TYPE, DATA_FLOWS}`.

**Why the existing probe missed it:** the fixture declared no type of its own — only `NSInteger` scalars — so nothing but `DATA_FLOWS` could fire however the file was written. The probe now carries a local `@interface Point` plus C-style declarations that name it.

---

### `internal/sem/provider.go:168` — Swift can emit `READS_FIELD` — **VALID**

The local-constructor binding resolves **both** directions, not only the write. The existing probe wrote through the binding but read only through a parameter, which is what made `READS_FIELD` look unreachable rather than merely unexercised:

```swift
func make(_ a: Int, _ b: Int) -> Point { let q = Point(); q.x = a; return q }   // WRITES_FIELD  ✅ declared
func pointSum(_ p: Point) -> Int { return add(p.x, p.y) }                        // parameter — not resolved
func readPoint() -> Int { let q = Point(); return q.x }                          // READS_FIELD  ❌ undeclared
```

The genuine asymmetry is the **parameter** shape (`_ p: Point` is not a shape `parameterVarTypes` recognises), not the direction of the access. The comment now records that instead. No other undeclared relation appeared in either Swift shape.

---

### Fix

```
provider.go  "Objective-C": {"USES_TYPE", "PARAM_TYPE", "RETURNS_TYPE", "DATA_FLOWS"}
provider.go  "Swift":       {"USES_TYPE", "PARAM_TYPE", "RETURNS_TYPE", "READS_FIELD", "WRITES_FIELD", "DATA_FLOWS"}
```

plus the two probes in `internal/sem/capability_relation_union_test.go` extended to exercise what those declarations now claim. Both probes stay strictly single-language.

### Test — failing before, passing after

`TestCapabilityRelationDeclarationsMatchIsolatedExtraction` asserts exact equality in both directions, so the extended probes fail against the old declarations:

```
--- FAIL: …/Objective-C
    Objective-C declares [DATA_FLOWS] but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE] (non-structural relations)
--- FAIL: …/Swift
    Swift declares [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE WRITES_FIELD] but emits [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE WRITES_FIELD] (non-structural relations)
```

With the declarations corrected, the whole test passes, and so does every sibling capability test (`TestCapabilityMatrixCoversEmittedRelations`, `TestCapabilityMatrixDeclaresTypeAndFlowRelations`, `TestCapabilityRelationProbesCoverEveryCallExtractionLanguage`, `TestCapabilitiesReportRelationSupportPerLanguage`, `TestCapabilitiesAdvertiseExpandedLanguageSet`, `TestLanguageTiersClassifiesSemanticAndInventory`, `TestGraphQLBodyScanHonoursCapabilityTable`, `TestImportCapabilityMatchesScannerRegistry`) — 41 passed.

### Mutation verification

Reverted both declarations to their previous values, `grep`-confirmed the revert had actually landed (the first attempt did not, because gofmt had re-aligned the map and the literal replacement silently matched nothing — worth flagging, since that would have read as a clean "not caught"):

```
GREP AFTER MUTATION:
164:	"Objective-C": {"DATA_FLOWS"},
175:	"Swift": {"USES_TYPE", "PARAM_TYPE", "RETURNS_TYPE", "WRITES_FIELD", "DATA_FLOWS"},
--- test MUST FAIL ---
        Objective-C declares [DATA_FLOWS] but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE]
        Swift declares […WRITES_FIELD] but emits […READS_FIELD…WRITES_FIELD]
FAIL
```

Restored, `grep`-confirmed, test passes. `go vet` and `gofmt -l` clean.

### One thing left open, not claimed as a capability gap

While probing Objective-C, `SIMILAR_TO` fired for the *method-parameter* shapes (`- (void)moveTo:(Point *)other`) but not for the C-function shapes. It traces to symbol duplication: `@interface` and `@implementation` each produce a `Point` class symbol, and the method appears twice (qualified `Point.moveTo` and bare `moveTo`), which the similarity heuristic then matches. That looks like a symbol-identity bug rather than an undeclared relation, so it is not folded into this change — flagging it for a separate look.

suhaanthayyil:
## The Zig difference with #181 is settled: **this PR is on the correct side**

#181 has merged, so `main` now carries `READS_FIELD` for Zig and this PR removes it. That removal is right, and `main` is over-declaring until this lands. Full evidence is on #181; the short version:

- **Reporting-only, confirmed.** `languageSupportsRelation`'s only callers gate `RESOURCE_DEPENDS_ON` (`provider.go:3906`), `CONFIGURES` (`provider.go:3914`) and `HANDLES_GRAPHQL` (`types.go:379`). Nothing consults it for `READS_FIELD`.
- **Zig cannot reach the emitter.** `READS_FIELD` is emitted only at `provider.go:9476`, behind `fieldsByContainer[containerID][access.Field]` at `provider.go:9472`. That index holds only `Kind == "field"` symbols (`provider.go:2774-2778`), and Zig produces none — Zig struct members parse as `container_field`, which `fieldEntities` (`parser.go:4131-4136`) does not accept. `grep -rn container_field --include='*.go' internal/` is empty outside the vendored grammars.
- **Fixtures agree.** Seven Zig field-read shapes (signature-typed parameter, `self.field`, pointer receiver read and write, nested access, `@field`, optional chain, local `const`) on current `main` emit `[CONTAINS DEFINES PARAM_TYPE USES_TYPE]` — no `READS_FIELD`, and no `field`-kind Zig symbol at all.
- **The apparent Zig `READS_FIELD` was borrowed from another language.** A Zig file with no local `Point` beside a Go `type Point struct { x float64 }` produces, on current `main`:
  ```
  READS_FIELD  from-lang=Zig  readX(a.zig) -> x(p.go, lang=Go)
  ```
  The short-name type lookup at `provider.go:9459` resolves the Zig signature onto the Go struct. That is the only path, and it is a cross-language false edge — the class #184 addresses.

Mutation-checked on this branch: putting `READS_FIELD` back fails this PR's own equality guard with

```
Zig declares [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE]
     but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE]
```

(mutation compiled, grep-confirmed applied, restores green). No change needed here.

## Zig is not the only divergence left

Comparing `Capabilities().RelationSupportByLanguage` between this branch and current `main`, five more languages differ, with this PR declaring strictly more in each:

| language | this PR adds vs `main` |
|---|---|
| ClojureScript | `USES_TYPE` (no entry on `main`) |
| Groovy | `RETURNS_TYPE`, `WRITES_FIELD` |
| JavaScript | `USES_TYPE` |
| Ruby | `USES_TYPE` |
| Swift | `WRITES_FIELD` |

Each is pinned by a probe here. Since `main`'s equality guard covers only its 14-entry probe map against 32 call-extraction languages, this PR's `capabilityRelationProbes` plus `TestCapabilityRelationProbesCoverEveryCallExtractionLanguage` is the version that closes the gap — it fails when any language `supportsCallExtraction` admits has no probe, which is what stops this drifting again.

Worth confirming those five the same way before merge, so the same "whichever lands second silently overwrites the other" problem does not repeat one round later.

suhaanthayyil:
### Correction to the divergence table above

That table was computed against `b76e8e98`; this branch has since moved to `622caba9`, which also touched Objective-C and Swift. Re-derived from the two `ooRelationSupport` literals directly — `origin/main` at `7b3252aa` vs `622caba9`:

| language | this PR declares beyond `main` | `main` declares beyond this PR |
|---|---|---|
| ClojureScript | `USES_TYPE` (no entry at all on `main`) | — |
| Groovy | `RETURNS_TYPE`, `WRITES_FIELD` | — |
| JavaScript | `USES_TYPE` | — |
| Objective-C | `PARAM_TYPE`, `RETURNS_TYPE`, `USES_TYPE` | — |
| Ruby | `USES_TYPE` | — |
| Swift | `READS_FIELD`, `WRITES_FIELD` | — |
| Zig | — | `READS_FIELD` |

Seven languages, not six, and Swift adds `READS_FIELD` as well as `WRITES_FIELD`. Zig remains the only entry where `main` declares more, and it is the only one where `main` is wrong. The same correction applies to the table in the note on #181, which was derived from the same older head.

suhaanthayyil:
### This no longer merges onto `main`

`git merge-tree origin/main 622caba9` against `main` at `7b3252aa`:

```
CONFLICT (add/add): internal/sem/capability_relation_union_test.go
CONFLICT (content): internal/sem/provider.go
CONFLICT (content): internal/sem/provider_test.go
```

The `capability_relation_union_test.go` collision is add/add — `main` now carries #181's version of that same file. Several `provider_test.go` hunks split **inside raw-string fixture literals**, so a both-sides union does not even parse; this needs the author's rebase rather than a mechanical resolution, and no resolution was invented here.

This is the failure both PRs' threads predicted: whichever lands second overwrites the other. #181 landing first is exactly why `main` kept the wrong Zig value.

To keep the decisive question separate from the rebase, the Zig change was tested **directly on current `main`** instead:

- `main` as-is: whole capability guard suite green **while over-declaring Zig** — `capabilityRelationProbes` there has 13 entries and no Zig probe, and `TestCapabilityMatrixCoversEmittedRelations` only checks emitted ⇒ declared.
- `main` + this PR's Zig value + a Zig probe: green, `gofmt`/`go vet` clean, no file under `testdata/` changed.
- Mutation, putting `main`'s current value back (grep-confirmed applied, compiles): `capability_relation_union_test.go:304: Zig declares [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE] but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE]` → FAIL. Restored → green.

One more thing worth surfacing, because it is the cleanest evidence the Zig removal is not a blanket "the field pass does not work for smaller languages": `622caba9` **adds** `READS_FIELD` to **Swift** while removing it from Zig. Same pass, opposite outcome, each measured — Swift resolves a field through a local constructor binding; Zig has no `field`-kind symbol for the pass to look up at all.

Until this lands, `main` ships a wrong Zig capability and under-declares six other languages, so anything feature-detecting against `capabilities --json` gets the wrong answer in seven places.

ashtom:

`main` has moved to `7b3252aa` — **#181 has merged**, which changes what this PR is.

**Verdict: the remaining fix is correct and still needed, but its tests no longer prove it**

### The Zig question is settled by #181 landing

#181 merged with `Zig` keeping `READS_FIELD`. This PR still removes it. That is now a change against
`main` rather than a disagreement between two PRs — worth stating deliberately in the PR body,
because the guard test only fails on relations emitted *without* a declaration and so cannot catch
an over-declaration either way.

### The tests are now non-discriminating (§3)

I reverted only `internal/sem/provider.go` on the merged tree, kept every test this PR adds, and
re-ran them:

```
own tests WITH the change     : PASS
own tests WITHOUT the change  : PASS   <-- no discriminating power
```

Before #181 landed these tests failed without the change. They pass now because #181 supplied most of
the declarations they assert. The delta this PR still carries — `JavaScript` and `Ruby` gaining
`USES_TYPE`, a new `ClojureScript` key, `Groovy` gaining `RETURNS_TYPE`/`WRITES_FIELD`, `Swift`
gaining `WRITES_FIELD`, `Zig` losing `READS_FIELD` — is **not covered by any test in the PR**.

### That delta is real, though — I verified the JavaScript and Ruby claims

Your comment says the `USES_TYPE` pass reads type names out of signature *text*, so a default
argument that constructs a local class reaches it without an annotation syntax existing. That is
correct on current `main`:

```go
// app.js:  function draw(p = new Point()) { return p.x; }
// app.rb:  def draw(p = Point.new) ... end
```

```
JavaScript   emits [CONSTRUCTS CONTAINS DEFINES USES_TYPE]   USES_TYPE=true
Ruby         emits [CALLS CONTAINS DEFINES USES_TYPE]        USES_TYPE=true
```

So `main` under-declares both today, and this PR fixes it. **Please add a test that fails without
this change** — a fixture per newly-declared language, asserted the way your isolation test already
works, would restore the property the revert check is looking for.

### Gate

Merges clean; `go vet`/`gofmt` pass; all touched packages pass on the merged tree. **No open, current
trail findings.**

suhaanthayyil:
## Re-checked on the merged tree: the non-discrimination is **half** true, and it is the wrong half

Merged current `origin/main` (`122333d8`), reverted **only** `internal/sem/provider.go` to
`origin/main`, kept every test this PR adds, re-ran. Per test, not per file:

```
$ git checkout origin/main -- internal/sem/provider.go
$ go test ./internal/sem/ -run 'TestCapabilityMatrixDeclaresTypeAndFlowRelations$' -count=1
ok      github.com/entireio/entire-graph/internal/sem            <- PASS, no discriminating power

$ go test ./internal/sem/ -run 'TestCapabilityMatrixCoversEmittedRelations$' -count=1
ok      github.com/entireio/entire-graph/internal/sem            <- PASS (pre-existing guard)

$ go test ./internal/sem/ -run 'TestCapabilityRelationDeclarationsMatchIsolatedExtraction' -count=1 -v
    capability_relation_union_test.go:773: Groovy declares [DATA_FLOWS PARAM_TYPE READS_FIELD USES_TYPE] but emits [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE WRITES_FIELD] (non-structural relations)
    capability_relation_union_test.go:773: Zig declares [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE] but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE] (non-structural relations)
    capability_relation_union_test.go:773: Ruby declares [DATA_FLOWS INHERITS] but emits [DATA_FLOWS INHERITS USES_TYPE] (non-structural relations)
    capability_relation_union_test.go:773: Swift declares [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE] but emits [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE WRITES_FIELD] (non-structural relations)
    capability_relation_union_test.go:773: JavaScript declares [DATA_FLOWS] but emits [DATA_FLOWS USES_TYPE] (generic relations)
    capability_relation_union_test.go:773: ClojureScript declares [] but emits [USES_TYPE] (non-structural relations)
    capability_relation_union_test.go:773: Objective-C declares [DATA_FLOWS] but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE] (non-structural relations)
--- FAIL: TestCapabilityRelationDeclarationsMatchIsolatedExtraction (0.00s)
    --- FAIL: .../Groovy      --- FAIL: .../Zig           --- FAIL: .../Ruby
    --- FAIL: .../Swift       --- FAIL: .../JavaScript    --- FAIL: .../ClojureScript
    --- FAIL: .../Objective-C
FAIL
```

So: `TestCapabilityMatrixDeclaresTypeAndFlowRelations` (provider_test.go, polyglot,
emitted⇒declared only) **is** non-discriminating now — #181 supplied what it asserted, exactly
as you say. But `TestCapabilityRelationDeclarationsMatchIsolatedExtraction`
(capability_relation_union_test.go) fails without the change on **7 of 7** delta languages, one
subtest each, one temp repo each. It is per-language, isolated, and an **equality** check, so it
already fails in both directions. The revert check you were looking for is in the PR; running the
whole suite as one unit hid it behind the weaker test.

The isolation you asked about is the thing that makes those seven readable one at a time —
`USES_TYPE` resolves a signature identifier against every type symbol in the repository by short
name, so a polyglot fixture credits one language with another language's edge. Every probe is a
single-language `t.TempDir()`, and the test additionally fails if any edge crosses out of the
probe's language.

## The delta, verified against `main` rather than taken on trust

Your list was right on five and short on two. `git diff origin/main -- internal/sem/provider.go`
touches the `ooRelationSupport` map and nothing else:

| language | `main` declares | this PR declares | delta |
|---|---|---|---|
| JavaScript | EXTENDS, INHERITS, HANDLES_GRAPHQL, HANDLES_TRPC, ASYNC_CALLS, DATA_FLOWS | + USES_TYPE | add USES_TYPE |
| Ruby | INHERITS, DATA_FLOWS | + USES_TYPE | add USES_TYPE |
| ClojureScript | *(no key at all)* | USES_TYPE | new key |
| Groovy | USES_TYPE, PARAM_TYPE, READS_FIELD, DATA_FLOWS | + RETURNS_TYPE, WRITES_FIELD | add ×2 |
| **Objective-C** | DATA_FLOWS | + USES_TYPE, PARAM_TYPE, RETURNS_TYPE | **add ×3 — not in your list** |
| **Swift** | USES_TYPE, PARAM_TYPE, RETURNS_TYPE, DATA_FLOWS | + READS_FIELD, WRITES_FIELD | **add ×2 — you listed WRITES_FIELD only** |
| Zig | …, **READS_FIELD**, … | *(READS_FIELD dropped)* | **REMOVE** |

Objective-C is the same shape as your JavaScript/Ruby finding: it compiles through the shared C
extraction path, so a parameter or return naming a repository type reaches the signature-type pass
the way C and C++ do. It looked like `DATA_FLOWS`-only because the earlier probe declared no type
of its own, so nothing else could fire however the file was written. Swift resolves a field
through a **local constructor binding** in both directions (`let q = Point(); q.x = a` writes,
`return q.x` reads); what it does not resolve is a field reached through a *parameter*, and
reading only through a parameter is what made `READS_FIELD` look unreachable rather than merely
unexercised.

## Zig: re-verified, and it is stated in the PR body now

Your point stands and is the important one — the existing guard fails only on a relation emitted
*without* a declaration, so it cannot catch an over-declaration in either direction. The evidence
for removing `Zig: READS_FIELD` is mechanical, not fixture-shaped:

1. `fieldAccessRelations` is the **sole** producer of `READS_FIELD`/`WRITES_FIELD` — the only place
   either string is constructed outside the declaration tables.
2. It resolves every edge through `fieldsByContainer[containerID][field]`.
3. `fieldsByContainer` is populated under exactly one condition: `symbol.Kind == "field"`.
4. `fieldEntities` admits only `field_declaration`, `public_field_definition`, `field_definition`,
   `property_signature`, `property_declaration`. tree-sitter-zig names a struct member
   `container_field` — 63 occurrences in the vendored `grammars/zig/parser.c`, and
   `grep -rn container_field --include='*.go' . | wc -l` → **0**.
5. Extracting the probe: Zig yields `struct Point`, `function add`, `function pointSum`,
   `function make`, and **zero** symbols of kind `field`.

## Added: the missing direction, as its own named property

Two guards, pushed as `cac14fbd`:

- **`TestCapabilityDeclaredFieldRelationsRequireAFieldSymbol`** — a language may not declare
  `READS_FIELD`/`WRITES_FIELD` unless its extractor produces a symbol of kind `field`, per
  language, over the isolated probes. Structural rather than incidental: the equality test catches
  Zig too, but only because that probe body happens to contain a field access, so editing the
  probe would silently drop the guard. This one holds whatever the body says.
- **`TestCapabilityReportCarriesEveryDeclaredRelation`** — every `ooRelationSupport` key reaches
  `capabilities --json`. `relationSupportByLanguage` builds the report by walking the language
  *specs* and folding the map in per language, so a key naming a language no spec produces is
  dropped without a word: the declaration reads as authoritative while the contract never carries
  it. A brand-new key — `ClojureScript` here — is precisely that shape.

## Mutation results

| mutation | result |
|---|---|
| revert `provider.go` to `origin/main`, keep every test | **FAIL** — 7/7 delta languages, named above |
| restore | **PASS** — 33 subtests |
| re-add `READS_FIELD` to `Zig` only (`grep` confirmed applied at provider.go:181) | **FAIL** — `…RequireAFieldSymbol/Zig`: *declares […READS_FIELD…] but extracts no symbol of kind "field" from src/flow.zig* **and** `…MatchIsolatedExtraction/Zig` |
| add `"Nonesuch": {"USES_TYPE"}` (a language no spec produces) | **FAIL** — `…CarriesEveryDeclaredRelation`: *declares [USES_TYPE] for Nonesuch but capabilities --json reports no such language* |
| first attempt at that mutation used `"Objective-C++"` | **PASS — my mistake, reported straight**: `.mm` is a real inventory-only spec (`inventory_languages.go:79`), so the key is legitimately carried. Redone with a fictional language. |
| with the change | **PASS** — `-run TestCapability`, 47 subtests |

`gofmt -l internal/` clean, `go vet ./internal/sem/` clean.

Full package on the pushed tree: `go test ./internal/sem/ -count=1` `ok  github.com/entireio/entire-graph/internal/sem  81.570s`.

## One thing I did not change

`TestCapabilityMatrixDeclaresTypeAndFlowRelations` stays. It is now a strictly weaker duplicate of
the isolated equality test and it no longer discriminates — but it is a passing one-directional
check over a polyglot repo, and removing a green test to tidy up the report is not something to do
on the way past. Flagging it rather than deleting it.

ashtom (approved):
Approved after syncing with current main. The refreshed Entire review has zero current findings; required CI is running on this exact head.

PR #169, opened 2026-08-28
fix(doctor): the provider never tells a consumer which repo_key or schema it will use, so the seam can only be checked after the snapshot

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/128

## Impact

`graph doctor --json` is the preflight consumers run before building a snapshot, but it did not advertise the `repo_key` or `schema_version` the snapshot would use. It could also preserve a relative `repo_root` even though snapshot construction emits an absolute path, making the preflight disagree with the artifact it predicts.

## Fix

- Export `sem.RepoKey` as the provider's published namespace rule.
- Add `repo_key` and `schema_version` to `doctor --json`.
- Normalize the resolved repository path exactly as snapshot construction does, so doctor and the snapshot agree on `repo_root`, `repo_key`, and `schema_version`.
- Document and pin the established remote precedence: the last configured `remote.origin.url` is checked first, then non-origin URLs in Git config order; the first supported GitHub form supplies `gh/<owner>/<name>`.
- Pin relative and uncleaned repository-path behavior.
- Harden the fallback reader inherited from current `main`: canonical paths are translation-only, the repository descriptor is identity-checked, child directories open only relative to that descriptor, and a reader without a retained root descriptor refuses all reads.

## Contract boundaries

`repo_key` is a symbol-ID namespace, not sufficient repository identity. Repositories without a supported GitHub remote use `local/<basename>`, which can collide. Consumers must pair the key with `repo_root`, `commit`, and `tree`, or their own storage identity.

Both persistent cache families include the absolute repository path. Tests create unrelated repositories with identical `repo_key`, tree, and commit and verify that neither the search cache nor the provider-record cache replays a foreign repository.

Symbol-ID fields remain unescaped. A repository name or POSIX file path may contain `:`, so consumers must not split IDs positionally; use whole-ID comparisons, defined anchored parsing, and the record's own `file_path`.

The strict fallback policy covers pathname and symlink replacement. It does not claim to defend against privileged mount-namespace or bind-mount replacement of an already-open filesystem object.

## Findings resolved

- `01M13CCRB7208ATNXVYBJ0N8HA` (medium): `RepoKey` now normalizes relative and uncleaned paths before deriving the namespace.
- `01M17S6Q0ZWYZ4VR4QS7X52PAN` (medium): doctor now reports the same absolute root as snapshot construction, and the anti-drift test compares doctor with a real snapshot built from `.`.
- `01M1H8VSW9A2CP7PB85GEPBGST` (high): fresh merged-tree Ubuntu CI proved an ABA directory/symlink swap could return outside-repository content. The child is now opened only through an identity-checked repository `os.Root`; rootless readers fail closed; deterministic swap, repeated stress, and race tests cover the fix.

Independent re-review also closed gaps in remote-precedence vectors, direct `RepoKey(".")` coverage, both persistent cache-family collision tests, and containment source guards. Three final independent reviews found no remaining issue.

## Verification

Integrated current `main` in `f38a99e1`; contract review fixes are in `c458f7be`; the merged-tree containment fix is in `ccbbd1dd`.

Passed on the final merged tree:

- `gofmt -s -l .`
- `go vet ./...`
- `go build -o entire-graph ./cmd/entire-graph`
- `go test -timeout 30m ./...`
- `go test -race -timeout 30m ./...` (`internal/sem`: 740.824s)
- `sh scripts/entire-graph-statusline_test.sh` (151 passed)
- containment stress test repeated 100 times and targeted race repetitions
- fresh GitHub CI on Ubuntu, macOS, and Windows (Windows: 26m28s); only the human-approval Entire Gate remains


ashtom:

**Verdict: clean**

- Live PR head `1491e49f`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `1491e49f` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

### Independent review state

suhaanthayyil:
Thanks. Staleness unchanged — `main` still `bb41f223`, still `13 behind / 1 ahead`.

The claim worth checking on this PR is not that the tests pass, it is that the advertised value is *actually* the one the snapshot will carry. The unit test asserts that in-process; I checked it out-of-process with built binaries on four real repository shapes, comparing `doctor --json` against the header of a real snapshot:

```
repo shape                doctor repo_key              snapshot header repo_key
no remote                 local/corpus                 local/corpus              MATCH
gitlab https remote       local/hcl                    local/hcl                 MATCH
github ssh, mixed case    gh/Acme/Widget               gh/Acme/Widget            MATCH
github https              gh/entireio/entire-graph     gh/entireio/entire-graph  MATCH
```

`schema_version` likewise: doctor says `1.1`, the header says `1.1`.

The additive claim also checks out at the wire level — the JSON gains keys and loses none:

```
main : ['environment','no_egress','phase_1_local_only','provider','repo_root','version']
#169 : ['environment','no_egress','phase_1_local_only','provider','repo_key','repo_root','schema_version','version']
```

### One note on the contract this publishes

`RepoKey`'s doc comment gives the consumer's reason for needing it: *"Consumers (entire-brain) must be able to predict this value to tell a snapshot of THIS repository apart from a foreign one."* For the `gh/<owner>/<name>` branch that holds. For the `local/<basename>` fallback it does not — the basename is not a namespace:

```
https://gitlab.com/orgA/ledger.git -> repo_key=local/ledger
                                       symbol id=local/ledger:Go:m.go:function:Add
https://gitlab.com/orgB/ledger.git -> repo_key=local/ledger
                                       symbol id=local/ledger:Go:m.go:function:Add
```

Two unrelated repositories, byte-identical symbol IDs. That is pre-existing behaviour and the doc comment states the rule accurately, so it is not a defect in this PR — but this PR is what turns the rule into a published contract, so it is the right place to say it out loud: for any non-GitHub remote the advertised key cannot deliver the distinguishing property, and a consumer that treats `repo_key` as an identity check needs to know `local/*` is advisory. Worth a sentence on `RepoKey`; not a blocker.

Usefully, the handshake makes that *visible* up front for the first time — the consumer now sees `local/ledger` before spending the snapshot rather than after, which is the whole point of the PR.

suhaanthayyil:
## `local/<basename>` collides. Measured what it costs before touching the key.

Raised on this PR: `repo_key` for a local repository is `local/<basename>`, which does not
distinguish same-named repositories. The defect predates this PR, but this PR is what
*publishes* the key as a handshake, so the question is now load-bearing. The doc comment
added here says a consumer predicts the key "to tell a snapshot of THIS repository apart
from a foreign one." That sentence is false for half the rule.

### The collision is real, and the handshake gate does false-pass

Two unrelated repositories, different content, different symbols, same basename:

```
$ (cd /tmp/.../collide/x/tools && entire-graph doctor --json) | jq -r .repo_key
local/tools
$ (cd /tmp/.../collide/y/tools && entire-graph doctor --json) | jq -r .repo_key
local/tools

$ entire-graph snapshot --repo .../collide/y/tools --format ndjson | head -1 | jq -r .repo_key
local/tools

consumer gate  header.repo_key == advertised  ->  True   (B is a FOREIGN repo)
symbols B would contribute under A's key:
  local/tools:Go:beta.go:function:BetaOnlySecretWidget
```

Note the failing shapes are broader than the report: `gh/org-a/tools` vs `gh/org-b/tools`
is fine — the github half of the rule is unique. It is everything *else* that collapses:
no remote, gitlab, bitbucket, self-hosted. `local/<basename>` is a directory name.

### It does not corrupt or leak anything today

Four experiments, one shared cache dir throughout.

**1. No shared entry, no eviction, no overwrite.** Indexing A then B writes two entries:

```
$ entire-graph index --repo .../x/tools --cache-dir $C   # entries: 1
$ entire-graph index --repo .../y/tools --cache-dir $C   # entries: 2
CACHE/search/search-snapshot-v12/fd37fa51...json.gz
CACHE/search/search-snapshot-v12/2c7efeb8...json.gz
```

Same for the other cache family via `snapshot`: `CACHE/records/provider-records-v8/{5252094d,257da052}...`.
Those are the only two families (`newCacheEntry` call sites: `search_cache.go:136,201,246,381`,
`records_cache.go:136`).

**2. Not even at an identical tree.** The adversarial case — same key, same tree hash, so
every other keyed term agrees:

```
trees: beef6992301dd4f9f921beff64aca6dc8dcd29ff vs beef6992301dd4f9f921beff64aca6dc8dcd29ff
cache entries after p=1 after q=2   (identical tree, identical repo_key, different path)
```

Because both key functions hash the absolute repository path beside the repo key —
`searchSnapshotKey` `internal/sem/search_cache.go:752`, `providerRecordsKey`
`internal/sem/records_cache.go:62`.

**3. No cross-repo leakage.** Each repository's only symbol is invisible to the other:

```
$ entire-graph search --repo .../x/tools --head --cache-dir $C --query BetaOnlySecretWidget
results: ['local/tools:Go:alpha.go:function:AlphaOnlySecretWidget']
$ entire-graph search --repo .../y/tools --head --cache-dir $C --query AlphaOnlySecretWidget
results: ['local/tools:Go:beta.go:function:BetaOnlySecretWidget']
```

**4. The key never reaches a filesystem path.** Cache paths are `family/version/<sha256hex>.json.gz`
and `newCacheEntry` rejects anything else (`cache_entry.go:30`, `validSHA256Hex`). The `/` in
`local/tools` is inert; there is no path-safety exposure to fix.

### The existing repo-key-mismatch machinery MISSES this case — and it does not matter

`validCachedSearchSnapshot` (`search_cache.go:807`), `selectiveSearchSnapshotFromFull`
(`:535`), and the records `Store` provenance check (`records_cache.go:181`) all compare
`RepoKey` for equality. Two colliding keys compare equal, so every one of them passes. They
never get the chance to matter, because the `repository-path` term in the cache key means a
colliding entry is never looked up in the first place. They are defence in depth; the path
is the discriminator. ADR 0002 already says as much — it kept the checkout path in the key
and listed identity-only keying as a change it explicitly did not take.

### The consumer is not affected either

`entire-brain` never uses this string as a storage key. Its own key
(`internal/cli/env.go:344-353`) is `local/<base>-<sha256(abs path)[:12]>`, and every durable
store hangs off it — facts dir, semantic snapshots, pattern sqlite, `gitmeta.git`, workspace
members. Its own commit states the invariant: *"the hash suffix is load-bearing: two
checkouts sharing a basename must not collide onto one brain."* Its on-disk gate compares
against that path-hashed key, so a colliding provider key can never pass it — the shipped
check is over-strict, not under-strict, which is the *other* seam bug this PR exists to fix.
The only place a `local/<basename>` gate is admitted is unmerged brain work, and there the
bundle manifest still carries the path-hashed key and rejects first.

### Decision: fix the contract, not the key

Re-keying `local/*` — a path hash, or extending the remote rule to every host — would:

- rewrite **every symbol ID in every non-github repository** (`repoKey:language:path:kind:name`),
  orphaning anything downstream keyed by ID;
- break `TestRepoKeyContractGoldenVectors`, which is asserted on *both* sides of the seam,
  while brain has in-flight branches built against the current vectors;
- for a path hash, make the key **machine-local**, so a gitlab or self-hosted repository's
  symbol IDs would stop being portable between a laptop and CI.

That is a real regression traded for a collision that no measurement here shows corrupting
anything. So the key is unchanged, and the sentence that was wrong is now right.

`e3850959`:

- `internal/sem/provider.go` — `RepoKey` states the boundary: it is a **namespace, not a
  repository identity**; only the `gh/` half is globally unique; `repo_key` mismatch proves
  a foreign snapshot, `repo_key` match does not prove a native one; `repo_root` (already in
  the same doctor report) is the discriminator, which is exactly what this provider's cache
  keys and brain's storage key both already do.
- `docs/snapshot-format.md` — the same boundary where a consumer reads the header format.
- `TestRepoKeyLocalIsNotGloballyUnique` — pins the collision, and pins that two github owners
  do *not* collide. The doc cannot quietly start claiming uniqueness again.
- `TestCollidingRepoKeysDoNotShareCacheEntries` — the guard that keeps the collision inert:
  same key, identical tree, one cache dir, two entries, each reporting its own `repo_root`.

That second test is not decorative. Dropping the path term — the change ADR 0002 names as
possible future work — produces this today:

```
MUTANT: index p  ->  repo_root=.../same/p/tools  hit=False
MUTANT: index q  ->  repo_root=.../same/p/tools  hit=True     <== q reports P's root
MUTANT: search q ->  repo_root=.../same/p/tools
entries: 1
```

A foreign snapshot handed to a consumer under a matching `repo_key`. The test stops that
landing while `local/<basename>` can collide.

### Verification

Mutation-verified in both directions, each grep-confirmed applied and compiling:

| mutation | result |
|---|---|
| drop `repository-path` from `searchSnapshotKey` | `TestCollidingRepoKeysDoNotShareCacheEntries` FAILS — *"reported a cache hit; a colliding repository's entry was served"* |
| give `repoKey` a path-hash suffix | `TestRepoKeyLocalIsNotGloballyUnique` FAILS — *"local repo keys = "local/tools-MUT2aecae269" and "local/tools-MUT217578f9a""* |
| restore both | all pass |

`go test ./internal/cli` green (22.8s), `go test ./internal/sem -run 'RepoKey|SearchSnapshotKey|CacheKey'` green,
`gofmt`/`go vet` clean.

### Left open

- The provider still cannot reproduce brain's storage key, and brain's shipped on-disk gate
  still rejects every non-github repository. That is brain's side of the seam and is what
  its unmerged `audit/job1-repokey` addresses; this PR's `doctor` handshake is the input it
  needs.
- A repository directory name containing `:` produces an ambiguous-looking stable ID
  (`local/to:ols:Go:w.go:function:WeirdOne`). Harmless here — IDs are compared as opaque
  strings, never split — but a consumer that parses one by `:` would mis-split it. Separate
  issue, not touched.

suhaanthayyil:
## `:` in a repo directory name — reproduced, enumerated, and NOT inert at the seam

I previously called this inert. Inside `entire-graph` that holds and is now proven. One layer down it does not: `entire-brain` has a positional ID parser that mis-fires, and the reachable case does not even need a colon in the repo directory name.

### Repro

```
$ mkdir 'weird:name' && cd 'weird:name' && git init -q
$ entire graph symbols --repo . --format ndjson
{"repo_key":"local/weird:name", ...}
{"record_type":"symbol","id":"local/weird:name:Go:main.go:method:Mem.Get", ...}
{"record_type":"symbol","id":"local/weird:name:Python:od:d/mod.py:class:Cache", ...}
```

macOS/APFS and Linux both accept `:` at the syscall level; Windows does not. The second ID is the important one: a file at `od:d/mod.py` injects a colon of its own, so the ID splits into **seven** fields, not five. `symbols`, `edges`, `neighbors`, `def`, `impact` and `search` all resolve these IDs end-to-end.

### Consumer enumeration — `entire-graph`

| site | parses | how | reachable with a `:` | verdict |
|---|---|---|---|---|
| `internal/sem/provider.go:6511` `goMethodSymbolByID` | symbol ID → method qualified name | `LastIndex(id, ":method:")`, then re-checks `candidate.ID == id` | yes | **safe** — marker is found from the right and the whole ID is re-verified |
| `internal/sem/provider.go:9614` `enclosingTypeShortName` | container ID → type short name | `LastIndex(id, ":")` | yes | **safe** — reads the last field |
| `internal/sem/provider.go:9776` `externalParts` | `external:<kind>:<value>` | `Cut` at the FIRST `:`, prefix-gated at both call sites (`provider.go:1134`, `search_cache.go:612`) | no — external IDs carry no repo key | **safe**; also why `external:import:std::collections::HashMap` already works |
| `internal/cli/def.go:515` `defExternalTypeName` | external endpoint ID → type name | `LastIndexByte` | only via external `ToID` | **safe** |
| `internal/cli/root.go:582` `idMatches` | `--to`/`--from` selector | `id == sel` or `HasSuffix(id, ":"+sel)` | yes | **safe** — exact or right-anchored |
| `internal/cli/symbolref.go:49` `parseSymbolRef` | `--symbol` value | `LastIndexByte`, then numeric tail AND must name a snapshot file | yes | **safe** — a full ID has a non-numeric tail; `od:d/mod.py:2` still addresses correctly |
| `internal/cli/symbolref.go:193` `resolveFocusSymbolsByID` | `--symbol` = full ID | `==` | yes | **safe** |
| `internal/sem/compact_snapshot.go:816` `scipSymbolPackage` | repo key → SCIP package component | SCIP components are space-delimited; only spaces are escaped | yes | **safe** — `:` is not a SCIP delimiter |
| `internal/sem/cache_key.go:11` `writeCacheKeyField` (via `search_cache.go:753`) | repo key → cache key | length-prefixed field, never delimiter-joined | yes | **safe** |
| `records_cache.go:182`, `search.go:1880,2148`, `search_cache.go:410,535,807` | repo key identity | `==` | yes | **safe** |
| `internal/sem/golden_test.go:592` | `Split(FromID, ":")` → `parts[1]` as language | unbounded, positional | no — fixture keys are `local/<fixture-dir>` | mis-parses in principle, **test-only and unreachable** |

Negative results: no `SplitN` on `:` anywhere; no regexp parses a symbol ID; no consumer in TS/JS/Python/shell/jq; the repo key never becomes a filesystem path component.

### Consumer enumeration — `entire-brain` (read-only inspection)

Ingest path is `entire graph snapshot --format ndjson` shelled out from `internal/cli/semantic_stream.go:371`; IDs are stored verbatim in SQLite.

| site | parses | how | verdict |
|---|---|---|---|
| **`internal/cli/semantic.go:1147` `semanticEndpointPath`** | relation endpoint ID → repo-relative file path | `parts := strings.Split(id, ":")`, guard `len(parts) < 5`, **`return parts[2]`** | **BROKEN — silent mis-attribution** |
| `internal/cli/semantic_stream.go:301,304` | callers, on every relation at ingest | applies `.brainignore` to the returned path | inherits it: relations get wrongly kept or wrongly redacted |
| `internal/cli/semantic_graph.go:1481` `graphFilePathFromID` | file node ID → path | `Cut(id, ":file:")` | safe |
| `internal/cli/semantic.go:3210` `externalBoundaryRecord` | external ID | first-colon only, prefix-gated | safe |
| `workspace.go` (849–944, 1174–1270), `semantic_graph.go:1504`, `semantic.go:1144` | endpoint classification | `HasPrefix`/SQL `LIKE 'external:%'` | safe |
| `internal/factgitmeta/backend.go:302` | brain's own key | hex-encoded before joining | safe |

`semanticEndpointPath`'s own comment states the premise that fails: *"the repo key, language, and kind segments never contain `:`, so the path is always the third field."* The first clause is true; the conclusion is not, because the **path** can contain one.

**The reachable case needs no colon in the repo directory name.** For an ordinary repo key, `local/x:Python:od:d/mod.py:class:Cache` gives `parts[2] == "od"` — a path that does not exist, matched against the ignore set. Nothing gates that.

One thing I could not settle from outside: `validateLiveSemanticHeader` (`semantic.go:976`) compares the snapshot header's `repo_key` against `storage.Key`, but `repoStorageKey` (`env.go:190`) returns `local/<sanitised-base>-<12 hex of sha256(abs dir)>` for a no-remote repo while `entire-graph` emits `local/<basename>`. Those never match for *any* local repo, colon or not, so I can't claim the gate shields the repo-key half — and it is irrelevant to the path case either way. Worth a look on the brain side.

### Decision

**`entire-graph`: inert — proven, not asserted.** Eleven parsers enumerated; every product one compares a whole ID, anchors on the last separator, or is gated to `external:` whose kind field cannot hold a `:`.

**I did not sanitise the repo key**, against the obvious first instinct, because it would be a false closure: the file path injects colons through the same joint, `local/weird-name` would then collide with a real `weird-name`, and it would fix nothing downstream — the broken parser stays broken on the path case. Sanitising a derived identifier is the right move when the identifier is the only injection site. Here it isn't.

**What landed instead**, on `audit/seam-repo-key-handshake`:

- `internal/sem/repo_key_colon_id_test.go`, `internal/cli/repo_key_colon_test.go` (8baeea15) — unit coverage of all six ID parsers plus `symbols`/`edges`/`neighbors`/`def` end-to-end against a `weird:name` repo containing `od:d/mod.py`. Unix-only; `:` is not a legal Windows path character.
- Contract docs (9da979f7) — `docs/snapshot-format.md` now states that ID fields are unescaped and names the four safe reads (whole-ID compare, last-separator anchor, first-separator cut inside `external:` only, and `file_path` from the record rather than the ID); `docs/semantic-provider-requirements.md` carries the same note at the ID definition. This is the premise `entire-brain` got wrong, stated where a consumer reads it.

Mutation verification: each of `goMethodSymbolByID`, `enclosingTypeShortName`, `externalParts`, `idMatches`, `parseSymbolRef`, `defExternalTypeName` was rewritten into its naive positional form. All six compiled, all six failed the new tests, all six restored to green.

### Left open (belongs in `entire-brain`, not here)

`semanticEndpointPath` should stop parsing the ID. The path is available on symbol records as `file_path`; for relation endpoints it can come from the endpoint's own symbol record via the ID→symbol map already built at ingest. A same-shaped fix would be `Cut(id, ":"+kind+":")` from the right — but taking the field from the record is the one that cannot drift again.

ashtom:

Head moved `1491e49f` → `9da979f7`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

```
01M17S6Q0ZWY  Medium  open    current    internal/cli/root.go:235  `resolveRepo` may return `ENTIRE_REPO_ROOT` verbatim, so a relative local path s…
```

### Verified vs. assumed

suhaanthayyil:
## Finding response — `internal/cli/root.go:235`

**VALID.** Head is now `8bf16e8d`.

### What was wrong

`resolveRepo` returns its input verbatim and falls back to `"."`:

```
internal/cli/root.go:917   if explicit != "" { return explicit, nil }
internal/cli/root.go:920   if env.RepoRoot != "" { return env.RepoRoot, nil }
internal/cli/root.go:940   return ".", nil
```

Every path that actually **builds** a snapshot derives the key from the absolutised path:

```
internal/sem/provider.go:1445   absRepo, err := filepath.Abs(repo)
internal/sem/provider.go:1455   key = repoKey(ctx, absRepo)
internal/sem/search_cache.go:126, 196, 343   repoKey(ctx, absRepo)
internal/sem/records_cache.go:123            repoKey(ctx, absRepo)
```

`doctor` was the only caller passing an unnormalised path. Since the `local/` half of the rule is `"local/" + filepath.Base(repo)`, the two disagreed for any unnormalised spelling. Reproduced on the branch before the fix — this is the literal `doctor --json` output:

```json
{"repo_key":"local/.","repo_root":".", …}
{"repo_key":"local/..","repo_root":"…/001/sub/..", …}
{"repo_key":"local/.","repo_root":"…/001/.", …}
```

`local/.` is a namespace shared by **every** repository on the machine, and it is never the namespace the snapshot itself carries. That inverts the point of publishing the key before a run: a consumer comparing its own prediction against `local/.` rejects a native snapshot, or accepts a foreign one, on a value nothing produces. `TestDoctorRepoKeyMatchesSnapshotHeader` calls itself the anti-drift gate, but it only ever passed an absolute path, so it never exercised this.

### Fix

`internal/sem/provider.go` — `repoKey` normalises the caller's spelling first:

```go
repo = absoluteRepoPath(repo)   // filepath.Abs, with the original kept as fallback
```

Placed at the rule rather than at the one call site. Every in-process caller already passes the absolute path the provider derived, so it is a no-op for them; the exported `RepoKey` has no such guarantee — `doctor` resolves the repository from `ENTIRE_REPO_ROOT`, `--repo` or the working directory, any of which may arrive relative or uncleaned. Normalising here is what keeps the documented promise that the key is reproducible from a path alone.

`report["repo_root"]` is deliberately left reporting what the caller asked for; only the key is normalised.

### Test

`TestDoctorRepoKeyIsIndependentOfHowTheRepoRootIsSpelled` in `internal/cli/doctor_repo_key_test.go`, running `doctor --json` end to end over four spellings of the same repository — absolute, `"."` (with `t.Chdir`), `<repo>/sub/..`, and `<repo>/.` — all of which must report `local/<basename>`.

### Mutation verification

Removed the `repo = absoluteRepoPath(repo)` line, `grep`-confirmed only the helper's definition and doc comment remained with no call site, rebuilt (the unused function still compiles):

```
--- FAIL: …/working_directory
    doctor repo_key for root "." = local/., want "local/001"
--- FAIL: …/uncleaned_absolute
    doctor repo_key for root "…/001/sub/.." = local/.., want "local/001"
--- FAIL: …/trailing_dot
    doctor repo_key for root "…/001/." = local/., want "local/001"
```

Restored, `grep`-confirmed the call back, all four subtests pass. Full `go test ./internal/cli` passes (25s), and `go test ./internal/sem -run 'RepoKey|Cache'` passes — the cache keys fold `repoKey` in, and they were already passing absolute paths, so nothing there moves. `go vet` and `gofmt -l` clean.

PR #170, opened 2026-08-28
fix(sem): name C++ functions from their declarator, not their return type

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/129

## Problem

C++ function and member extraction read the declared name off the first name node in pre-order. That node is the **return type** whenever the return type is a named type rather than a builtin keyword:

| declaration | extracted name (before) |
|---|---|
| `MyType makeThing()` | `MyType` |
| `std::string toText(int a)` | `string` |
| `std::vector<int> *ptrFn(int a)` | `vector` |
| `const Widget &refFn()` | `Widget` |
| `ns::Thing Widget::build()` | `Thing` |
| `MyType get() const` (member) | `Holder.MyType` |
| `T identity(T x)` (template) | `T` |
| `int plain(int a)` | `plain` (correct) |

Only primitive-returning functions were named correctly, which is why the failure never showed up in symbol counts — the symbols were all present, just misnamed and colliding with each other.

The damage went past the name:

- **Symbol identity.** IDs are derived from the qualified name, so same-named functions needed `#sig:` disambiguators to coexist. The `multilang-relations` golden had already baked this in: `Cache.cpp` shipped `function:Entry#sig:0cdc3e16…` and `function:Entry#sig:f911f460…` where the source declares `lookup` and `fetch`.
- **Call resolution.** Resolution matches on name, so a call to `lookup` never found the definition that had been renamed `Entry`. The regenerated golden gains a `lookup -> fetch` CALLS edge that the misnaming had suppressed, plus two `USES_TYPE` edges.
- **Search.** `entire graph search`/`def` for a function name returned nothing for most of a real C++ codebase's API surface.

## Fix

C and Objective-C already take the name from the `declarator` field for exactly this reason (`CURLcode curl_easy_perform(...)`), and were explicitly gated to exclude C++. This extends the same treatment to C++ through a shared declarator walk.

Pointer, reference, array and function declarators all nest their target under the same `declarator` field, so the name is reached by walking that chain. C++ adds two name shapes C does not have, both handled here:

- `field_identifier` — a member definition (`MyType get() const`)
- `qualified_identifier` / `destructor_name` — an out-of-line definition (`Config::name`, `~Config`)

Anything the walk does not model falls back to the previous pre-order identifier search, so no declarator shape loses a name it had. Operator names, constructors and destructors keep their existing extraction.

## Verification

- `internal/sem/parser_cpp_declarator_name_test.go` — three tests covering named/templated/pointer/reference returns, members, and the already-correct qualified/constructor/destructor shapes. Verified to fail on the unpatched parser.
- `multilang-relations.ndjson.golden` regenerated. The only symbol change is the two `Entry#sig:…` records becoming `function:lookup` and `function:fetch`; all relation changes are additive.
- `quality_coverage.json` moves `CALLS` 34→35, `DATA_FLOWS` 22→23, `USES_TYPE` 24→26.
- `mise run check` green.

Found during a per-language extraction-correctness audit of the semantic tier.


ashtom:

**Verdict: clean, with a consumer-churn note**

- Live PR head `f2af8025`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `f2af8025` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking. The fix is right and the golden update proves it.

**1. (note, §6) This changes `compound-v1` symbol IDs.** The golden fixture shows the identity move:

```
- ...:C++:Cache.cpp:function:Entry#sig:f911f460d7735d8f   name:"Entry"
+ ...:C++:Cache.cpp:function:lookup                        name:"lookup"
```

The old IDs were wrong (functions named after their return type), so changing them is a correctness
fix, not a stability violation. But `AGENTS.md` treats `compound-v1` stability as a contract, and
every downstream consumer (Entire Brain) re-keys those symbols when this lands.

**This PR is one of eight open PRs that all change C-family symbol identity or membership** — #170,
#172, #173, #174, #177, #179, #182, #183. Landing them one at a time churns consumers eight times.
Consider merging the C-family set together, or behind one announced identity bump.

**2. (merge order) Conflicts with #174 and #177:**

```
$ git merge-tree --write-tree refs/prs/170 refs/prs/174   ->  CONFLICT
$ git merge-tree --write-tree refs/prs/170 refs/prs/177   ->  CONFLICT
```

### Independent review state

suhaanthayyil:
Thanks. Staleness unchanged — `main` still `bb41f223`, still `13 behind / 1 ahead`. Both notes need refining; the second one materially.

## 1. The conflicts are two generated files, not the parser

#174 and #177 have both moved since the review (`121353cc` and `7ea287a9`). Against current heads:

```
$ git merge --no-ff refs/prs/170 refs/prs/174    |  $ git merge --no-ff refs/prs/170 refs/prs/177
Auto-merging internal/sem/parser.go              |  Auto-merging internal/sem/parser.go
CONFLICT: multilang-relations.ndjson.golden      |  CONFLICT: multilang-relations.ndjson.golden
CONFLICT: quality_coverage.json                  |  CONFLICT: quality_coverage.json
```

`internal/sem/parser.go` — the only production file either pair shares — auto-merges with **zero** conflict markers in both cases. The two conflicted files are both machine-generated, so the resolution is not a merge at all:

```
$ go test ./internal/sem -run TestProviderGoldenSnapshots -update -count=1
$ go test ./internal/sem -run TestProviderGoldenFixtureQualityCoverageReport   # prints the exact JSON to write
$ go test ./internal/sem -run 'TestProviderGoldenSnapshots|TestProviderGoldenFixtureQualityCoverageReport' -count=1
ok
```

Verified for both pairs, then ran the merged trees' own tests — all 12 `TestCPlusPlus*` and all 6 `TestC*Members*`/`TestCFamily*` cases pass on the `170+177` tree.

Also, `#174` is not a separate conflict from `#177`: `git merge-base --is-ancestor refs/prs/174 refs/prs/177` is true — #177 is four commits including #174's two. One conflict, not two.

## 2. "Eight PRs churn consumers eight times" — measured, and it over-counts

I built nine binaries (`main` plus all eight) and ran each over one fixed three-file C/C++/Objective-C corpus exercising all eight constructs, then compared `compound-v1` ID sets against `main` (47 baseline symbols):

| PR | existing IDs invalidated | IDs added | what actually happens |
|---|---|---|---|
| #170 | 6 | 6 | **true re-key** — `function:string#sig:…` → `function:toText` etc. |
| #172 | 3 | 0 | **deletes duplicates only** |
| #173 | 5 | 2 | dedupe: 3 phantoms deleted, 2 survivors lose their `#sig:` suffix |
| #174 | **0** | 12 | **purely additive** |
| #177 | **0** | 15 | **purely additive** |
| #179 | 11 | 2 | dedupe (contains #173) + 4 phantom member-named definitions deleted |
| #182 | **0** | 3 | **purely additive** |
| #183 | 4 | 4 | **true re-key** — `type:NSInteger`/`type:v` → `type:MyInt`/`type:OB` |

Three corrections fall out of that:

- **Three of the eight invalidate nothing at all.** #174, #177 and #182 only add symbols. A consumer keyed on `compound-v1` re-keys zero rows for those.
- **#172's three "removed" IDs are duplicates, not identities.** `main` emits *both* forms for the same method: `method:add` **and** `method:Ledger.add` are both present in the baseline. #172 deletes the unqualified copy; no surviving symbol changes ID.
- **#173 and #179 delete phantoms.** The removed IDs are `struct:Ledger#sig:cde6…`, `…#2`, `…#3` — three "definitions" of one struct manufactured from tag references — collapsing to a single `struct:Ledger`. Two real symbols re-key, and only because the disambiguating suffix is no longer needed.

So only **#170 and #183** re-key correctly-targeted symbols. And the set is not eight independent landings: `#173 ⊂ #179` and `#174 ⊂ #177` (both confirmed with `git merge-base --is-ancestor`), leaving six change-sets.

The recommendation still stands on its merits for #170 and #183 — those two genuinely move identities and are worth announcing together. But "eight PRs, eight churns" is not what the code does, and holding #174/#177/#182 back for an identity bump they do not cause would delay them for nothing.

## Reproduction caveat

Anyone repeating this must pass `--worktree` (or a per-binary `--cache-dir`). My first run used plain `symbols` and returned **identical output from all nine binaries** — the index cache is keyed by tree, not by binary, so every build after the first served `main`'s cached answer and the measurement read as "zero churn everywhere". That result was wrong; the table above is from `--worktree`, which is never cached.

ashtom:

no longer does.

```
01M13TJ7842K  Medium  open    current    internal/sem/parser.go  Standard overloaded operators use an `operator_name` declarator, which this walk…
```

### Verified vs. assumed

suhaanthayyil:
## Finding addressed — `internal/sem/parser.go` (operator declarators)

**VALID.** Pushed `6707f294`.

### Reproduction

`operator_name` does reach the walk, and it contains no identifier node at all — the AST of the source the pipeline actually parses:

```
declarator: function_declarator ["operator delete(void* p, size_t n)"]
  declarator: operator_name ["operator delete"]
    operator ["operator"]      <- anonymous
    delete   ["delete"]        <- anonymous
  parameters: parameter_list ["(void* p, size_t n)"]
```

With no case for it the walk fell through to the historical pre-order identifier search, which steps straight past the identifier-free `operator_name` into the parameter list. Extracted symbols on this branch before the fix:

```
method   Grid.n     line 11   void *operator new(size_t n)
method   Grid.p     line 12   void operator delete(void *p, size_t n)
method   Grid.n     line 13   void *operator new[](size_t n)   <- collides with line 11
function sz         line 22   void *operator new(size_t sz)
(no symbol at all for line 14, `operator const char *() const`)
```

Every one named after its **first parameter**, and two of them collapsed onto the same name.

### A second miss of the same class, found by the A/B against main

`reference_declarator` has no `declarator` **field** in tree-sitter-cpp — it hangs the declarator it wraps off an unnamed child — so the field walk stopped at the `&`. For a member function the fallback then skips the method's own `field_identifier` and returns the first parameter:

```
int &at(int i)                 branch: "i"   main: "at"    <- regression vs main
const Cell &cell(int r) const  branch: "r"   main: "Cell"
```

That first line is a regression this PR introduced against `main`, which the finding's fixture would not have surfaced.

### Fix

- `parser.go:7830` — `case "operator_name":` returns the whole spelling (`operator new`, `operator delete`, `operator new[]`).
- `parser.go:7841` — `case "operator_cast":` names a conversion operator from its text up to the parameter list (`operator const char *`); previously it produced no name and the symbol was dropped entirely.
- `parser.go:7872` — when `ChildByFieldName("declarator")` is absent, fall to `firstDeclaratorChild` (new, `parser.go:7889`), which takes the first named child whose type ends in `declarator`.

### Test

`TestCPlusPlusOperatorNamesComeFromDeclarator` (`parser_cpp_declarator_name_test.go`). Asserts the eight correct names **and** that `Grid.i`, `Grid.r`, `Grid.n`, `Grid.p`, `n`, `sz` never appear as symbol names — a parameter name must not become a callable's name.

### Mutation

Three hunks, three independent mutations, each grep-confirmed applied, each compiling (`go vet` clean under each):

| revert | FAIL |
|---|---|
| `case "operator_name"` renamed | `C++ "Grid.operator delete" not extracted as a method: map[... "Grid.n":"method", "Grid.p":"method", "sz":"function"]` |
| `case "operator_cast"` renamed | `C++ "Grid.operator const char *" not extracted as a method` — symbol absent |
| `next = firstDeclaratorChild(cur)` → `next = nil` | `C++ "Grid.at" not extracted as a method: map[... "Grid.i":"method", "Grid.r":"method" ...]` |

Restored → `grep -c MUTANT` = 0, `TestCPlusPlus*` PASS. Re-run independently after the fact: the first and third mutations applied together still fail on `Grid.at`, and the tree restores clean.

### main vs branch

Same fixture on both. **16 symbols → 17, zero regressions.** Every main symbol has a correctly renamed counterpart and the branch recovers one main dropped:

| source | main | branch |
|---|---|---|
| `const Cell &cell(int r) const` | `Grid.Cell` | `Grid.cell` |
| `Cell *ptr(int i)` | `Grid.Cell` | `Grid.ptr` |
| `void *operator new(size_t n)` | `Grid.n` | `Grid.operator new` |
| `void operator delete(void *p, size_t n)` | `Grid.p` | `Grid.operator delete` |
| `void *operator new[](size_t n)` | `Grid.n` (collides) | `Grid.operator new[]` |
| `operator const char *() const` | **missing** | `Grid.operator const char *` |
| `std::string label() const` | `Grid.string` | `Grid.label` |
| free `void *operator new(size_t sz)` | `sz` | `operator new` |

Relations 34 → 40, none lost: +1 CONTAINS, +1 DEFINES (the recovered conversion operator), +4 USES_TYPE that were previously suppressed because the method name equalled its return type.

### Golden delta

**None, and none was possible.** The only C/C++ golden fixtures are `multilang-relations/Cache.cpp` and `buffer.c`; neither contains a reference return, an operator overload or a conversion operator. `TestProviderGoldenSnapshots` (30 subtests), `TestProviderGoldenFixtureQualityCoverageReport` and `TestCapabilityMatrixCoversEmittedRelations` pass unmodified.

### A larger, pre-existing defect this uncovered — deliberately not fixed here

`operator==`, `operator+`, `operator[]`, `operator()` and `operator<<` **never reach this walk at all**. `maskCPlusPlusOperatorCall` (`parser.go:2889`, pattern at `parser.go:2586`) rewrites `operator<op>(` to `op(` before tree-sitter runs, and the conversion pattern rewrites `operator bool()` to `convert()`. The masked source:

```
  bool op(        const Foo& o) const { ... }   // was: bool operator==(const Foo& o) const
  convert()       const { ... }                 // was: operator bool() const
```

Node offsets then map back onto the original bytes, so those come out as `Foo.op` and `Foo.operato`. Every `operator==`/`+`/`<<` in a repository collapses onto the single name `op` — the same bug class this PR fixes, one layer earlier. `main` does this identically, so it is not a regression from #170, and it lives in the pre-parse mask rather than the declarator walk. Undoing that mask needs the nlohmann/fmt corpus to prove parse rate does not regress, so it is left alone and recorded in a comment on the new test. Worth its own issue.

### Also checked

LF and CRLF produce identical symbol sets with `parseError=false` both. 60 test functions matching `CPlusPlus|ObjectiveC|Declarator|Golden|CFamily|Cxx` pass. `gofmt`/`go vet` clean. `fastCFamilyEntities` — the inventory fallback path — has its own naming logic and was not audited for these shapes.

suhaanthayyil:

- TestCPlusPlus / TestCFamily, plus `TestProviderGoldenSnapshots` — pass on the merged tree

PR #171, opened 2026-08-28
test(sem): assert every semantic language extracts symbols and relations

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/130

## Why

A grammar that loads, parses without a syntax error, and yields nothing is invisible. Aggregate symbol counts across a repository never show it, and `capabilities --json` keeps advertising the language regardless. Nothing in the suite asserted that any given language extracts anything at all — 10 of the 36 semantic languages had golden coverage; the other 26 were unverified.

This matters more now that entire-graph is proposed as the org's single symbol producer (#146): a language that silently yields nothing would then yield nothing for every consumer.

## What

One minimal fixture per semantic language — a function, a type, a member, and a call between them — under `internal/sem/testdata/langmatrix/`, plus four assertions:

| test | guards against |
|---|---|
| `TestLanguageExtractionMatrix` | a language extracting nothing, or the wrong things, from unmistakable definitions |
| `TestLanguageMatrixCoversEverySemanticLanguage` | a language registered and counted in `capabilities --json` with no evidence it extracts anything |
| `TestDocumentedLanguageListsMatchTheRegistry` | either tier list in `docs/language-support.md` drifting away from the registry |
| `TestLanguageMatrixCallsGapsAreExactlyAsAudited` | a language advertising `CALLS` while extracting none — and a fix landing without its audited gap being retired |

## Audited CALLS gaps

Four languages advertise `CALLS` in the capability report and extract none from their fixture's idiomatic call site. They are pinned by name with the reason rather than papered over, and the test fails if the list grows *or* goes stale:

- **C++** — a default-constructed stack receiver (`Ledger ledger;`) carries no inferred type, so its method calls resolve to nothing. `Ledger l = Ledger();` and `new Ledger()` both do resolve, so this is a hole in declaration parsing rather than in C++ receiver resolution.
- **F#** — application by juxtaposition (`add ledger amount`) has no scanner: the dotted scanners need a `.` and the generic one needs `name(`. (#163 covers the forward-pipe form.)
- **Julia** — a definition inside `module ... end` is emitted as a method, and bare calls are barred from resolving to methods. (#161 addresses this.)
- **Objective-C** — a local's declared type is not inferred, so `[ledger add:amount]` resolves to nothing; only `[self …]` binds.

## Reconciliation result

No docs/registry disagreement was found in either direction. Both the semantic list (36) and the inventory-only list (149) match the registry exactly, and the counts stated in the doc (185 / 36 / 149) match `capabilities --json`.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .`, and the full `go test ./...` (3833 tests) are green. This branch and #172/#173/#174 merge together cleanly; the combined tree passes 3841 tests.

### DIFF
--- a/docs/language-support.md
+++ b/docs/language-support.md
@@ -58,6 +58,20 @@ Current generated counts:
 - Zig
 - Zsh
 
+Every language in this list has a minimal fixture under
+`internal/sem/testdata/langmatrix/` and is asserted by
+`TestLanguageExtractionMatrix` to extract real symbols and relations. A grammar
+that loads and parses without error but yields nothing is invisible in
+aggregate symbol counts, so the per-language assertion is the only thing that
+catches it. `TestLanguageMatrixCoversEverySemanticLanguage` keeps that fixture
+set in lockstep with the registry, and `TestDocumentedLanguageListsMatchTheRegistry`
+keeps both lists on this page in lockstep with the registry.
+
+Languages that advertise `CALLS` but extract none from their fixture's
+idiomatic call site are pinned in `languageMatrix` with the reason. That list is
+enforced in both directions: a language going dark fails, and so does a fix
+landing without the audited gap being retired.
+
 Protocol Buffers support covers proto3 and legacy proto2 declarations,
 including files that omit the syntax declaration, proto2 field labels, and
 groups. Compatibility parsing preserves original source locations, signatures,
--- a/internal/sem/language_matrix_test.go
+++ b/internal/sem/language_matrix_test.go
@@ -24,8 +24,28 @@ type languageFixture struct {
 	// callsGap documents a language whose capability report advertises CALLS
 	// while this fixture's idiomatic call form produces none. It is an audited
 	// defect list, not a licence: closing one of these fails this test until the
-	// entry is removed, and adding a new one requires justifying it here.
+	// entry retires (by removal, or by callsGapHolds below reporting the gap's
+	// cause gone), and adding a new one requires justifying it here.
 	callsGap string
+	// callsGapHolds, when non-nil, is the condition callsGap's stated cause
+	// reduces to, evaluated against the extractor on every run: the gap is
+	// claimed only while it reports true. It makes an audited gap
+	// self-verifying rather than a hand-maintained literal, and it tightens the
+	// assertion in both directions. The entry then claims a cause rather than
+	// restating its effect, so naming the wrong cause fails here; and the commit
+	// that closes the gap retires the entry by flipping the condition, which is
+	// what keeps this test order-independent. A literal gap silently disagrees
+	// with an extractor fix that lands on a different branch — the two merge
+	// cleanly and only the merged tree fails — whereas a condition read from the
+	// extractor at run time cannot go stale that way.
+	//
+	// It receives the fixture's own extraction result, because the two shapes a
+	// cause takes need different evidence: a resolution rule is a predicate over
+	// the provider, while a missing symbol is only visible in what the fixture
+	// produced. Leave it nil for a gap that reduces to neither — an absent
+	// scanner or an uninferred type has no witness beyond the missing CALLS
+	// itself — and that gap is asserted unconditionally, exactly as before.
+	callsGapHolds func(got languageMatrixResult) bool
 }
 
 // languageMatrix covers every language in Capabilities().SemanticLanguages.
@@ -77,6 +97,11 @@ var languageMatrix = []languageFixture{
 		// method, so a module-scoped package produces no same-file CALLS.
 		relations: []string{"DEFINES", "CONTAINS", "CONSTRUCTS", "PARAM_TYPE", "USES_TYPE"},
 		callsGap:  "a Julia definition inside `module ... end` is emitted as a method, and bare calls are barred from resolving to methods",
+		// The whole gap reduces to one rule: bare-call resolution skips method
+		// targets for Julia. Reading it here rather than restating its
+		// consequence means the fix that admits Julia retires this entry on its
+		// own, and that a wrong diagnosis is caught instead of recorded.
+		callsGapHolds: func(languageMatrixResult) bool { return !nameCallMayTargetMethod("Julia") },
 	},
 	{dir: "kotlin", symbols: []string{"class:Ledger", "field:total", "method:add", "function:ledgerDouble"}, relations: []string{"DEFINES", "CONTAINS", "CALLS", "CONSTRUCTS"}},
 	{dir: "lua", symbols: []string{"function:new", "function:add", "function:ledger_double"}, relations: []string{"DEFINES", "CALLS"}},
@@ -208,6 +233,13 @@ func TestLanguageExtractionMatrix(t *testing.T) {
 // call site. Both directions are failures worth knowing about: a new entry
 // means a language went dark, and a stale entry means a fix landed without the
 // audited gap list being retired with it.
+//
+// The audit is read from the extractor wherever a gap reduces to a checkable
+// condition (see callsGapHolds), so a gap closes itself in the same commit that
+// closes it in the extractor. That is what makes the assertion safe to state
+// exactly: a gap pinned only as a literal disagrees with a fix that lands on a
+// different branch, and because the two touch different files they merge
+// cleanly and fail only afterwards, on a tree neither author tested.
 func TestLanguageMatrixCallsGapsAreExactlyAsAudited(t *testing.T) {
 	t.Parallel()
 	claimsCalls := map[string]bool{}
@@ -229,10 +261,20 @@ func TestLanguageMatrixCallsGapsAreExactlyAsAudited(t *testing.T) {
 				return
 			}
 			got := runLanguageFixture(t, fixture.dir)
+			auditedGap := fixture.callsGap
+			if auditedGap != "" && fixture.callsGapHolds != nil && !fixture.callsGapHolds(got) {
+				// The gap's stated cause no longer holds, so the entry has
+				// retired itself and the language is now held to the same
+				// standard as every other language that advertises CALLS.
+				auditedGap = ""
+			}
 			switch {
-			case got.relations["CALLS"] && fixture.callsGap != "":
-				t.Fatalf("%s now extracts CALLS; remove its audited gap (%s) from languageMatrix", language, fixture.callsGap)
-			case !got.relations["CALLS"] && fixture.callsGap == "":
+			case got.relations["CALLS"] && auditedGap != "":
+				t.Fatalf("%s now extracts CALLS; remove its audited gap (%s) from languageMatrix", language, auditedGap)
+			case !got.relations["CALLS"] && auditedGap == "":
+				if fixture.callsGap != "" {
+					t.Fatalf("%s still extracts no CALLS, but its audited gap (%s) reports itself closed; callsGapHolds names the wrong cause", language, fixture.callsGap)
+				}
 				t.Fatalf("%s advertises CALLS but extracted none from its fixture; relations seen: %v", language, matrixSortedKeys(got.relations))
 			}
 		})
--- a/internal/sem/language_matrix_test.go
+++ b/internal/sem/language_matrix_test.go
@@ -1,0 +1,330 @@
+package sem
+
+import (
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+)
+
+// languageFixture pins what one semantic language must extract from a minimal
+// fixture that contains a handful of unmistakable definitions: a function, a
+// type, and a member. It exists because a zero-symbol language is invisible in
+// aggregate counts — a grammar can keep loading, keep parsing without error,
+// and keep being advertised in `capabilities --json` while producing nothing at
+// all. Only a per-language assertion catches that.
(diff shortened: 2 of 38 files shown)


ashtom:

**Verdict: clean**

- Live PR head `2c599378`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `2c599378` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|
| `go test ./internal/sem/` | merged onto main | pass |
| PR's own tests | merged onto main | pass |
| revert-the-fix check | — | N/A — test-only PR (see below) |

### Findings

Nothing blocking.

**Correction to my own automated check, for honesty:** my harness reported "tests discriminate" for
this PR, but that result is meaningless here — the only non-test `.go` file in the diff is
`internal/sem/testdata/langmatrix/go/ledger.go`, a *fixture*. Reverting it breaks the matrix test
because the fixture is gone, not because a production change was undone. **This is a test-only PR**
and the revert check does not apply to it.

What I can say from execution: the matrix tests pass on the merged tree, and they do assert per
language that symbols *and* relations are non-empty, which is the property that matters.

**Note (§6):** #176 and #178 are both stacked on this PR's commit `2c599378`, so this needs to land
first (or they need rebasing).

### Independent review state

- Entire trail **#130** exists but is still `phase=building` — the agent review has **not completed**.

suhaanthayyil:
Thanks — the self-correction is right, and understated. Staleness unchanged: `main` still `bb41f223`, still `13 behind / 1 ahead`.

**Test-only: confirmed, and it is broader than one file.** The diff is 38 files: one `_test.go`, 36 fixtures under `internal/sem/testdata/langmatrix/`, and `docs/language-support.md`.

```
$ git diff --name-only <base> refs/prs/171 | grep -v '_test\.go$' | wc -l
37
$ ... | grep -vc 'testdata/'
1        # docs/language-support.md
```

Zero files outside `internal/sem/testdata/` and `docs/`. `ledger.go` was only singled out because the harness filtered on `.go`; the same argument covers 35 sibling fixtures in other languages. There is no production code here to revert.

## What the matrix actually guarantees

Since the revert check does not apply, I substituted mutation testing — three mutations to production code (each confirmed to compile and to have actually changed the file), run against the merged tree:

**Mutation 1 — remove `"Ruby"` from `supportsCallExtraction` (the CALLS *declaration*):** matrix stays **GREEN**.
**Mutation 2 — `fileNeedsCallScan := needsCallScan && file.Language != "Ruby"` (the bare-name call scan):** matrix stays **GREEN**. Ruby's CALLS come from the receiver pass, not this one.
**Mutation 3 — `if spec.callResolution == "full" && file.Language != "Ruby"` (the receiver pass):** matrix goes **RED**, precisely and only on ruby, in both tests:

```
--- FAIL: TestLanguageExtractionMatrix/ruby
    language_matrix_test.go:199: fixture ruby is missing relation "CALLS";
                                 got [CONTAINS DATA_FLOWS DEFINES]
--- FAIL: TestLanguageMatrixCallsGapsAreExactlyAsAudited/ruby
```

Restoring the file returns it to green, 36 language subtests.

So the guarantee is: **per-language emission**, not declaration. For each of 36 languages it pins that the fixture parses as the expected language label, that extraction is non-empty at all, that specific named symbols appear, and that specific named relation types appear — plus the audited CALLS-gap list in both directions. That is exactly the failure the header comment claims (a grammar that loads, parses cleanly, and yields nothing), and mutation 3 shows it is not a paper assertion.

Mutation 1 is worth recording as a boundary rather than a defect: when a language stops *advertising* CALLS, `TestLanguageMatrixCallsGapsAreExactlyAsAudited` takes the `!claimsCalls[language]` early return and stays silent. A lost declaration is invisible here. That direction is covered by `TestCapabilityMatrixCoversEmittedRelations` and by #168, so nothing is unguarded — but this suite should not be read as covering the capability table.

## Merge-order note is right, with one addition

#176 and #178 are stacked on this commit (`git merge-base --is-ancestor refs/prs/171 refs/prs/{176,178}` true for both) and each already retires or rewrites its own `callsGap` entry — #178 drops the F# entry outright, #176 replaces the C++ one. Merged against this PR, both are green.

**#161 is the one that is not.** It merges with zero textual conflict and then fails:

```
$ git merge --no-ff refs/prs/171 refs/prs/161      # clean, no conflicted files
$ go test ./internal/sem -run TestLanguageMatrixCallsGapsAreExactlyAsAudited -count=1
--- FAIL: .../julia
    language_matrix_test.go:234: Julia now extracts CALLS; remove its audited gap
      (a Julia definition inside `module ... end` is emitted as a method, and bare
      calls are barred from resolving to methods) from languageMatrix
```

#161 is the fix for the exact gap julia's entry audits, and it never touches `language_matrix_test.go`. I merged every other open PR against this one — #162, #165, #168, #169, #170, #172, #176, #178, #182, #183 all green — so #161 is the only pair that breaks, and `git merge-tree` cannot see it.

Whichever lands second must delete julia's `callsGap` and add `"CALLS"` to its relations in the same commit; verified green. It cannot be pre-applied here — on this branch alone that edit fails the opposite way (`Julia advertises CALLS but extracted none from its fixture`). Purely a landing-order item, and the test behaving this way is the suite working as designed.

suhaanthayyil:
## A merge landmine between this branch and #161, and a second one behind it

`TestLanguageMatrixCallsGapsAreExactlyAsAudited` was order-dependent. This branch and #161 touch
disjoint files, so they merge with no textual conflict — and then the merged tree fails:

```
$ git checkout -q 2c599378 && git merge --no-edit e0a2b01b
Merge made by the 'ort' strategy.
--- FAIL: TestLanguageMatrixCallsGapsAreExactlyAsAudited/julia
    language_matrix_test.go:234: Julia now extracts CALLS; remove its audited gap (a Julia
    definition inside `module ... end` is emitted as a method, and bare calls are barred from
    resolving to methods) from languageMatrix
```

Each branch is correct on its own; only the merge is wrong, and neither author's CI sees it.
`git merge-tree` cannot report it because the disagreement is between a literal in this file and
behaviour in `provider.go`.

## Fix: read the audit from the extractor instead of pinning it as a literal

`languageFixture` gains an optional `callsGapHolds`, the condition the gap's stated cause reduces
to, evaluated on every run. The gap is claimed only while it reports true, so the commit that
closes a gap retires its audit entry by flipping the condition — in whichever order the fix and
this matrix land.

The Julia gap reduces to exactly one rule, so it now states that rule rather than restating its
consequence:

```go
callsGapHolds: func(languageMatrixResult) bool { return !nameCallMayTargetMethod("Julia") },
```

This tightens the assertion rather than relaxing it. The entry now claims a *cause*, and a third
outcome is now a failure: a gap whose named cause has flipped while the language still extracts no
CALLS is a wrong diagnosis, and says so. Gaps that reduce to neither a resolution rule nor a
missing symbol (Objective-C's uninferred local type, F#'s absent scanner) leave it nil and stay
asserted unconditionally, exactly as before. Nothing was skipped, loosened or deleted.

`callsGapHolds` receives the fixture's own extraction result, because the two shapes a cause takes
need different evidence: a resolution rule is a predicate over the provider, while a missing symbol
is only visible in what the fixture produced. The second landmine below needs that second form.

## Verification

| tree | before | after |
|---|---|---|
| this branch alone | pass | pass |
| this branch + #161 | **FAIL** (julia) | pass |

Mutation-tested, each mutation `go vet`-clean and confirmed applied by grep before running:

| mutation | result |
|---|---|
| drop `callsGapHolds` (literal gap, #161 landed) | FAIL — "Julia now extracts CALLS; remove its audited gap" |
| `callsGapHolds` returns a wrong cause (`false`), #161 reverted | FAIL — "reports itself closed; callsGapHolds names the wrong cause" |
| delete the whole julia gap entry, #161 reverted | FAIL — "Julia advertises CALLS but extracted none" |
| revert #161's `nameCallMayTargetMethod` clause, entry intact | pass — the gap re-claims itself |

The last row is the point: the assertion now tracks the extractor in both directions.

## A second landmine of the same class: #176 + #177

Not fixed here, because the entry involved lives on #176. It reproduces against the *original*
`2c599378` as well, so it is independent of this change:

```
$ git checkout -q 2c599378 && git merge --no-edit <#176> && git merge --no-edit <#177>
    languageMatrix
```

#176 and #177 each pass with this branch individually — #176 types the receiver, #177 qualifies the
method symbol, and the C++ gap needs *both* closed. Because #176 is stacked on this branch and #177
is not, the two merge cleanly and only the combined tree fails.

#176's re-audited cause is fixture-observable, which is what the second form of `callsGapHolds`
is for. Verified in both directions:

```go
```

| tree | with that one line |
|---|---|
| this branch + #176 (gap still open) | pass |
| this branch + #176 + #177 (gap closed) | pass |

Causal check behind it: without #177 the cpp fixture emits `function:Add`; with #177 it emits
`method:Add`, which is exactly the qualification #176's comment says is missing.

## Swept for further instances

Every open PR that changes extraction (#162, #163, #170, #172, #173, #174, #179, #182, #183) was
checked against all four matrix assertions. None of them flips one: in each case the changed code
path's trigger — a forward-pipe operator, a named return type, an anonymous specifier, a `typedef`
node — is absent from the specific fixture. #178 is stacked on this branch and retires the F# gap
itself, so it is already order-independent. A tree with #161, #162, #163, #170, #172, #173, #176,
#179, #182 and #183 merged together on top of this branch builds and passes the full matrix.

Unrelated, but noticed while sweeping: #179 reimplements the same `cFamilyTagReference` helper #173
introduces, so those two will likely conflict with each other.

suhaanthayyil:
Follow-up: the mechanism this branch introduced has now been applied to the second literal gap, and I swept for the rest of the class.

**cpp is now derived too** — landed on #176 (`dfa72a7e`), since that branch is stacked on this one and #177 (which actually closes the gap) is not. Reproduced first: #171 + #174 + #176 + #177 failed `TestLanguageMatrixCallsGapsAreExactlyAsAudited/cpp`, with every PR green alone and every merge clean.

Its cause reduces to a missing symbol rather than a resolution rule, so it uses the `got languageMatrixResult` form your comment anticipated:

```go
```

**Sweep result.** After the Julia fix, two literal gaps remained. Merging this branch with every open PR that touches the extractor or the matrix:

| gap | status |
|---|---|
| julia | derived (this branch) |
| cpp | derived (#176) — landmine found and fixed |
| fsharp | **no landmine** — #178 already removes its own entry, which is the correct pattern |
| objc | **no landmine** — #172 and #183 merge green; neither does local type inference, so the gap correctly still holds |

Also merged clean with the matrix tests passing: #161, #162, #165, #168, #181.

I did not run all 231 pairs, and want to be explicit that this is a reasoned scope rather than exhaustive coverage: the matrix test can only break when a gap *closes*, and only an extractor change closes one, so the extractor-touching set is the complete set for this class.

ashtom:

**Verdict: clean and mergeable**

Head moved `2c599378` → `51d07eb7`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### Note

Still a test-only PR; the revert check does not apply (my harness's "discriminates" line for this
one reverts a *fixture*, not a fix). No open, current findings.

**#176 and #178 are stacked on this**, so it should land first.

### Entire trail findings

Re-queried immediately before posting (`--status open --freshness current`): **no open, current findings**.

### Verified vs. assumed

suhaanthayyil:
Agreed on the verdict, and confirming the state rather than just accepting it: 38 matrix subtests pass on the current head (`51d07eb7`).

Recording where the audited-gap mechanism this PR carries now stands, since it is the thing most likely to go stale between branches:

| gap | form | where |
|---|---|---|
| julia | **derived** — `callsGapHolds` reads `nameCallMayTargetMethod("Julia")` | this branch |
| cpp | **derived** — `callsGapHolds` reads `got.symbols["method:Add"]` | #176 (`dfa72a7e`), which is stacked on this branch and is the only one of the pair that can carry the condition |
| fsharp | literal — **no landmine**: #178 removes its own entry, which is the correct pattern | #178 |
| objc | literal — **no landmine**: #172 and #183 merge green and neither does local type inference, so the gap correctly still holds | — |

Both derivations were found the same way: merging this branch with the PR that closes the gap and running the matrix, because a clean `git merge-tree` does not see this class at all. Julia (#161) and cpp (#176+#177) each merged cleanly and then failed. I swept this branch against every open PR that touches the extractor or the matrix — the only PRs that can close a gap — and found no third instance.

One thing I want to be explicit about rather than imply: I did **not** run all pairs. The matrix can only break when a gap *closes*, and only an extractor change closes one, so the extractor-touching set is the complete set for this class — a reasoned scope, not exhaustive coverage.

Also confirming your correction about the revert check on this PR: you are right that it is meaningless here. The only non-test file in the diff is `internal/sem/testdata/langmatrix/go/ledger.go`, a fixture — reverting it breaks the matrix because the fixture is gone, not because a production change was undone. I checked separately what the matrix does guarantee, with three compiling mutations: it asserts per language that symbols **and** relations are non-empty, so a language whose extraction goes dark fails here rather than going unnoticed.

PR #172, opened 2026-08-28
fix(sem): qualify recovered Objective-C methods by their container

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/131

## The defect

`objectiveCMethodEntities` is a recovery scanner: it supplements the tree-sitter walk with implementation methods the walk misses. It named them by their bare selector (`add`) while the walk names them `Class.selector` (`Ledger.add`), and `appendMissingEntities` dedupes on `kind + name` — so the two never collided.

The result: **every Objective-C method the walk already found gained a second, unqualified record** at the same file, the same lines, the same signature, and the same container.

```
method   add   qn=Ledger.add   L6-6  sig="- (int)add:(int)amount"  container=…:class:Ledger
method   add   qn=add          L6-6  sig="- (int)add:(int)amount"  container=…:class:Ledger
```

This is not cosmetic:

- the symbol table reports each method twice;
- each call into a method produces two `CALLS` edges (`CALLS(twice->add)` twice on a three-line file);
- the duplicate pair matches itself as a `SIMILAR_TO` near-clone.

The repository's own `TestObjectiveCMethodFallbackExtractsColonSelectors` was asserting on the *duplicate* — it looked for the unqualified name, which only ever existed because of this bug.

## The fix

Track the enclosing `@interface` / `@implementation` / `@protocol` while scanning and qualify the recovered name with it, matching the walk. A category (`@implementation Foo (Bar)`) attributes to `Foo`, exactly as the walk names the same `class_implementation` node, and a method written outside any container is still recovered unqualified rather than dropped.

The existing fallback test now asserts the qualified name and counts the records, so the duplication cannot come back.

Found by the per-language extraction audit; the fixture matrix in #171 covers Objective-C.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .`, full `go test ./...` (3759 tests) green.

### DIFF
--- a/internal/sem/parser.go
+++ b/internal/sem/parser.go
@@ -7099,8 +7099,24 @@ func objectiveCMethodEntities(content string) []Entity {
 	}
 
 	var entities []Entity
+	scope := ""
 	for i := 0; i < len(lines); i++ {
 		trimmed := strings.TrimSpace(lines[i])
+		// Track the enclosing @interface/@implementation/@protocol so a
+		// recovered method carries the same qualified name the tree-sitter walk
+		// gives it. Without the scope every method the walk DID find gained an
+		// unqualified twin here: appendMissingEntities keys on kind+name, so
+		// `Ledger.add` and `add` were two different symbols at one source
+		// location, duplicating the method in the symbol table, duplicating
+		// every call edge into it, and pairing it with itself as SIMILAR_TO.
+		if declared, ok := objectiveCContainerName(trimmed); ok {
+			scope = declared
+			continue
+		}
+		if trimmed == "@end" {
+			scope = ""
+			continue
+		}
 		if !strings.HasPrefix(trimmed, "- (") && !strings.HasPrefix(trimmed, "+ (") {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


ashtom:

**Verdict: clean**

- Live PR head `aa7eab68`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `aa7eab68` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

### Independent review state

suhaanthayyil:
Thanks — nothing to change. Confirmed on my side that the qualification is on the container and that reverting it fails the new tests.

One note for the sequencing conversation rather than for this diff: this is one of the eight C-family PRs that move symbol identity or membership. Nothing here conflicts with the others, but see the point on #174 — I have now measured that #174's inline-aggregate change was a **regression against `main`** (a named inner `struct` lost its symbol), which is the kind of thing that only shows up when these are checked against `main` rather than against each other. This PR does not have that problem; I checked.

PR #173, opened 2026-08-28
fix(sem): stop C-family tag references becoming struct definitions

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/132

## The defect

In C the tag name is part of the type syntax. `struct Ledger *l` and `enum Mode m` parse to the same `struct_specifier` / `enum_specifier` node type as `struct Ledger { … }` does — the definition is the one carrying a body. Both were extracted as definitions.

So every *mention* of a type produced another "definition" of it:

```
struct   Ledger   L1-3   <- the definition
struct   Ledger   L5-5   <- `struct Ledger *l` in a parameter list
struct   Ledger   L10-10 <- `struct Ledger l = {0}` in a body
```

The shape that makes it worst is a public C header: it is mostly prototypes, so nearly every struct mention in it is a parameter type. A header containing one forward declaration and two prototypes emitted three struct "definitions" and zero function symbols. Those phantoms then compete with the real definition in search results and in `def`.

## The fix

Skip a body-less struct/enum specifier for C, C++ and Objective-C. A file-scope `struct Ledger;` is a forward declaration — an incomplete type, not a definition either — and is skipped on the same rule.

## One test changed

`TestCMaskInterDeclarationAnnotations` used the presence of a `struct FOO` entity as a proxy for "the C attribute-macro mask did not blank the tag `FOO` in `struct FOO bar = { 0 };`". That entity only ever existed because of this bug, so the guard now asserts the mask output directly — which is what it was actually about.

Found by the per-language extraction audit; the fixture matrix in #171 covers C and C++.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .`, full `go test ./...` (3759 tests) green.

### DIFF
--- a/internal/sem/parser.go
+++ b/internal/sem/parser.go
@@ -4933,12 +4933,18 @@ func entityFromNode(node *sitter.Node, src []byte, language, scope string) (Enti
 		if language == "Objective-C" && node.Type() == "struct_declaration" {
 			return Entity{}, false
 		}
+		if cFamilyTagReference(node, language) {
+			return Entity{}, false
+		}
 		kind = "struct"
 		name = nodeName(node, src)
 	case "enum_item", "enum_declaration", "enum_specifier":
 		// Same as struct_declaration: Zig enums are anonymous literals named by
 		// the enclosing variable_declaration.
 		if language == "Zig" {
+			return Entity{}, false
+		}
+		if cFamilyTagReference(node, language) {
 			return Entity{}, false
 		}
 		kind = "enum"
@@ -7837,6 +7843,28 @@ func goReceiverName(node *sitter.Node, src []byte) string {
 		name = name[index+1:]
 	}
 	return strings.TrimSpace(name)
+}
+
+// cFamilyTagReference reports whether a C-family struct/enum specifier NAMES a
+// tag rather than defining one. In C the tag name is part of the type syntax, so
+// `struct Ledger *l` and `enum Mode m` parse to the same struct_specifier /
+// enum_specifier node type as the definition does — the definition is the one
+// carrying a body. Without this check every mention of a type produced another
+// "definition" of it: a three-line C file that declares `struct Ledger` once and
+// uses it twice emitted three struct symbols at three different lines, and a
+// header whose only content is prototypes emitted a phantom definition for each
+// parameter type.
+//
+// A body-less specifier at file scope is a forward declaration
+// (`struct Ledger;`) — an incomplete type, not a definition either — so it is
+// skipped on the same rule.
+func cFamilyTagReference(node *sitter.Node, language string) bool {
+	switch language {
+	case "C", "C++", "Objective-C":
+	default:
+		return false
+	}
+	return !validNode(node.ChildByFieldName("body"))
 }
 
 func qualify(scope, name string) string {
--- a/internal/sem/parser_c_macro_regression_test.go
+++ b/internal/sem/parser_c_macro_regression_test.go
@@ -1,6 +1,7 @@
 package sem
 
 import (
+	"strings"
 	"testing"
 )
 
@@ -256,8 +257,12 @@ int tail_fn(void)
 	if countEntity(entities, "struct", "Curl_addrinfo") == 0 {
 		t.Errorf("struct Curl_addrinfo missing: %+v", entities)
 	}
-	if countEntity(entities, "struct", "FOO") == 0 {
-		t.Errorf("struct tag FOO must not be blanked: %+v", entities)
+	// `struct FOO bar` names a tag, it does not define one, so it is correctly
+	// not a symbol. Assert the mask guard directly on the masked source instead:
+	// the declarator heuristics must not mistake the tag for an attribute macro
+	// and blank it.
+	if masked := maskCUnsupportedSyntax(src); !strings.Contains(masked, "struct FOO bar") {
+		t.Errorf("struct tag FOO must not be blanked:\n%s", masked)
 	}
 	if countEntity(entities, "function", "tail_fn") != 1 {
 		t.Errorf("tail_fn missing: %+v", entities)
--- a/internal/sem/parser_c_tag_reference_test.go
+++ b/internal/sem/parser_c_tag_reference_test.go
@@ -1,0 +1,89 @@
+package sem
+
+import "testing"
+
+// TestCTagReferencesAreNotDefinitions pins the phantom definitions that a
+// C-family struct or enum tag used to produce. In C the tag name is part of the
+// type syntax, so `struct Ledger *l` parses to the same struct_specifier node
+// type as `struct Ledger { ... }` — the definition is the one carrying a body.
+// Extracting both meant every mention of a type produced another "definition"
+// of it, at a different line, competing with the real one in search results and
+// in `def`.
+func TestCTagReferencesAreNotDefinitions(t *testing.T) {
+	t.Parallel()
+	entities, language := TreeSitterParser{}.Parse("ledger.c", `struct Ledger {
+    int total;
+};
+
+enum Mode {
+    MODE_FAST,
+};
+
+int ledger_add(struct Ledger *ledger, int amount) {
+    return ledger->total + amount;
+}
+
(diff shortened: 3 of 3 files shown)


ashtom:

**Verdict: clean, with one recall trade worth naming**

- Live PR head `6f601f5f`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `6f601f5f` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking.

**1. (deliberate trade, flagging it) Opaque forward-declared types now produce no symbol at all.**
`cFamilyTagReference` skips any body-less `struct_specifier`/`enum_specifier`, and the doc comment
says forward declarations are skipped "on the same rule". That is defensible — a forward declaration
is not a definition — but the consequence is that a type declared only as `struct Ledger;` in a
header and defined in a `.c` the graph does not index becomes invisible rather than incomplete. If
that matters for opaque-handle-style C APIs, it may deserve a distinct incomplete-type record rather
than a skip. Not a blocker, and the current behaviour is strictly better than emitting a phantom
definition per mention.

**Note (§6):** #179 is stacked on this PR's commit `6f601f5f`, so this lands first.

### Independent review state

suhaanthayyil:
Thanks — no change needed here, and I want to confirm the trade you flagged rather than let a "clean" verdict imply nobody checked it.

**Verified on this branch.** A header carrying only an opaque declaration:

```c
struct Ledger;
struct Ledger *ledger_new(void);
```

contributes **no symbol at all** — indexing it alongside a `.c` that uses it yields only the using function. So the consequence you describe is exactly right: an opaque handle type declared in a header and defined in a translation unit the graph does not index becomes invisible, not incomplete.

**I agree with the trade as it stands, and would not change it in this PR.** The alternative this PR replaced was worse in a way that compounds: a phantom `struct` definition emitted per *mention*, so the same type got several competing definition records and any consumer asking "where is `Ledger` defined" got a wrong answer with confidence. Invisible is recoverable by indexing more; confidently-wrong is not recoverable at all, because nothing downstream can tell it apart from the truth.

Where it will bite, concretely, so the decision is on the record: opaque-handle C APIs (`sqlite3`, `FILE`-style handles, most C library boundaries) are *designed* so the definition lives somewhere a consumer never sees. For those the type is not missing by accident — it is missing by design, permanently, and no amount of indexing will fix it.

So a distinct incomplete-type record is the right eventual answer, and it is a feature rather than a correction to this change: it needs a kind that consumers understand as "declared, definition not in this graph", which is a schema question (ADR 0001, additive minor) and not a parser one. Filing it that way rather than widening this PR.

Agreed on the sequencing note: #179 is stacked on `6f601f5f`, so this lands first.

suhaanthayyil:
Following up on the opaque-type item I said I would file rather than fold in. **Decision: it stays open, and here is the concrete reason rather than a deferral.**

I checked what landing it would actually cost, because "follow-up" should not be a way of not deciding.

An incomplete-type record needs a symbol kind (or flag) that means *declared, definition not in this graph*. That is a **wire-format change**, and under ADR 0001 an added optional kind is additive — so a `SchemaVersion` MINOR bump. That bump is no longer free, and deliberately so, because of two things that landed since:

1. **#154** makes cache-entry validity an EXACT `SchemaVersion` match. A minor bump therefore invalidates every existing user's provider-records and search-snapshot caches and forces one full cold rebuild.
2. **#156** now pins a digest of the frozen `Result` wire shape beside the exact version string, so the shape and the version have to move together in one reviewed edit.

Neither is a reason not to do it. They are reasons it has to be **one deliberate schema change**, not a rider on a parser fix.

The sequencing argument is the stronger one. There are currently **eight open C-family PRs that change symbol identity or membership** (#170, #172, #173, #174, #177, #179, #182, #183 — see the note on #170). Adding a new symbol kind on top of that means consumers re-key *and* learn a new kind in the same window, with the identity churn making it hard to tell which change caused what. The right order is: land the C-family set, let identity settle, then make the schema change on a quiet base where its blast radius is attributable.

So: **not in this PR, not in this batch, and not because it is unimportant** — because it is a schema decision that wants a stable base and a version bump nobody else is competing for. Filing it that way. If you would rather it go in now and take the cache rebuild alongside the C-family churn, say so and I will do it as its own PR with the digest re-pin and the ADR note together.

suhaanthayyil:
### Checked against #179 for a duplicated helper — not duplicated

This PR and #179 were flagged as reimplementing the same C-family specifier helper, and as being on a collision course. Both halves are wrong, and here is the evidence.

**They are two different predicates on two different tree-sitter fields.**

```go
// this PR
func cFamilyTagReference(node *sitter.Node, language string) bool {
	switch language {
	case "C", "C++", "Objective-C":
	default:
		return false
	}
	return !validNode(node.ChildByFieldName("body"))
}

// #179
func cFamilyAnonymousSpecifier(node *sitter.Node, language string) bool {
	switch language {
	case "C", "C++", "Objective-C":
	default:
		return false
	}
	return !validNode(node.ChildByFieldName("name"))
}
```

`body` vs `name`. The only thing they share is the five-line language gate, which is presumably what made them look alike.

**They disagree in both directions.** Probing both predicates over every `struct_specifier`/`enum_specifier`/`union_specifier` node in a fixture set:

| fixture | node | `cFamilyTagReference` | `cFamilyAnonymousSpecifier` |
|---|---|---|---|
| `int f(struct Ledger *l)` | `struct Ledger` | **true** | false |
| `struct Ledger;` (forward decl) | `struct Ledger` | **true** | false |
| `typedef struct {int x; int y;} Point;` | `struct {…}` | false | **true** |
| `typedef enum {RED, GREEN} Color;` | `enum {…}` | false | **true** |
| `typedef union {int i; float f;} Value;` | `union {…}` | false | **true** |
| `struct Outer { struct {int nested;} anon; };` | inner `struct {…}` | false | **true** |
| `typedef struct Named {int a;} Alias;` | `struct Named {…}` | false | false |

Neither is a subset of the other. A body-less named tag is invisible to #179's check; a bodied anonymous specifier is invisible to this one. The four states are the real taxonomy: name+body is the definition (emit), name-no-body is a tag reference or forward declaration (this PR skips it), body-no-name is an anonymous type whose only name is the typedef's (#179 skips it).

**Collapsing them onto one helper breaks the other fix.** I deduped in each direction on the merged tree and ran the two PRs' tests:

- keep only `cFamilyTagReference` → `TestCAnonymousTypedefStructIsNotNamedAfterItsFirstField` and `TestCPlusPlusAnonymousSpecifiersAreNotNamedAfterMembers` FAIL (`emitted "struct:x", a definition named after a member`). This PR's three tests still pass.
- keep only `cFamilyAnonymousSpecifier` → `TestCHeaderPrototypesEmitNoPhantomStructs` and `TestCPlusPlusTagReferencesAreNotDefinitions` FAIL (`header with no struct body emitted a struct definition at line 1 … line 3 … line 4`; `struct Point has 3 records, want exactly 1`). #179's three tests still pass.
- restore the `||` → all six pass again.

Both mutations compiled, so the failures are behavioural, not build breakage.

**No conflict.** #179 is stacked on `6f601f5f` and already *calls* this helper rather than redefining it — it composes the two with `||` at the same two call sites. Merging `audit/lang-c-anonymous-typedef` into `main` (which carries this PR's commit with it) applies cleanly with the `ort` strategy, no conflicts. On that real merged tree: `go build ./...` clean, the 157 C/struct/typedef/enum/Objective-C tests in `internal/sem` pass, and `TestProviderGoldenSnapshots` / `TestProviderGoldenFixtureQualityCoverageReport` / `TestCapabilityMatrixCoversEmittedRelations` pass with no golden churn.

No change requested here. Recommend landing this one first, then #179.

One unrelated pre-existing gap noticed while probing, on `main` and unaffected by either PR: `typedef struct {int a;} A, *PA;` emits only `type:PA`, dropping `type:A` — a multi-declarator typedef only yields its last declarator. Separate from both fixes.

PR #174, opened 2026-08-28
fix(sem): extract C and C++ data members

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/133

## The defect

`fieldEntities` returned early for every C and C++ `field_declaration`:

```go
if node.Type() == "field_declaration" && (language == "C" || language == "C++") {
	return nil, false
}
```

Neither language produced a single data member. A C struct is the language's only aggregate type and a C++ class is mostly its members, so both extracted as opaque names — no field symbol to search for, no `CONTAINS` edge, no target for a field read.

A realistic C++ header made the size of it clear. Nine members, two types:

```
class Ledger { public: Ledger(); virtual ~Ledger(); int Add(int) const;
               virtual int Total() const = 0; std::string Name() const;
       private: int total_; std::string name_; };
struct Config { int retries; std::string host; };
```

extracted three symbols: `class Ledger`, `method Total` (only because `= 0` makes it a different node type), and `struct Config`.

The opt-out arrived in 6f567b81 ("speed up large repo snapshots") — the commit that first threaded `language` through `walkEntities`. It carries no comment and no test, and removing it changes exactly one golden, additively (`Buffer.data`, `Entry.key`, and their `CONTAINS`/`DEFINES`).

## Three parts

1. **Drop the opt-out.** An in-class method *declaration* is also a `field_declaration` in tree-sitter-cpp; it names nothing through `fieldDeclNames`, so it is still left to the entity walk rather than becoming a field. There is a test for that.

2. **Unwrap the declarator chain.** `char *data;`, `char name[32];` and `int (*on_flush)(int);` carry the name under a pointer/array/function-pointer declarator rather than beside the type, so they were dropped even once the pass ran. A plain function declarator still names nothing.

3. **Classify an initialised C++ member correctly.** tree-sitter-cpp parses `int total_ = 0;` as a `function_definition` whose body is a `pure_virtual_clause` — `= 0` is also the pure-virtual marker and the grammar cannot tell them apart. Taken at face value it produced:

   ```
   method  Ledger.total_  L6  sig="int total_ = 0"
   ```

   a method named after a data member, which then joined the class's method inventory and competed for receiver call resolution. The tell is the absence of any `function_declarator`; a genuine pure virtual keeps its method kind.

Found by the per-language extraction audit; the fixture matrix in #171 covers C and C++.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .`, full `go test ./...` (3760 tests) green. Goldens and the fixture quality-coverage report regenerated; both diffs are additive.


suhaanthayyil:
All four audit branches (#171, #172, #173, #174) merge together cleanly onto `main`; the combined tree builds and passes the full suite (3841 tests).

ashtom:

**Verdict: changes required — 3 open review findings are unaddressed**

- Live PR head `fc39ba07`; current `main` `bb41f223`; **1 ahead, 13 behind**.
- **Every green check on this PR ran against `fc39ba07` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

The change itself passes every check I ran: clean merge, green packages on the merged tree, and its
tests fail without it.

**1. (blocking) Trail #133 carries 3 open findings against this PR that have not been addressed.**
Re-queried immediately before posting:

```
$ entire trail finding list --trail 133 --status any --freshness any
01M15AHMH520  Low     open  current  internal/sem/parser.go:4192
  The fixed eight-step limit drops valid deeply nested declarators.
01M15A9WKRRH  Medium  open  current  internal/sem/parser.go:4147
  Extracting a declarator here causes the enclosing `field_declaration` to short-c...
01M15A9WKGMZ  Medium  open  current  internal/sem/parser.go:4140
  The new dispatch omits `reference_declarator`, although `cFamilyMemberDeclarator...`
```

The two Medium findings are about the declarator dispatch this PR adds, which is the core of the
change. Please resolve or explicitly reject them before merge.

**2. (note, §6) Symbol-set change + merge order.** This adds `field` symbols for C/C++ data members
(golden: `field:Entry.key`, `field:Buffer.data`; `field` 20→22, `CONTAINS` 73→75, `DEFINES` 178→180),
so it changes what consumers see. It conflicts with #170:

```
$ git merge-tree --write-tree refs/prs/170 refs/prs/174   ->  CONFLICT
```

and **#177 is stacked on this PR's commit `fc39ba07`**, so this lands first. See #170 for the
broader point about eight C-family PRs churning symbol identity one merge at a time.

### Independent review state

- Entire trail **#133** exists but is still `phase=building` — the agent review has **not completed**.
- It nonetheless carries **3 open, unaddressed finding(s)** (`entire trail finding list --trail 133 --status any --freshness any`). Please work those before merge.

suhaanthayyil:
All three review findings reproduce and are fixed in `121353cc`. Each fixture below was indexed with `TreeSitterParser{}.Parse`; the symbol lists are the actual output.

### `parser.go:4140` — missing `reference_declarator` (Medium) — **VALID**

Fixture `class Window { int &value; Widget &&item; int plain; };`

| | extracted |
|---|---|
| before | `class:Window`, `field:Window.plain` |
| after | `class:Window`, `field:Window.plain`, `field:Window.value`, `field:Window.item` |

`cFamilyMemberDeclaratorName` already stepped through a `reference_declarator`; the dispatch that calls it never sent one. tree-sitter-cpp uses `reference_declarator` for both `&` and `&&`, so lvalue and rvalue reference members were both dropped.

Fix: `internal/sem/parser.go:4140` — `reference_declarator` added to the case list.
Test: `TestCFamilyReferenceMembersAreExtracted`.
Mutation: removing `reference_declarator` from the case list → `FAIL`; restored → `ok`.

### `parser.go:4147` — inline type body short-circuited (Medium) — **VALID, and a regression against main**

Fixture `struct Outer { struct Inner { int value; } inner; int other; };`

| | extracted |
|---|---|
| main (`8935cfe9`) | `struct:Outer`, `struct:Inner` |
| `fc39ba07` | `struct:Outer`, `field:Outer.inner`, `field:Outer.other` |
| after | `struct:Outer`, `struct:Inner`, `field:Outer.inner`, `field:Outer.other`, `field:Inner.value` |

The member pass stops the walk at the declaration, and a C/C++ declaration can *define* its type inline. `Inner` had a symbol on main and lost it here. The C++ form (`class Inner { public: int value; } inner;`) behaved identically.

Fix: `internal/sem/parser.go` — `walkEntitiesScoped` descends into `cFamilyInlineTypeDefinitions(node, language)` after emitting the fields, at the same scope and `inFunc` the walk used before. Only *named* definitions with a body: an anonymous aggregate (`union { int i; float f; } value;`) declares no type, and its members are reached through `packet.value.i`, so hoisting them onto the enclosing type would file them where nothing reaches them.

Tests: `TestCFamilyInlineTypeDefinitionSurvivesMemberExtraction` (C and C++ subtests), `TestCFamilyAnonymousInlineAggregateIsNotHoisted`.
Mutations: dropping the descent → `FAIL`; dropping the *named* requirement → the anonymous test `FAIL` (it hoists `field:Packet.i`/`field:Packet.f`); restored → `ok`.

### `parser.go:4192` — fixed eight-step limit (Low) — **VALID**

Fixture `struct Deep { int *one; ... int ********eight; int *********nine; };`

| | extracted |
|---|---|
| before | `one` … `seven` (`eight` and `nine` silently dropped) |
| after | `one` … `nine` |

Not refuting this one: tree-sitter models each `*` as its own `pointer_declarator`, so the budget is a cap on pointer depth, and it is a cap that drops the member with no diagnostic. The bound was also protecting nothing — every branch already moved to a strict descendant of the current node, and a parse tree is finite.

Fix: `internal/sem/parser.go` — the budget is replaced by `descendsStrictly(node, next)`, which requires each step to narrow the byte span. That is the termination argument made explicit; a step that fails to narrow gives up instead of spinning.
Test: `TestCFamilyDeeplyNestedDeclaratorIsExtracted`.
Mutation: restoring `depth < 8` → `FAIL`; restored → `ok`.

### Goldens

**No change.** `multilang-relations.ndjson.golden` and `quality_coverage.json` are byte-identical — no fixture holds a reference member, a named inline aggregate, or a declarator that deep. `TestProviderGoldenSnapshots` and `TestProviderGoldenFixtureQualityCoverageReport` pass without regeneration. (The golden delta from this PR is still the one `fc39ba07` recorded; these fixes add nothing to it.)

### Verification

`gofmt` clean, `go vet ./internal/sem/` clean, `go build ./...` clean. The C-family suite (`TestCStruct*`, `TestCPlusPlus*`, `TestCFamily*`, `TestObjectiveC*`, `TestTreeSitterParserCPlusPlus*`, `TestBuildRelationsResolvesCPlusPlusSameFileOverloadSet`) and the golden/snapshot tests pass.

ashtom:

Head moved `fc39ba07` → `121353cc`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

The three findings I quoted last time are resolved by `121353cc`. Two new ones took their place on
the new declarator code — both about valid C/C++ member forms the walk still misses.

```
01M17H7MAV7V  Low     open    current    internal/sem/parser.go:4150  Valid parenthesized C/C++ members such as `int (value);` have a direct `parenthe…
01M17H7MADMD  Medium  open    current    internal/sem/parser.go:4187  The bare-field/function-declarator assumptions miss valid pointer fields initial…
```

### Verified vs. assumed

suhaanthayyil:
## Both findings addressed — pushed `4ae053e2`

---

### 1. `internal/sem/parser.go:4150` — parenthesized members — **VALID**

The C-family switch in `fieldDeclNames` listed `pointer_`, `array_`, `reference_` and `function_declarator` but not `parenthesized_declarator`. Redundant parentheses are legal around any declarator, so it is a shape the declaration itself can carry, not only one reached through a pointer.

Confirmed against the real AST rather than assumed — inside a struct/class body tree-sitter-c **and** -cpp both emit a *direct* `parenthesized_declarator` child (no function-declaration reinterpretation, which was the counter-hypothesis worth checking):

```
field_declaration  "int (value);"
  primitive_type  "int"
  parenthesized_declarator  "(value)"
    field_identifier  "value"

extracted: [struct:S @1]        <- field:S.value MISSING
```

Same for `class S { int (value); }`, and for `int (*(px));`. `cFamilyMemberDeclaratorName` already handled the shape; it was simply never dispatched to.

**Fix:** `parser.go:4150-4151` — add `"parenthesized_declarator"` to the case. After: `field:S.value` and `field:S.px` present in both C and C++.

---

### 2. `internal/sem/parser.go:4187` — bare-field/function-declarator assumptions — **VALID**, and wider than the summary

`cPlusPlusInitialisedMemberName` carried two assumptions:

1. `firstDescendantOfType(node, "function_declarator") → return ""` (4187)
2. `firstNamedChildOfType(node, "field_identifier")` (4190) — which requires a **bare** name

tree-sitter-cpp parses every initialised in-class member as `function_definition` + `pure_virtual_clause`, and `maskCPlusPlusUnsupportedSyntax` rewrites `= {}` to `= 0 `, so array and brace-initialised members land there too (verified: `"int a[3] = {};"` → `"int a[3] = 0 ;"`). Every non-bare form was therefore published as a **method named after a data member**:

```
int *p = 0;            method:S.p      sig="int *p = 0"
int **pp = 0;          method:S.pp     sig="int **pp = 0"
int a[3] = {};         method:S.a      sig="int a[3] = {}"
int (value) = 0;       method:S.value  sig="int (value) = 0"
int (*cb)(int) = 0;    method:S.cb     sig="int (*cb)(int) = 0"
```

Only `int x = 0;`, `static int n = 0;` and `std::string s = {};` came out right.

What separates a data member from a pure virtual is the **declarator**, not the presence of a `function_declarator` anywhere below: `int (*cb)(int) = 0;` is a function-pointer member and `virtual int *Get() = 0;` is a pure virtual, and both contain one.

**Fix:** `parser.go:4183-4206` — drop both assumptions; scan direct named children, take a bare `field_identifier`, otherwise run the member pass's own `cFamilyMemberDeclaratorName` over the declarator, which names a function-*pointer* declarator and refuses a plain one.

**Supporting fix, required for the above to be safe:** `parser.go:4232` plus new `firstChildDeclarator` at `4254`. `parenthesized_declarator` and `reference_declarator` carry no `declarator` field, so the unwrapper fell back to `firstDescendantOfType(node, "field_identifier")` — a blind descent that steps **over** a nested `function_declarator`. Without repairing that, fix 2 turns `virtual int &Get() = 0;` and `virtual int (*H())(int) = 0;` into data members. It also removes a latent bug already on this branch: `int (*Method())(int);` was being filed as `field:S.Method`.

---

### Tests

- `TestCFamilyParenthesizedMemberIsExtracted` (C and C++ subtests).
- `TestCPlusPlusInitialisedNonBareMembersAreFields` — six forms, asserting the field is present **and** that no method of that name exists.
- `TestCPlusPlusPureVirtualWithWrappedDeclaratorStaysAMethod` — `Plain`, `*Pointer`, `&Reference`, `(*Callback())` must stay methods and must not appear as fields.

### Mutation

Three hunks, three mutations, each grep-confirmed applied and each compiling. Each failed **only** its own guard:

| revert | FAIL |
|---|---|
| 4150 back to the four-type case | `missing "field:Wrapped.value"` → `TestCFamilyParenthesizedMemberIsExtracted/C` and `/C++` |
| old `cPlusPlusInitialisedMemberName` body | `missing "field:Config.pointer_"` + `data member published as a method` (×5) |
| `firstChildDeclarator(node)` → `firstDescendantOfType(node, "field_identifier")` | `pure virtual lost its method kind, missing "method:Iface.Reference"` + `pure virtual published as a field` |

Restored → all PASS. Re-verified independently after the fact: mutations 1 and 3 each fail only their own guard and the tree restores clean.

### main vs branch

Three-way comparison on a 35-form C/C++ corpus (15 C member forms, 20 C++, plus a typedef-struct header), run in an isolated worktree at `origin/main`.

- **added by this change (7):** `field Buffer.value`, `field Buffer.indirect`, and `field Ledger.{pointer_, deep_, array_, parens_, callback_}`
- **removed by this change (6):** the five wrong `method Ledger.*` (now fields), plus `field Buffer.Method` for `int (*Method())(int);` — a method declaration that was wrongly filed as data, and for which main emits no symbol either, so this matches main.
- **in main but not on the branch (10), none attributable to this change:** seven `method Ledger.*` — the intended method→field conversion this PR exists for — and three phantom duplicate `struct` symbols (`Buffer 6-6`, `Node 2-2`, `Node 3-3`) that main emits when it walks into `struct Buffer *next;` and hits the bodyless elaborated type reference. Verified pre-existing by re-running the same dump against this branch's head with the new fixes stashed: identical set. The real `struct Buffer 1-16`, `struct Node 1-6` and `type Node` all survive.

**No regression attributable to this round.**

### Golden delta

**None.** `TestProviderGoldenSnapshots` and the coverage report pass unmodified, `testdata/` clean. The only C/C++ fixtures are `multilang-relations/{buffer.c,Cache.cpp}` and neither contains a parenthesized declarator or a non-bare initialised member.

### Also checked

LF and CRLF give the same symbol set and the same line numbers (the declarator walk uses relative byte containment and names are trimmed). 68 tests across `TestCFamily*|TestCPlusPlus*|TestCStruct*|TestCpp*|TestTreeSitterParserCPlusPlus*|TestProviderGoldenSnapshots|TestFastCFamilyEntitiesTopLevelInventory|TestBuildRelationsResolvesCPlusPlusSameFileOverloadSet` pass, plus 18 cross-language field tests — `parenthesized_declarator` is a C-family node type and `firstChildDeclarator` is reachable only from `cFamilyMemberDeclaratorName`, so Go/Java/C#/Kotlin/TS are untouched. `gofmt`/`go vet` clean.

### Left open, pre-existing on this branch and not touched

- `int (T::*mp)(int);` gets signature `"mp int (T"` because the C++ mask rewrites pointer-to-member syntax.
- `int (*Method())(int);` — a method returning a function pointer — yields no symbol in C or C++, same as main.
- Elaborated type references inside member declarations (`struct Node *left;`) no longer emit their duplicate `struct` symbol, as above.

suhaanthayyil:

- TestCFamily / TestCPlusPlus / TestCStruct, plus `TestProviderGoldenSnapshots` — pass on the merged tree

suhaanthayyil:
Two results from working the trail findings on this PR.

### `01M1JC78ANPW` (anonymous enum members) — dismissed as refuted

Verified three ways:

| fixture | `origin/main` | this branch |
|---|---|---|
| `struct S { enum { A, B } value; int other; };` | `struct:S` | `struct:S`, `field:S.other`, `field:S.value` |
| `enum Color { RED, GREEN };` | `enum:Color` | `enum:Color` |

The branch is **additive only** — nothing is lost. Ablating `enum_specifier` out of the switch at `parser.go:7717` changed output across 7 C/C++ fixtures by exactly one line, in the *added* direction, losing zero entities: the walk stops at the field because `fieldEntities` returns `ok` and `walkEntitiesScoped` returns (`parser.go:3677`/`3719`), so the cited line has no causal role. And C/C++ enumerators are extracted **nowhere** in this parser on either side, so "preserves their named constants" describes a feature that has never existed.

The false positive came from the branch being **140 commits behind `main`**: at the merge-base an anonymous enum still produced a phantom entity named after its first *enumerator*, and the finding measured against that. `main` deleted it independently in #179 (`c5620ed6`), whose commit message names this exact case.

### A real one, though — please rebase before merging

Because the branch predates #179, it still carries that phantom:

```
typedef enum { RED, GREEN } Color;

origin/main   ->  [type:Color]
this branch   ->  [type:Color, enum:RED]     <-- names nothing that exists
```

`enum:RED` is an enum named after its first enumerator. This resolves on rebase onto `main`, not by any edit here — but it is worth doing before merge, since this PR touches the same anonymous-aggregate area (`cFamilyAnonymousAggregateBodies`) that #179 changed (`cFamilyAnonymousSpecifier`), so the two are worth reconciling deliberately rather than by merge resolution.

PR #176, opened 2026-08-29
fix(sem): type C++ locals declared without an initialiser

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/134

Stacked on #171 (`audit/lang-matrix`), which pins this gap as an audited `callsGap` entry. Basing on `main` instead would have left that entry describing a cause that no longer holds.

## The defect

The generic local-variable scanner (`localVarTypes`) types a local only from an assignment whose right-hand side names the type: `x = new T(...)` or `x = T(...)`. C++'s most common way to create a local names the type on the *left* of the name and assigns nothing at all:

```cpp
Ledger ledger;          // default construction
Ledger ledger(seed);    // direct initialisation
Ledger ledger{seed};    // brace initialisation
const Ledger& ledger = other;
```

None of those are constructor assignments, so the receiver carried no inferred type and every `ledger.Add(...)` on it resolved to nothing — while `Ledger l = Ledger();` and `new Ledger()` both resolved. The hole was in reading declarations, not in C++ receiver resolution.

Nothing about this is loud. The symbols are all still emitted, the file still parses, `capabilities --json` still advertises `CALLS` for C++. Only the edges go missing.

## The fix

A C++-gated declaration scanner (`cppLocalVarTypes`) feeding the same `localTypes` map the Kotlin, Swift, TypeScript, Java and Rust scanners already extend.

It stays narrow deliberately, because a *wrong* receiver type resolves calls to the wrong symbol, which is worse than the missing edge this exists to supply. A declaration must begin a statement (line start, `;`, `{`, `}`), its type must be capitalized, and it must end at a declarator terminator (`;`, `(`, `{`, `=`). That excludes parameters and defaulted parameters, catch clauses, template arguments, wrapped call arguments, local type definitions, and string/comment content. A name declared twice with two different types is dropped rather than guessed at, matching the Kotlin and Swift scanners.

## Verification

`TestCPlusPlusDefaultConstructedReceiverResolvesCalls` asserts the edge **and its resolution**: `type_inferred` is the only path that reads the receiver's type, so a bare-name fallback resolving the same edge by luck cannot make it pass.

`TestCPlusPlusDeclaredLocalScannerStaysNarrow` pins the seventeen shapes the scanner must and must not read.

Four mutations, each killed by a different test case:

| mutation | result |
|---|---|
| scanner disabled at the provider hook | `TestCPlusPlusDefaultConstructedReceiverResolvesCalls` fails |
| statement anchor widened to `(` and `,` | `/defaulted_parameter` fails |
| capitalization guard dropped | `/lowercase_type` and `/template_argument` fail |
| conflicting-declaration drop removed | `/conflicting_declarations` fails |

`go test ./...`, `go vet ./...` and `gofmt -s -l` are clean. No golden fixture moved.

## What this does not close, and why the matrix entry stays

The C++ `callsGap` in `languageMatrix` remains, with its reason corrected to what now actually causes it. The matrix fixture is laid out the way C++ is normally written — the method is declared in the class body and defined out of line as `int Ledger::Add(...)`:

- the in-class declaration is not extracted at all, and
- the out-of-line definition is emitted as a top-level `function:Add`, not `method:Ledger.Add`.

So the newly typed receiver has no container-qualified method to resolve to. A class with an **inline** method body resolves end to end, which is exactly what isolates the residue to method-symbol qualification rather than to receiver typing. That is a separate defect and gets a separate change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,12 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-go@v6
         with: { go-version: "1.26.x" }
-      - run: go test ./...
+      # internal/sem runs ~550s on the Windows runner against Go's 600s default
+      # per-package wall, with run-to-run swings of about 40s. That is 92% of an
+      # unset budget: the suite was one ordinary new test away from failing for a
+      # reason unrelated to the change that tripped it, which is what happened.
+      # Name a budget with headroom instead of inheriting one nobody chose.
+      - run: go test -timeout 30m ./...
 
   bm25-adapter:
     runs-on: ubuntu-latest
--- a/docs/language-support.md
+++ b/docs/language-support.md
@@ -58,6 +58,20 @@ Current generated counts:
 - Zig
 - Zsh
 
+Every language in this list has a minimal fixture under
+`internal/sem/testdata/langmatrix/` and is asserted by
+`TestLanguageExtractionMatrix` to extract real symbols and relations. A grammar
+that loads and parses without error but yields nothing is invisible in
+aggregate symbol counts, so the per-language assertion is the only thing that
+catches it. `TestLanguageMatrixCoversEverySemanticLanguage` keeps that fixture
+set in lockstep with the registry, and `TestDocumentedLanguageListsMatchTheRegistry`
+keeps both lists on this page in lockstep with the registry.
+
+Languages that advertise `CALLS` but extract none from their fixture's
+idiomatic call site are pinned in `languageMatrix` with the reason. That list is
+enforced in both directions: a language going dark fails, and so does a fix
+landing without the audited gap being retired.
+
 Protocol Buffers support covers proto3 and legacy proto2 declarations,
 including files that omit the syntax declaration, proto2 field labels, and
 groups. Compatibility parsing preserves original source locations, signatures,
--- a/docs/semantic-provider-requirements.md
+++ b/docs/semantic-provider-requirements.md
@@ -207,6 +207,13 @@ When a move has multiple equally similar destinations (within 0.05), the
 provider reports the pair as remove/add and emits a `W_MOVE_AMBIGUOUS` warning
 in the diff `warnings` array rather than guessing.
 
+A file rename whose content is byte-identical on both sides has no entity-level
+delta to reconcile, but the file path is a component of every symbol ID in it,
+so the rename still re-identifies every symbol. Such a file is reported with a
+single module-scope `moved` change carrying `old_path`/`new_path` and
+`reconciliation: MOVED`, so a pure rename is never indistinguishable from an
+empty diff.
+
 ## Relations
 
 Relations should include:
--- a/internal/cli/search.go
+++ b/internal/cli/search.go
@@ -468,11 +468,15 @@ func writeNdjsonSearch(out interface{ Write([]byte) (int, error) }, response sem
 		encoder.SetEscapeHTML(false)
 		if err := encoder.Encode(map[string]any{
 			"record_type": "search_header",
-			"query":       response.Query,
-			"repo_root":   response.RepoRoot,
-			"commit":      response.Commit,
-			"tree":        response.Tree,
-			"profile":     response.Profile,
+			// The header record carries the envelope version for the same reason
+			// the JSON payload does: a consumer reading the stream must be able to
+			// branch on the shape before it parses a single result record.
+			"format_version": response.FormatVersion,
+			"query":          response.Query,
+			"repo_root":      response.RepoRoot,
+			"commit":         response.Commit,
+			"tree":           response.Tree,
+			"profile":        response.Profile,
 		}); err != nil {
 			return err
 		}
--- a/internal/cli/search_format_version_test.go
+++ b/internal/cli/search_format_version_test.go
@@ -1,0 +1,71 @@
+package cli
+
+import (
+	"bytes"
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/entireio/entire-graph/internal/sem"
+)
+
+// Every other machine-readable surface tells a consumer which shape it is
+// looking at. Measured on origin/main against one fixture repository:
+//
+//	schema_version : diff, commit, capabilities, symbols, edges, snapshot,
+//	                 compact-ndjson
+//	format_version : impact, neighbors, def, index, stats
+//	NEITHER        : search --format json, search --format ndjson
+//
+// search is the documented entry point ("start here") and JSON is its DEFAULT
+// format, so the one surface a consumer is most likely to parse was the one it
+// could not version-check. These tests keep the discriminator on both formats.
+func TestSearchJSONCarriesFormatVersion(t *testing.T) {
+	t.Parallel()
+	var out bytes.Buffer
+	response := sem.SearchResponse{FormatVersion: sem.SearchFormatVersion, Query: "q", Results: []sem.SearchResult{}}
+	if err := writeSearchResponse(&out, response, "json", 0); err != nil {
+		t.Fatal(err)
+	}
+	var payload struct {
+		FormatVersion *int `json:"format_version"`
+	}
+	if err := json.Unmarshal(out.Bytes(), &payload); err != nil {
+		t.Fatalf("decode search json: %v\n%s", err, out.String())
+	}
+	if payload.FormatVersion == nil {
+		t.Fatalf("search json carries no format_version:\n%s", out.String())
+	}
+	if *payload.FormatVersion != sem.SearchFormatVersion {
+		t.Fatalf("format_version = %d, want %d", *payload.FormatVersion, sem.SearchFormatVersion)
+	}
+}
+
+func TestSearchNDJSONHeaderCarriesFormatVersion(t *testing.T) {
+	t.Parallel()
+	var out bytes.Buffer
+	response := sem.SearchResponse{FormatVersion: sem.SearchFormatVersion, Query: "q", Results: []sem.SearchResult{}}
+	if err := writeNdjsonSearch(&out, response); err != nil {
+		t.Fatal(err)
+	}
+	first, _, found := strings.Cut(out.String(), "\n")
+	if !found {
+		t.Fatalf("ndjson stream has no header line:\n%s", out.String())
+	}
+	var header struct {
+		RecordType    string `json:"record_type"`
+		FormatVersion *int   `json:"format_version"`
+	}
+	if err := json.Unmarshal([]byte(first), &header); err != nil {
+		t.Fatalf("decode ndjson header: %v\n%s", err, first)
+	}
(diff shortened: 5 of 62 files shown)


ashtom:

**Verdict: changes required — 1 open review finding is unaddressed**

- Live PR head `caa4b6bb`; current `main` `bb41f223`; **2 ahead, 13 behind**.
- **Every green check on this PR ran against `caa4b6bb` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

**1. (blocking) Trail #134 carries an open finding against this PR.** Re-queried before posting:

```
$ entire trail finding list --trail 134 --status any --freshness any
01M15H2VKVRJ  Medium  open  current  internal/sem/provider.go:4775
  This function-wide map can overwrite a typed parameter when a nested scope shado...
```

Shadowing in a function-wide type map is exactly the kind of thing that produces a confidently wrong
edge rather than a missing one, so it is worth resolving rather than deferring.

**2. (note) This PR is stacked on #171** (`2c599378`, the language matrix), so #171 needs to land
first or this needs rebasing. The `internal/sem/testdata/langmatrix/**` files in this diff belong to
#171, not to this change — the actual change here is `internal/sem/cpp.go` (79 lines) plus 11 lines
of `provider.go`.

### Independent review state

- Entire trail **#134** exists but is still `phase=building` — the agent review has **not completed**.
- It nonetheless carries **1 open, unaddressed finding(s)** (`entire trail finding list --trail 134 --status any --freshness any`). Please work those before merge.

suhaanthayyil:
The trail finding is **VALID**, and it understates the severity: this was a regression that produced a **confidently wrong edge**, not a missing one. Fixed in `b29fa548`.

## Reproduction

```cpp
class Ledger  { public: int Commit(int a) const { return a; } };
class Account { public: int Commit(int a) const { return a; }
                        int Deposit(int a) const { return a; } };

int Post(Ledger ledger, int amount) {
    if (amount > 0) {
        Account ledger;              // nested-scope shadow
        ledger.Deposit(amount);
    }
    return ledger.Commit(amount);    // `ledger` is the Ledger PARAMETER
}
```

On this branch, before the fix:

```
CALLS Post -> Account.Commit   confidence 0.85  resolution=type_inferred
```

The correct edge to `Ledger.Commit` was **absent**. Both classes declare `Commit`, so the wrong edge points at a real symbol and nothing downstream can tell it apart from the right one.

**It is a regression, and that is measurable.** `cppLocalVarTypes` does not exist on `main`, and `main` resolves the identical fixture correctly:

```
# origin/main, same fixture
CALLS Post -> Ledger.Commit    resolution=type_inferred     ← correct
```

So the scanner this PR adds did not fail to close a gap; it opened one.

## Mechanism

`localTypes` is function-wide, and `provider.go:4821` copies it into `varTypes` **unconditionally** — no `exists` check — after which `provider.go:4965` deletes every local name from `paramTypes`. The existing `if _, exists := localTypes[name]; !exists` guard at the C++ block only protects against collisions *within* `localTypes`; parameters live in `varTypes` and were never consulted. So one nested `Account ledger;` retyped the parameter for the entire function.

## Fix

The guard is a rule of the language, not a heuristic: **C++ forbids redeclaring a parameter in the function body's top-level scope**, so a declared local sharing a parameter's name is *necessarily* in a nested block and shadows the parameter only there. The parameter therefore wins — its type is declared, the local's is inferred.

Cost, stated plainly: calls on the shadowing local *inside* the block now go unresolved (`ledger.Deposit` produces no edge) until receiver typing is scope-aware. That is the right trade — a missing edge beats a confidently wrong one, which is the whole reason this finding mattered.

## Mutation verify (both compile)

| mutation | result |
|---|---|
| remove the parameter guard | **FAIL** — `TestCPlusPlusNestedShadowDoesNotRetypeAParameter` |
| over-broad guard: skip *every* declared local | **FAIL** — `TestCPlusPlusDeclaredLocalStillTypesNonParameterNames` |
| unmutated | pass |

The second mutation is the one that matters for confidence: a guard that simply suppressed all declared locals would undo this PR's entire feature and still pass the shadowing test. Two tests pin the guard from both sides.

`go test ./internal/sem/ -run 'TestCPlusPlus|TestCpp|TestLangMatrix|TestProviderGolden'` — pass. No golden change.

## On the stacking note

Agreed that this is stacked on #171 (`2c599378`) and that the `internal/sem/testdata/langmatrix/**` files in the diff belong there, not here. The substantive change on this branch remains `cpp.go` plus the `provider.go` hunk, now +1 guard.

suhaanthayyil:
## Silent merge landmine: this branch + #177

These two merge cleanly and then fail. Each passes with #171 on its own — this branch types the
receiver, #177 qualifies the method symbol, and the C++ CALLS gap needs both closed — so the
failure exists only in the combined tree, and neither branch's CI can see it:

```
$ git checkout -q 2c599378 && git merge --no-edit <this branch> && git merge --no-edit <#177>
    languageMatrix
```

The re-audited cause on this branch is precisely right, and #177 is what closes it: without #177 the
cpp fixture emits `function:Add`; with #177 it emits `method:Add`.

## One-line fix, order-independent either way

#171 now carries a `callsGapHolds` field — the condition a gap's stated cause reduces to, read from
the extraction on every run, so an audit entry retires itself in whichever commit closes it. This
branch's cause is fixture-observable, so it can state it directly:

```go
callsGap:      "a C++ method declared in the class body and defined out of line as `Type::method` produces no container-qualified method symbol, so a typed receiver has nothing to resolve to",
```

Verified in both directions after rebasing this branch on #171's current head:

| tree | with that one line |
|---|---|
| #171 + this branch (gap still open) | pass |
| #171 + this branch + #177 (gap closed) | pass |

Without it, whichever of the two lands second has to remove the `callsGap` line at
`internal/sem/language_matrix_test.go` in the cpp entry — and that edit is wrong until the other one
has landed, which is what makes this a landmine rather than a conflict.

suhaanthayyil:
Landed the cpp gap fix I verified but did not push last round — leaving a verified fix unlanded was the wrong call, since it left exactly the shape I had just called the worst kind. `dfa72a7e`.

## The landmine, reproduced

#176 and #177 each merge cleanly and each is green alone. On the tree that has both:

```
    C++ now extracts CALLS; remove its audited gap (...) from languageMatrix
```

**Correcting my own earlier diagnosis:** I had said "#176+#177 together close the gap". More precisely, **#176 alone does not break it** — I checked, `#176 + #171` is green. #176 closes the *first* cause (a default-constructed stack receiver carries no inferred type) and re-documents the remaining one; **#177 closes that second cause** and the gap falls. It takes #171 + #174 + #176 + #177.

## The fix, using the mechanism built for #161↔#171

The cpp cause reduces to a **missing symbol**, not to a resolution rule, so the condition reads the fixture's own extraction rather than a predicate over the provider:

```go
```

Measured: on this branch the cpp fixture yields `class:Ledger`, `function:Add`, `function:LedgerDouble`, `method:total_`, and **no CALLS**. Once in-class declarations are extracted it yields `method:Add` **and** CALLS, and the entry retires itself.

**It lives here, not on #177, and that is not arbitrary** — this branch is stacked on the matrix and #177 is not, so this is the only one of the two that can carry the condition at all. **#177 therefore needs no edit and no merge-order note**: whichever order they land in, the condition is read from the extractor at run time.

Naming the *missing symbol* rather than the missing CALLS is deliberate: it claims the cause, so a change that produced CALLS by another route fails here instead of quietly retiring a gap whose stated reason was never addressed.

## Verification

| tree | matrix tests |
|---|---|
| #176 + #171 (gap holds) | pass |
| #171 + #174 + #176 + #177 (gap retires) | pass |
| merged tree, predicate removed | **FAIL** — the exact landmine above |
| merged tree, predicate names a symbol that never appears | **FAIL** — a wrong cause is caught |

`go test ./internal/sem/` — **pass, 77.3s**, full package. This branch now also carries a merge of current #171, which it was already stacked on.

## Sweep for the rest of the class

Since a clean `merge-tree` demonstrably does not catch these, I merged #171 with **every** open PR that touches the extractor or the matrix — the only PRs that can close a gap — and ran the matrix tests. Two audited gaps remained literal after the Julia fix:

- **fsharp** — no landmine. **#178 already removes its own entry**, which is the correct pattern and needed no help.
- **objc** — no landmine. #172 and #183 both merge green; neither does local type inference, so the gap correctly still holds.

Also clean: #161, #162, #165, #168, #181, each merged with #171.

I did **not** run all 231 pairs. The matrix test can only break when a gap *closes*, and only an extractor change closes one, so the targeted set is the complete set for this class — that is a reasoned scope, not a shortcut, and I would rather say so than imply exhaustive coverage.

ashtom:

Head moved `caa4b6bb` → `dfa72a7e`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

`b29fa548` addresses the shadowing finding for parameters. The replacement finding says the guard
only covers parameters, so a nested **local** can still retype a same-named symbol — the same class
of problem one level out.

```
01M17GN8WJCR  Medium  open    current    internal/sem/provider.go:4792  This guard only protects parameters, so a nested local can still retype a same-n…
```

### Verified vs. assumed

suhaanthayyil:
**REFUTED as stated**, with one part of it standing as a pre-existing recall gap rather than a defect in this PR.

The finding says a nested local "can still retype a same-named local", which describes the dangerous class — a confidently wrong edge. That is not what happens. Measured on this branch:

```cpp
int Post(int amount) {
    Ledger ledger;                       // outer local
    if (amount > 0) {
        Account ledger;                  // nested local, different type
        ledger.Deposit(amount);
    }
    return ledger.Commit(amount);
}
```

```
TOTALCALLS 0
```

No edge is produced at all — not a wrong one. `cppLocalVarTypes` (`cpp.go:60-73`) already tracks a `conflicted` set: a name declared twice with different types is **deleted** from the map and marked conflicted, so it carries no type and resolves to nothing. The parameter guard is not what is protecting this case; the conflict logic is, and it predates the parameter guard.

**And it is not a regression.** The same fixture on `origin/main` gives `TOTALCALLS 0` as well, so this PR neither introduced nor worsened it.

What is left standing is a **recall** gap: `ledger.Deposit` inside the block and `ledger.Commit` after it both go unresolved, where a scope-aware receiver typing would resolve each to its own type. That is worth recording, and it is the same limitation I named when the parameter guard landed — "calls on the shadowing local inside the block go unresolved until receiver typing is scope-aware". It is a missing edge, which is the safe failure, and closing it means real scope tracking rather than a wider guard.

So: no change here. If you would rather the conflicted-name case emit a low-confidence edge to the nearest declaration instead of nothing, that is a deliberate recall/precision trade and I would want it decided explicitly rather than folded into this PR — the whole reason the parameter guard exists is that a confidently wrong receiver edge is worse than a missing one.

PR #177, opened 2026-08-29
fix(sem): extract C++ in-class method declarations

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/135

C++ classes that declare methods in headers previously contributed no method symbols for those declarations. This extracts bodyless in-class methods, constructors, destructors, templates, operator overloads, and comma-separated declarations while keeping function-pointer data members classified as fields.

Calls through declarations now redirect to a unique out-of-line definition using AST names and lexical owners, including nested classes and inline namespaces. Ambiguous overloads remain conservative, and inherited calls retain their confidence and resolution metadata. Declaration signatures and type metadata preserve template constraints and shared specifiers.

Targets main directly. The former parent, PR #174 / trail 133 (C and C++ data members), has merged. Current main is integrated without conflicts while preserving the branch's existing commits and checkpoints.

Known limitation: C++ declaration IDs use signature hashes to survive overload insertion and reordering; signature edits change those IDs. This tradeoff is tracked in #214, and its associated trail finding was explicitly dismissed pending that design decision.

Validation: all 25 hosted checks passed on c31f81f6, including Linux, macOS, Windows, and Entire Gates. Local vet, build, formatting, the focused contract test, C++ regression tests, and all 162 status-line checks passed. Fresh trail review has zero open findings; its template-bracket finding was dismissed with verified caller-precondition evidence. The supplementary local full race run was stopped after hosted CI passed; it is not counted as a local pass. Merged into main as 6c70e7b1.


ashtom:

**Verdict: changes required — 2 open review findings are unaddressed**

- Live PR head `ec490ccb`; current `main` `bb41f223`; **2 ahead, 13 behind**.
- **Every green check on this PR ran against `ec490ccb` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

**1. (blocking) Trail #135 carries 2 open findings against this PR.** Re-queried before posting:

```
$ entire trail finding list --trail 135 --status any --freshness any
01M15HN8N488  Medium  open  current  internal/sem/parser.go:4211
  Reading only the declaration's single `declarator` field misses additional metho...
01M15HN8MS9J  Medium  open  current  internal/sem/parser.go:4208
  The direct-parent requirement drops member-function templates: in `template<clas...`
```

Both are recall gaps in the in-class declaration extraction this PR adds — i.e. about the change
itself, not its surroundings.

**2. (note, §6) Stacked on #174** (`fc39ba07`), so #174 lands first. Its golden delta is identical
to #174's for that reason. It also conflicts with #170:

```
$ git merge-tree --write-tree refs/prs/170 refs/prs/177   ->  CONFLICT
```

### Independent review state

- Entire trail **#135** exists but is still `phase=building` — the agent review has **not completed**.
- It nonetheless carries **2 open, unaddressed finding(s)** (`entire trail finding list --trail 135 --status any --freshness any`). Please work those before merge.

suhaanthayyil:
Both review findings reproduce and are fixed in `7ea287a9`. The branch is rebased onto the updated #174 head (`121353cc`), which carries the three fixes filed against that PR.

### `parser.go:4208` — direct-parent requirement drops member function templates (Medium) — **VALID**

Fixture:
```cpp
template<class U> U FreeGet();
class Box {
public:
    template<class U> U Get();
    template<class U> U* GetPtr();
    int Plain();
};
```

| | extracted |
|---|---|
| before | `class:Box`, `method:Box.Plain` |
| after | `class:Box`, `method:Box.Plain`, `method:Box.Get`, `method:Box.GetPtr`, `method:Box.GetRef` |

The AST confirms the mechanism: the `field_declaration_list` holds a `template_declaration`, and the `declaration` sits one level inside it, so the direct-parent check returned before reading the declarator. Most of the interface of a container or a traits class is declared this way.

Fix: `internal/sem/parser.go`, `cPlusPlusMemberDeclarations` — one `template_declaration` is stepped through before requiring `field_declaration_list`. A namespace-scope template still has no class body above it: `method:Box.FreeGet` is asserted absent, and the friend/local/prototype exclusions in `TestCPlusPlusMemberDeclarationScannerStaysInClassBodies` still hold.
Test: `TestCPlusPlusMemberFunctionTemplatesAreExtracted`.
Mutation: removing the hop → `FAIL`; restored → `ok`.

### `parser.go:4211` — only the first `declarator` is read (Medium) — **VALID**

Fixture `class Runner { public: void Start(), Stop(); int a, b; };`

| | extracted |
|---|---|
| before | `class:Runner`, `field:Runner.a`, `field:Runner.b`, `method:Runner.Start` |
| after | … plus `method:Runner.Stop` |

The AST shows two `declarator:` children on the one `field_declaration`; `ChildByFieldName` returns the first. As the finding notes, nothing else recovers the rest — the field pass ignores plain function declarators by design, so `Stop` disappeared entirely.

Fix: `internal/sem/parser.go` — every child whose field name is `declarator` is read (`FieldNameForChild`). Because one node now declares two symbols, which `entityFromNode` (one entity per node) cannot express, the member methods are emitted from `walkEntitiesScoped` beside the field pass instead. The walk still descends into the declaration afterwards — it has no body, but a default argument can hold one. Methods sharing a declaration take their own declarator text as their signature (`void Start()` / `void Stop()`) rather than the shared declaration text, so they do not carry the same fingerprint; the single-declarator case keeps `signatureFromNode(node)` exactly as before.
Test: `TestCPlusPlusCommaSeparatedMemberDeclarations` (asserts both methods and that their fingerprints differ).
Mutations: reverting to `ChildByFieldName("declarator")` → `FAIL`; forcing the shared signature → `FAIL` on the fingerprint assertion; restored → `ok`.

### Also carried in this commit

The same fixed-budget defect that #174 was asked to remove existed in this PR's own return-type unwrapper: `int ********Eight();` nests one `pointer_declarator` per `*`, so the walk never reached the `function_declarator` and the method was dropped. Replaced with the same `descendsStrictly` guard.
Test: `TestCPlusPlusDeeplyPointedReturnTypeIsExtracted`. Mutation: restoring `depth < 8` → `FAIL`; restored → `ok`.

### Goldens

**No change.** `multilang-relations.ndjson.golden` and `quality_coverage.json` are byte-identical — no fixture declares a member function template or comma-separated methods. `TestProviderGoldenSnapshots` and `TestProviderGoldenFixtureQualityCoverageReport` pass without regeneration.

### Verification

`gofmt` clean, `go vet ./internal/sem/` clean, `go build ./...` clean. The C-family suite and the golden/snapshot/determinism tests pass.

suhaanthayyil:
## Silent merge landmine: this branch + #176

These two merge cleanly and then fail. Each passes with #171 on its own — #176 types the
receiver, this branch qualifies the method symbol, and the C++ CALLS gap needs both closed — so the
failure exists only in the combined tree, and neither branch's CI can see it:

```
$ git checkout -q 2c599378 && git merge --no-edit <#176> && git merge --no-edit <this branch>
    languageMatrix
```

The re-audited cause on #176 is precisely right, and this branch is what closes it: without this branch the
cpp fixture emits `function:Add`; with it, the fixture emits `method:Add`.

## One-line fix, order-independent either way

#171 now carries a `callsGapHolds` field — the condition a gap's stated cause reduces to, read from
the extraction on every run, so an audit entry retires itself in whichever commit closes it. #176's cause is fixture-observable, so its entry can state it directly:

```go
callsGap:      "a C++ method declared in the class body and defined out of line as `Type::method` produces no container-qualified method symbol, so a typed receiver has nothing to resolve to",
```

Verified in both directions against #171's current head:

| tree | with that one line |
|---|---|
| #171 + #176 (gap still open) | pass |
| #171 + #176 + this branch (gap closed) | pass |

Without it, whichever of the two lands second has to remove the `callsGap` line at
`internal/sem/language_matrix_test.go` in the cpp entry — and that edit is wrong until the other one
has landed, which is what makes this a landmine rather than a conflict.

suhaanthayyil:
**No change needed on this branch, and no merge-order note either** — recording that as a result rather than leaving it implied.

This PR closes the second of the two causes behind the C++ audited CALLS gap in the language matrix, so on the tree that has #171 + #174 + #176 + #177 the matrix test used to fail:

```
    C++ now extracts CALLS; remove its audited gap (...) from languageMatrix
```

Each PR merges cleanly and each is green alone, which is why nothing caught it.

Fixed on **#176** (`dfa72a7e`), not here, for a structural reason: #176 is stacked on the matrix branch and this one is not, so #176 is the only one of the two that can carry the condition. The gap entry is now derived from the fixture's own extraction —

```go
```

— and **this PR is what flips it**. Extracting the in-class declaration makes `method:Add` appear, which retires the audited gap automatically, in either merge order, with no edit to this branch.

Verified on the real merged tree, not `merge-tree`: #171+#174+#176+#177 passes; removing the predicate reproduces the failure above.

One earlier claim of mine corrected while doing this: I had written that "#176+#177 together close the gap" as though either alone was inert. **#176 alone does not break the test** — it closes the first cause (default-constructed receiver typing) and re-documents the remaining one; this PR closes that remaining one.

ashtom:

**Verdict: changes required — 4 open findings, all new since my last review**

Head moved `ec490ccb` → `7ea287a9`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### Entire trail findings — 4 open and current

```
01M17HC6KV4Q  Medium  open    current    internal/sem/parser.go:4333  For member templates this is the inner declaration node, so the emitted signatur…
01M17HC6KJ0D  Medium  open    current    internal/sem/parser.go:4246  Only `template_declaration` is unwrapped before requiring a direct class-body pa…
01M17HC6KA70  Medium  open    current    internal/sem/parser.go:4292  This accepts only a direct name child of the function declarator, so valid decla…
01M17HC6JY5B  Medium  open    current    internal/sem/parser.go:4356  These bodyless header methods occupy the same `methodsByContainer[container][nam…
```

### Verified vs. assumed

suhaanthayyil:
Worked all four findings against `7ea287a9`, merged onto `bb41f223`. Three are real and are fixed in `f1c2ef36`; one is refuted with a fixture.

The most important result: **one of these was a regression this PR itself introduced**, and it only showed up by indexing the same fixture on `origin/main` and on the branch.

---

### `01M17HC6JY5B` — `parser.go:4356`, bodyless header methods shadow the definition — **VALID, and a regression**

The stated *mechanism* is not what happens: the out-of-line definition parses as `function` `Add` with qualified name `Add` and an EMPTY container, so it never enters `methodsByContainer` and nothing is overwritten. The *impact* is exactly right, and worse than described — it is a behaviour change against `main`.

Fixture: `ledger.hpp` declares `int Add(int) const;` in `class Ledger`, `ledger.cpp` defines `int Ledger::Add(int) const {…}`, `main.cpp` calls `l->Add(3)`.

```
main   (bb41f223)  CALLS run -> function:Add @ ledger.cpp:2   conf 0.68  resolution=name_only
branch (7ea287a9)  CALLS run -> method:Add   @ ledger.hpp:4   conf 0.85  resolution=type_inferred
```

Extracting the in-class declaration gave `methodsByContainer[Ledger]["Add"]` an entry, so the typed-receiver path now succeeds — and beats the unique-name fallback that used to find the real code. The call landed on a bodyless declaration: the implementation lost its callers, and impact on it reported nothing.

**Fix** `provider.go:6622` `cPlusPlusOutOfLineDefinition` + the redirect at `provider.go:5177`. When a receiver call resolves to a bodyless C++ declaration, prefer the non-bodyless definition that names this container (`Container::Name` in its signature). Two such definitions are overloads the name-keyed index cannot separate, so it refuses and the declaration stays the honest target.

**Test** `TestCPlusPlusReceiverCallResolvesToTheOutOfLineDefinition` (`parser_cpp_member_declaration_test.go:322`).
**Mutation** — gate the redirect off (`found && false`), test fails: `run() calls [method:Add@ledger.hpp:4], want exactly [function:Add@ledger.cpp:2]`. Restored, passes.

---

### `01M17HC6KV4Q` — `parser.go:4333`, member templates lose their `template<…>` head — **VALID**

Reproduced directly. Two classes differing only in their template head:

```
template<class T>          -> signature "T Get()"  fingerprint 7fb6ac11  bodyhash 3483b34f
template<class T, class U> -> signature "T Get()"  fingerprint 7fb6ac11  bodyhash 3483b34f
```

Identical on every derived value. An edit confined to the head produced no entity change at all, and two overloads separated only by their constraints shared one fingerprint — which rename matching reads as one symbol. The source range also started at the declaration line, one line past the `template` keyword.

**Fix** `parser.go:4403` `cPlusPlusDeclarationSpan`, used at `parser.go:4347` for signature, body hash and start/end line. Now:

```
template<class T>          -> "template<class T> T Get()"          fp fe3465d9  lines 3-4
template<class T, class U> -> "template<class T, class U> T Get()" fp 1835ae94  lines 3-4
```

**Test** `TestCPlusPlusMemberTemplateCarriesItsTemplateHead` (`:237`).
**Mutation** — make the span function always return the inner node; test fails on all four assertions (signature, fingerprint, body hash, start line). Restored, passes.

---

### `01M17HC6KA70` — `parser.go:4292`, function-pointer-returning declarations — **VALID in part; the stated impact is REFUTED**

The rejection is real. "Leaving the method entirely absent from the graph" is not — I could not reproduce it. On `7ea287a9`, `int (*Factory())(double);` in a class body yields:

```
field  Ledger.Factory   sig="Factory int"  lines 3-3
```

It is in the graph, with the right name and the right lines, as a **field**. The finding's reason for the absence is also wrong: the field extractor does not stop at the nested function declarator — `cFamilyMemberDeclaratorName` (`parser.go:4448`) falls back to `firstDescendantOfType(node, "field_identifier")`, which jumps straight past it and reaches `Factory`.

So the defect is a misclassification, not an omission. It is worth fixing, and the tree makes the two shapes exactly separable:

```
int (*Factory())(double);  function_declarator > parenthesized_declarator > pointer_declarator > function_declarator > field_identifier
int (*handler_)(int);      function_declarator > parenthesized_declarator > pointer_declarator > field_identifier
```

A nested `function_declarator` inside the parens is the method; a bare identifier is a function-pointer data member.

**Fix** `parser.go:4270` (descend to the nested function declarator) and `parser.go:4448` (stop the field pass naming the same construct, so it does not enter the graph twice). Now `method:Ledger.Factory` sig `int (*Factory())(double)`; `field:Ledger.handler_` and `field:Ledger.table_[4]` unchanged.

**Test** `TestCPlusPlusFunctionPointerReturningMemberIsAMethod` (`:283`) — pins both directions, including that no `field:Ledger.Factory` remains.
**Mutation** — drop the nested-declarator loop; test fails: `missing method:Ledger.Factory`. Restored, passes.

---

### `01M17HC6KJ0D` — `parser.go:4246`, in-class `#if`/`#ifdef` members are omitted — **REFUTED**

This one cannot happen: tree-sitter never sees a preprocessor directive in a C or C++ file. `maskCUnsupportedSyntax` (`parser.go:1405`) blanks every `#`-prefixed line — plus its continuations and any block comment it opens — before the parse, so `preproc_if` and `preproc_ifdef` nodes do not exist in the tree this code walks, and the declaration's parent is the `field_declaration_list` directly.

The exact fixture from the finding, on unmodified `7ea287a9`:

```cpp
class Ledger {
public:
    int Add(int amount) const;
#ifdef ENABLE_RESET
    void Reset();
#endif
#if FOO
    void Guarded();
#else
    void Other();
#endif
};
```

```
method Ledger.Add      lines 3-3
method Ledger.Reset    lines 5-5     <- the #ifdef member the finding says is omitted
method Ledger.Guarded  lines 8-8
```

`Other()` is absent, but for an unrelated and pre-existing reason: the same mask emulates the preprocessor and blanks the not-taken branch (`cPreprocessorSkipping`, `parser.go:1436`). That is language-wide C/C++ behaviour on `main` too, not the direct-parent guard this finding names. No change made.

---

### Verification

- `go build ./...`, `go vet ./internal/sem/`, `gofmt -s -l internal/` — clean.
- `go test ./internal/sem/ -run 'TestCPlusPlus|TestCpp|TestC…Field|TestCStruct|TestCMacro|TestCFamily'` — pass.
- `go test ./internal/sem/ -run 'CallResolution|Receiver|Overload|Bodyless|Inherit|Ambient|Impact|Callers'` — pass. The redirect is gated on `Language == "C++"` and `bodyless`, so TypeScript overload signatures and ambient declarations are untouched.
- **Golden delta: none.** `TestProviderGoldenSnapshots` and `TestProviderGoldenFixtureQualityCoverageReport` pass WITHOUT `-update`, and the commit touches no file under `testdata/`. The fixture corpus contains no member template, no function-pointer-returning member and no header/`.cpp` receiver pair, so recall on it is unchanged — the three fixes are additive on shapes the corpus does not exercise.

Every fix was compared against `origin/main` by indexing the same fixture on both trees; that comparison is what surfaced the `JY5B` regression, which the branch's own green test run did not.

suhaanthayyil:
On finding `01M1J004JZ1N` (bodyless operator overloads dropped) — **confirmed real, but leaving it to you rather than forcing a fix**, since the code comment already scopes it out deliberately.

Reproduced:

```cpp
class Ledger {
  Ledger& operator=(const Ledger&);   // no symbol emitted
  int Add(int);                       // Ledger.Add      ✓
  void Plain() {}                     // Ledger.Plain    ✓
};
```

Confirmed the cause is exactly what the comment says. `maskCPlusPlusOperatorCall` rewrites before tree-sitter sees the text:

```
in   "  Ledger& operator=(const Ledger&);"
out  "  Ledger& op(       const Ledger&);"
```

Two things that may help whoever picks it up:

**Recovery is mechanical, because the masking is same-length.** `sameLengthReplacement` pads to the match width, so every node byte offset in the masked text indexes the same position in the original. `ParseWithStatus(path, content)` still holds the unmasked `content`; it just isn't threaded down to `cFamilyMemberDeclaratorName`. Slicing the node's range from the original would give the true spelling with no re-parse.

**The footprint is small but not local**: 5 `walkEntitiesScoped` sites, 3 `cFamilyMemberDeclaratorName`, 2 `cPlusPlusMemberDeclarationEntities`.

The alternative — narrowing the pattern so it stops matching declarations — is the riskier one: the mask exists for parser robustness, and a regression there is silent.

One thing the finding gets wrong: it is **not** the cause of the separate `operator int()` → `S.operato` truncation I filed as `01M1HX87F7MJ`. That shape is never matched by `cPlusPlusOperatorCallPattern` (it needs an operator symbol), and I verified the mask leaves it byte-identical. Different bug, still open.

suhaanthayyil:
Following up on my last comment — I said I was leaving `01M1J004JZ1N` (bodyless operator overloads dropped) to you. **I withdraw that; it is now fixed** in this branch.

The reason I deferred it was wrong. I assumed recovering the real operator name meant either threading unmasked source through five-plus signatures or weakening a parser-robustness mask. It needs neither — the existing `maskCPlusPlusConversionOperatorDecl` already solves exactly this shape, and I had just used the same technique elsewhere without noticing it applied here.

**The actual defect.** `maskCPlusPlusOperatorCall` preserved the width of the whole *match* but not of the operator’s own *name*:

```
in   "  Ledger& operator=(const Ledger&);"
out  "  Ledger& op(       const Ledger&);"
```

Symbol names are sliced from the **unmasked** source at the parsed node’s byte range, so a two-character stand-in where the name is nine characters reads the wrong bytes back — and every masked operator collapsed onto the same `op`. That collision is why the code chose to emit nothing at all rather than risk two symbols sharing one name on a class that overloads more than one operator.

**The fix** pads the stand-in with `_` to the width of the operator name, so it stays a single identifier token and the byte range slices back the real spelling:

```
out  "  Ledger& op_______(const Ledger&);"
```

Result on a class that overloads three:

```
Ledger, Ledger.operator=, Ledger.operator==, Ledger.operator!=, Ledger.Add
```

Distinct names, real spellings, no plumbing, and the mask’s robustness is unchanged — it still replaces the operator token, just at the right width.

Verified the regression test fails without the fix (all three operators absent, only `Ledger` and `Ledger.Add` emitted), and the full suite is green.

PR #178, opened 2026-08-29
fix(sem): extract F# calls applied by juxtaposition

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/136

Stacked on #171 (`audit/lang-matrix`), which pins this as an audited `callsGap`. Basing on `main` would have closed the gap while leaving the entry claiming it was open — and #171's own test treats a stale entry as a failure, by design.

## The defect

F# applies a function by writing its arguments beside it. No dot, no parentheses:

```fsharp
let double (amount: int) : int =
    add { Total = 0 } amount * 2
```

That is not an exotic corner of the language; it is how F# calls a function. No scanner saw it — `fsharpDottedCallRe` and `fsharpDottedApplyRe` need a `.`, and the generic scanner needs `name(`. So a module of F# functions calling each other produced **no CALLS edge at all**, while `capabilities --json` advertises `CALLS` for F#.

## Why it was left open before

#163 scanned the forward-pipe position and said so explicitly: *"Bare juxtaposition elsewhere (`add 1 2`) needs the kind of application parser Haskell and OCaml have and is not guessed at here — an over-eager head-position heuristic would trade the precision the call graph is built on for recall."*

That reasoning is right, and it also names the way through: **Haskell and OCaml already have that parser in this package.** F#'s closest relative is OCaml — the same ML binding forms, the same application by juxtaposition — so this follows `ocamlBareCallSites` rather than inventing a heuristic.

## What makes it precise

The difficulty is real: in `List.map helper add`, all three names are written identically and only one is applied. Three conditions carry the precision together:

1. **Only known callables.** A name is considered only if the file declares it as a function or method. An unrecognized identifier in application position is left alone rather than resolved against a same-named symbol elsewhere in the repository.
2. **Binding heads apply nothing.** Between `let`/`and`/`fun`/`function`/`member`/`val`/... and the `=` or `->` that starts the body, every name is a binder or a parameter. `let add x y = x + y` defines `add`; it does not call it.
3. **Head position AND an argument following** — or application through a pipe, where the argument is on the other side.

Together those separate a real application from a trailing argument, a bare value reference, a record field, a member access, and a parameter that happens to share a callable's name.

## Verification

`TestFSharpJuxtapositionScannerStaysNarrow` covers seventeen shapes. Every rejection case names something that **is** a known callable in the file, so only position and context can reject it — the test cannot pass by accident of the name filter alone.

Five mutations, each killed by a different case:

| mutation | result |
|---|---|
| known-callable restriction dropped | `/unknown_name`, `/and-binding_head` |
| binding-head tracking dropped | `/binding_head`, `/and-binding_head` |
| head-position check dropped | `/passed_as_a_trailing_argument` |
| argument-follows check dropped | `/reference_before_an_operator`, `/record_field`, `/parameter_named_after_a_callable` |
| provider hook disabled | `TestFSharpJuxtapositionCallExtraction` |

And a sixth on the matrix itself: with the gap retired but the hook disabled, `TestLanguageMatrixCallsGapsAreExactlyAsAudited/fsharp` fails with *"F# advertises CALLS but extracted none from its fixture"*. The matrix now guards the fix in both directions.

`go test ./...`, `go vet ./...` and `gofmt -s -l` are clean, and no golden fixture moved.

## Matrix

The `fsharp` entry retires from the audited-gap list to a plain row:

```go
{dir: "fsharp", symbols: []string{"module:Ledger", "type:Ledger", "function:add", "function:double"}, relations: []string{"DEFINES", "CONTAINS", "CALLS", "PARAM_TYPE", "USES_TYPE"}},
```

## Relationship to #163

Complementary, not overlapping, and neither depends on the other. #163 reads the name after a pipe wherever it appears, including names this file does not define; this reads application position for names the file does define, including piped ones. Verified on a scratch merge that #163 + this produces the same result as this alone on the matrix fixture, whose call is pure juxtaposition.

They do insert adjacent lines into the same `if file.Language == "F#"` block, so whichever lands second takes a one-hunk additive conflict there.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,12 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-go@v6
         with: { go-version: "1.26.x" }
-      - run: go test ./...
+      # internal/sem runs ~550s on the Windows runner against Go's 600s default
+      # per-package wall, with run-to-run swings of about 40s. That is 92% of an
+      # unset budget: the suite was one ordinary new test away from failing for a
+      # reason unrelated to the change that tripped it, which is what happened.
+      # Name a budget with headroom instead of inheriting one nobody chose.
+      - run: go test -timeout 30m ./...
 
   bm25-adapter:
     runs-on: ubuntu-latest
--- a/docs/language-support.md
+++ b/docs/language-support.md
@@ -58,6 +58,20 @@ Current generated counts:
 - Zig
 - Zsh
 
+Every language in this list has a minimal fixture under
+`internal/sem/testdata/langmatrix/` and is asserted by
+`TestLanguageExtractionMatrix` to extract real symbols and relations. A grammar
+that loads and parses without error but yields nothing is invisible in
+aggregate symbol counts, so the per-language assertion is the only thing that
+catches it. `TestLanguageMatrixCoversEverySemanticLanguage` keeps that fixture
+set in lockstep with the registry, and `TestDocumentedLanguageListsMatchTheRegistry`
+keeps both lists on this page in lockstep with the registry.
+
+Languages that advertise `CALLS` but extract none from their fixture's
+idiomatic call site are pinned in `languageMatrix` with the reason. That list is
+enforced in both directions: a language going dark fails, and so does a fix
+landing without the audited gap being retired.
+
 Protocol Buffers support covers proto3 and legacy proto2 declarations,
 including files that omit the syntax declaration, proto2 field labels, and
 groups. Compatibility parsing preserves original source locations, signatures,
--- a/docs/semantic-provider-requirements.md
+++ b/docs/semantic-provider-requirements.md
@@ -207,6 +207,13 @@ When a move has multiple equally similar destinations (within 0.05), the
 provider reports the pair as remove/add and emits a `W_MOVE_AMBIGUOUS` warning
 in the diff `warnings` array rather than guessing.
 
+A file rename whose content is byte-identical on both sides has no entity-level
+delta to reconcile, but the file path is a component of every symbol ID in it,
+so the rename still re-identifies every symbol. Such a file is reported with a
+single module-scope `moved` change carrying `old_path`/`new_path` and
+`reconciliation: MOVED`, so a pure rename is never indistinguishable from an
+empty diff.
+
 ## Relations
 
 Relations should include:
--- a/internal/cli/search.go
+++ b/internal/cli/search.go
@@ -468,11 +468,15 @@ func writeNdjsonSearch(out interface{ Write([]byte) (int, error) }, response sem
 		encoder.SetEscapeHTML(false)
 		if err := encoder.Encode(map[string]any{
 			"record_type": "search_header",
-			"query":       response.Query,
-			"repo_root":   response.RepoRoot,
-			"commit":      response.Commit,
-			"tree":        response.Tree,
-			"profile":     response.Profile,
+			// The header record carries the envelope version for the same reason
+			// the JSON payload does: a consumer reading the stream must be able to
+			// branch on the shape before it parses a single result record.
+			"format_version": response.FormatVersion,
+			"query":          response.Query,
+			"repo_root":      response.RepoRoot,
+			"commit":         response.Commit,
+			"tree":           response.Tree,
+			"profile":        response.Profile,
 		}); err != nil {
 			return err
 		}
--- a/internal/cli/search_format_version_test.go
+++ b/internal/cli/search_format_version_test.go
@@ -1,0 +1,71 @@
+package cli
+
+import (
+	"bytes"
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/entireio/entire-graph/internal/sem"
+)
+
+// Every other machine-readable surface tells a consumer which shape it is
+// looking at. Measured on origin/main against one fixture repository:
+//
+//	schema_version : diff, commit, capabilities, symbols, edges, snapshot,
+//	                 compact-ndjson
+//	format_version : impact, neighbors, def, index, stats
+//	NEITHER        : search --format json, search --format ndjson
+//
+// search is the documented entry point ("start here") and JSON is its DEFAULT
+// format, so the one surface a consumer is most likely to parse was the one it
+// could not version-check. These tests keep the discriminator on both formats.
+func TestSearchJSONCarriesFormatVersion(t *testing.T) {
+	t.Parallel()
+	var out bytes.Buffer
+	response := sem.SearchResponse{FormatVersion: sem.SearchFormatVersion, Query: "q", Results: []sem.SearchResult{}}
+	if err := writeSearchResponse(&out, response, "json", 0); err != nil {
+		t.Fatal(err)
+	}
+	var payload struct {
+		FormatVersion *int `json:"format_version"`
+	}
+	if err := json.Unmarshal(out.Bytes(), &payload); err != nil {
+		t.Fatalf("decode search json: %v\n%s", err, out.String())
+	}
+	if payload.FormatVersion == nil {
+		t.Fatalf("search json carries no format_version:\n%s", out.String())
+	}
+	if *payload.FormatVersion != sem.SearchFormatVersion {
+		t.Fatalf("format_version = %d, want %d", *payload.FormatVersion, sem.SearchFormatVersion)
+	}
+}
+
+func TestSearchNDJSONHeaderCarriesFormatVersion(t *testing.T) {
+	t.Parallel()
+	var out bytes.Buffer
+	response := sem.SearchResponse{FormatVersion: sem.SearchFormatVersion, Query: "q", Results: []sem.SearchResult{}}
+	if err := writeNdjsonSearch(&out, response); err != nil {
+		t.Fatal(err)
+	}
+	first, _, found := strings.Cut(out.String(), "\n")
+	if !found {
+		t.Fatalf("ndjson stream has no header line:\n%s", out.String())
+	}
+	var header struct {
+		RecordType    string `json:"record_type"`
+		FormatVersion *int   `json:"format_version"`
+	}
+	if err := json.Unmarshal([]byte(first), &header); err != nil {
+		t.Fatalf("decode ndjson header: %v\n%s", err, first)
+	}
+	if header.RecordType != "search_header" {
+		t.Fatalf("first record = %q, want search_header", header.RecordType)
+	}
+	if header.FormatVersion == nil {
+		t.Fatalf("ndjson search_header carries no format_version:\n%s", first)
+	}
+	if *header.FormatVersion != sem.SearchFormatVersion {
+		t.Fatalf("format_version = %d, want %d", *header.FormatVersion, sem.SearchFormatVersion)
+	}
+}
--- a/internal/sem/analyze.go
+++ b/internal/sem/analyze.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"path"
 	"path/filepath"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/entireio/entire-graph/internal/gitutil"
@@ -94,7 +96,7 @@ func AnalyzeGitRangeWithOptions(ctx context.Context, repo, base, head string, pa
 	// When both labels are byte-identical, reuse the first resolution so a ref
 	// advance between subprocesses cannot turn an intended empty diff into an
 	// old-tree/new-tree comparison. Keep the labels below for result provenance.
-	pinnedBase, pinnedHead, err := resolveDiffTrees(ctx, repo, base, head, gitutil.RevParse)
+	pinnedBase, pinnedHead, rootRelativeNames, err := resolveDiffTrees(ctx, repo, base, head, gitutil.RevParse)
 	if err != nil {
 		return Result{}, err
 	}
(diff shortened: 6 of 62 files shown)


ashtom:

**Verdict: changes required — 2 open findings, and it overlaps #163**

- Live PR head `3cbff617`; current `main` `bb41f223`; **2 ahead, 13 behind**.
- **Every green check on this PR ran against `3cbff617` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

**1. (blocking) Trail #136 carries 2 open findings against this PR.** Re-queried before posting:

```
$ entire trail finding list --trail 136 --status any --freshness any
01M15J9CDX7G  Medium  open  current  internal/sem/fsharp.go:77
  Skipping newlines here conflates separate layout-sensitive expressions...
01M15J9CDJ9J  Medium  open  current  internal/sem/fsharp.go:174
  The file-wide callable set ignores lexical shadowing. For `let add x = ...` follo...
```

Both are precision risks in the juxtaposition heuristic, which is the whole of this PR.

**2. (blocking, cross-PR) This overlaps and conflicts with #163**, which solves the narrow version
of the same problem (pipe position only):

```
$ git merge-tree --write-tree refs/prs/163 refs/prs/178   ->  CONFLICT
```

**3. (doc accuracy) `internal/sem/fsharp.go`'s header comment references an identifier this branch
does not contain.** It reasons that "`fsharpPipelineCallRe` needs a `|>`" — but:

```
$ git grep -n fsharpPipelineCallRe refs/prs/178
refs/prs/178:internal/sem/fsharp.go:7:// fsharpPipelineCallRe needs a `|>`, and the generic ...
                                     ^ the only occurrence: a comment, never a definition
$ git grep -l fsharpPipelineCallRe origin/main     # absent
$ git grep -l fsharpPipelineCallRe refs/prs/163    # internal/sem/call_scanners.go
```

The comment is written against a tree in which #163 has landed. Either stack this on #163 and make
the comment true, or fold #163's pipe case in here and drop the reference.

### Independent review state

- Entire trail **#136** exists but is still `phase=building` — the agent review has **not completed**.
- It nonetheless carries **2 open, unaddressed finding(s)** (`entire trail finding list --trail 136 --status any --freshness any`). Please work those before merge.

suhaanthayyil:
## Both trail findings reproduced and fixed; the "duplicate of #163" framing is refuted

Pushed `736b9a54`.

### Finding `01M15J9CDX7G` — `internal/sem/fsharp.go:77`, whitespace skip conflates layout-sensitive expressions — **VALID**

Reproduced against the unfixed scanner. `fsharpArgumentFollows` and `fsharpHeadPosition` both skipped `\n` like ordinary whitespace, so two expressions separated only by layout were fused. Input, callables `{add, scale, helper, double}`:

```fsharp
let report amount =
    let shown = double
    add shown 1
```

```
got map[double:{}], want [add]
```

Two errors from one cause: `double` is a bare value reference that ends its line, but the forward skip crossed the newline, found `add`, and read it as `double`'s argument — an invented CALLS edge. And `add` is a real application, but the backward skip crossed the newline, found the last name of the previous line, and read `add` as its trailing argument — a lost edge. The second case (`let shown = helper` / `add shown 1`) gives `got map[helper:{}], want [add]`.

Fix: both positions are now judged under F#'s offside rule (helpers at `internal/sem/fsharp.go:80-110`; the two call sites at `fsharpArgumentFollows` `:121-133` and `fsharpHeadPosition` `:163-175`). A token on a later line belongs to the expression only when it is indented past the line the name sits on. Genuine continuations still read as one application:

```fsharp
let report amount =
    add
        amount
        1                   -> add
let f xs =
    List.map helper
        add                 -> nothing (still a trailing argument)
```

Test: `TestFSharpJuxtapositionRespectsLayout` (5 cases) in `internal/sem/provider_fsharp_juxtaposition_test.go`.

### Finding `01M15J9CDJ9J` — `internal/sem/fsharp.go:174`, file-wide callable set ignores lexical shadowing — **VALID**

Reproduced. The callable set is file-wide, so a parameter that shares a file-level function's name resolved to that function:

```
let apply add ledger = add ledger 1     got map[add:{}], want []
let f = fun add x -> add x 1            got map[add:{}], want []
let report amount =
    let add = helper
    add amount 1                        got map[add:{}], want []   (after the layout fix)
```

In all three the applied `add` is the local binding. The edge to a module-level `add` is invented.

Fix: `fsharpLocallyBoundNames` (`internal/sem/fsharp.go:238-299`, with `fsharpDotFollows` at `:214` and `fsharpHeadTerminator` at `:303`) collects the names a block binds for itself — every parameter of every binding head, plus the name of every value binding — and `fsharpJuxtapositionCallIdentifiers` consults it at `internal/sem/fsharp.go:328` and skips them at `:351-353`.

Deliberately *not* shadowed: the name of a binding that takes arguments, because it names the function being defined. So recursion and unit-argument functions still resolve:

```
let rec fact n = fact (n - 1)   -> fact
let run () = run ()             -> run
member this.Add x = add x 1     -> add   (`this` is a qualifier, not the bound name)
let f (helper: int) = helper    -> nothing  (the parameter shadows; its type does not)
```

The set is collected over the whole block before it is consulted, so a `let` shadows for the block rather than from its own line onward. That is the conservative direction — it drops an edge rather than inventing one — and it is documented at the function.

Test: `TestFSharpJuxtapositionRespectsLocalBindings` (7 cases).

### Finding 3 — the header comment names `fsharpPipelineCallRe`, which this branch does not define — **VALID**

```
$ git grep -l fsharpPipelineCallRe origin/main                            # absent
$ git grep -n fsharpPipelineCallRe origin/audit/lang-fsharp-juxtaposition # one comment, no definition
```

Fixed by removing the clause rather than by stacking, so the comment is true on this branch and stays true after #163 lands. The same forward reference in `provider_fsharp_juxtaposition_test.go` is gone too. `git grep 'fsharpPipelineCallRe\|#163' -- internal/ docs/` is now empty on this branch.

### Finding 2 — "two different solutions to the same problem" — **REFUTED as stated; the conflict itself is real but trivial**

They are complementary, not redundant, because they are bounded differently:

- `fsharpPipelineCallIdentifiers` (#163) reads the pipe position and resolves **any** name, including one this file does not define;
- `fsharpJuxtapositionCallIdentifiers` (this PR) reads general application but only for names that are **callable bindings in this file**.

Demonstrated on a tree with both applied:

```
let run x = x |> Other.transform    #163 -> transform      this PR -> (nothing)
let run x = normalize x             #163 -> (nothing)      this PR -> normalize
```

So neither subsumes the other. Dropping #163 would lose every cross-module piped call; dropping this PR would lose ordinary F# application.

The `git merge-tree` conflict is real but is a pure adjacent-insertion in one file: both PRs append a `for name := range ...` loop directly after `fsharpDottedCallIdentifiers` in `internal/sem/provider.go`. Resolution is to keep both loops — no logic changes, no test changes.

### Verification on the merged tree

`HEAD` merged onto current `origin/main` (clean), then, on that tree:

```
go vet ./internal/sem/                                    pass
gofmt -s -l internal/sem/                                 clean
go test ./internal/sem/ -run 'TestFSharp|TestLanguage...' 122 pass
```

Combined tree (`origin/main` + this PR + #163, conflict resolved by keeping both loops): `go vet` pass, `gofmt` clean, both PRs' F# suites and the language matrix pass together.

Mutation checks — each production change reverted separately, each mutation compiles:

| mutation | result |
|---|---|
| `fsharpArgumentFollows` back to unconditional whitespace skip | `TestFSharpJuxtapositionRespectsLayout` FAIL (`got map[add double]`, `got map[add helper]`) + one shadowing case FAIL |
| `fsharpHeadPosition` back to unconditional whitespace skip | `TestFSharpJuxtapositionRespectsLayout` FAIL (`got map[]`, want `[add]`; `got map[add]`, want `[add scale]`) |
| `boundLocally[word]` guard removed | `TestFSharpJuxtapositionRespectsLocalBindings` FAIL (3 cases, `got map[add:{}]`) |
| all restored | pass |

### Sequencing

Recommended order: **`audit/lang-matrix` -> #163 -> this PR**, all three kept as separate PRs.

- This PR is already based on `audit/lang-matrix` (its testdata and `language_matrix_test.go` entry come from there), so that constraint exists independently of #163.
- #163 is based on `main`, is small, and is independent — it can land at any point.
- After #163 lands, rebase this branch and keep both loops in the `if file.Language == "F#"` block of `internal/sem/provider.go`. That is the whole resolution.

Folding them into one PR is not recommended: they are separately reviewable, separately testable, and each is a complete change on its own.

ashtom:

**Verdict: clean gate, 2 open findings; still overlaps #163**

Head moved `3cbff617` → `736b9a54`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

`736b9a54` reworks the heuristic under the offside rule, which addresses the two findings I quoted
last time. Two new ones replaced them.

**The overlap with #163 is unchanged** — both still add F# call extraction for application without
a dot, and they still conflict.

```
01M17FREES8N  Medium  open    current    internal/sem/fsharp.go:327     The shared literal stripper is line-scoped for quotes, so the contents of an F# …
01M17FREEE0F  Medium  open    current    internal/sem/provider.go:3291  Restricting candidates to `currentFileSymbols` drops ordinary juxtaposition call…
```

### Verified vs. assumed

suhaanthayyil:
## Both findings addressed — pushed `07d4392c`

---

### 1. `internal/sem/fsharp.go:327` — line-scoped literal stripper — **VALID**

`stripCodeLiteralsAndComments` ends a quoted run at the first newline (`provider.go:19659`), which is right for a language whose strings cannot cross one. F# has three forms that can: a triple-quoted string, a verbatim `@"..."` string (where `""` is an escaped quote), and a `(* ... *)` block comment, which nests.

That matters more here than for the dotted scanners, because this one reads **bare identifiers** — a worked example inside a docstring is written exactly like the call it documents.

Scanner-level, with `add` a known callable:

```
tripleQuoted  => [add]
tripleIndent  => [add]
verbatim      => [add]
blockComment  => [add]
singleLine    => []        (already handled)
```

End to end, on this branch before the fix:

```fsharp
module Doc

let add (a: int) (b: int) = a + b

let describe () =
    let usage = """
Example:
add 1 2
"""
    usage.Length
```

```
CALLS = [describe@5 -> add@3]
```

An edge invented entirely out of a documentation string.

**Fix:** `maskFSharpLiteralsAndComments` masks the three multi-line forms — and ordinary strings, so a `(*` or `//` inside one cannot open a comment that swallows the rest of the file — then hands the rest to the shared stripper. Masked bytes keep their line structure, which matters because the offside-rule checks measure columns: a mask that collapsed lines would move every following token. After: `CALLS = []`.

Scope note: the mask is applied in the juxtaposition scanner only, not to the whole F# call block. The dotted scanners require an `X.y` shape before anything can match, so their exposure is far smaller, and widening the mask would change behaviour this PR is not about.

---

### 2. `internal/sem/provider.go:3291` — `currentFileSymbols` bound — **VALID**

Restricting the callable set to this file's own definitions makes the ordinary shape of a multi-file F# project invisible. Before:

```fsharp
// src/Ledger.fs
module Ledger
let add (total: int) (amount: int) = total + amount

// src/Program.fs
module Program
open Ledger
let run (total: int) = add total 5           // line 5-6
let qualified (total: int) = Ledger.add total 5   // line 8-9
```

```
CALLS = [qualified@8 -> add@3]      <- the qualified sibling resolves
                                    <- run@5 -> add@3 is MISSING
```

**Fix:** add the callables an `open` brings into scope unqualified — `fsharpOpenedModules` + `fsharpOpenedCallableNames`, the same bound `ocamlOpenModules`/`ocamlOpenedCallableNames` already use in this package. Members of the opened module only: opening a module does not bring its neighbours in the same file into scope. Computed once per file rather than per symbol. After: `run@5 -> add@3` present.

The bound itself is kept, which is the point — a bare name with no `open` behind it still resolves to nothing.

---

### Tests

- `TestFSharpJuxtapositionIgnoresMultilineLiterals` — eight cases including a nested block comment, a verbatim string with an escaped quote, a `(*` inside a string that must **not** open a comment, and an application after a closed triple-quoted string that must still be seen.
- `TestFSharpDocstringExampleIsNotACall` — the end-to-end docstring case above.
- `TestFSharpJuxtapositionResolvesOpenedModuleCalls` — the `open Ledger` call resolves, **and** a third file calling `add` with no `open` still resolves to nothing, so the test pins the bound as well as the recall.

### Mutation

Each hunk reverted separately; each mutation compiles and was grep-confirmed applied.

| revert | result |
|---|---|
| `maskFSharpLiteralsAndComments(block)` → `stripCodeLiteralsAndComments(block)` | `--- FAIL: TestFSharpDocstringExampleIsNotACall` (`an example inside a triple-quoted docstring became a call`) and six subtest failures in `TestFSharpJuxtapositionIgnoresMultilineLiterals`; the opened-module test still passes |
| drop the opened-module names from the callable set | `--- FAIL: TestFSharpJuxtapositionResolvesOpenedModuleCalls` — `missing F# CALLS run->add through 'open Ledger': []`; the literal tests still pass |

Restored → all nine `TestFSharp*` pass.

### main vs branch vs this PR's head

A six-file F# corpus indexed on all three trees.

```
lost vs this PR's head (1):
  CALLS Doc.fs:described@5 -> Doc.fs:helper@3     <- the false docstring edge

gained vs this PR's head (5):
  Program.fs:run@5      -> Ledger.fs:add@3
  Program.fs:chained@8  -> Ledger.fs:add@3
  Program.fs:chained@8  -> Ledger.fs:scale@4
  Program.fs:indirect@13-> Ledger.fs:add@3
  Program.fs:Program@1  -> Ledger.fs:scale@4

lost vs origin/main: none
```

### Golden delta

**None.** `TestProviderGoldenSnapshots`, `TestProviderGoldenFixtureQualityCoverageReport`, `TestCapabilityMatrixCoversEmittedRelations`, `TestLanguageExtractionMatrix`, `TestLanguageMatrixCallsGapsAreExactlyAsAudited`, `TestLanguageMatrixCoversEverySemanticLanguage` and `TestDocumentedLanguageListsMatchTheRegistry` all pass unmodified; `git status` on `testdata/` is clean. The audited F# CALLS row is unchanged — this fixes which edges are emitted, not whether F# emits any.

### Also checked

LF and CRLF produce identical CALLS sets, including through the multi-line mask. `gofmt`/`go vet` clean.

### On the overlap with #163

Still real. #163 now also touches the F# block in `provider.go` (it wires the pipeline scanner into the file-level pass and keeps qualifiers through resolution). The two are independent fixes to different scanners, so whichever lands second needs that block merged rather than replaced. Happy to sequence in either order.

suhaanthayyil:
### State of the branch now that this is closed, and the exact delta it needs against current `main`

`07d4392c` is pushed to `audit/lang-fsharp-juxtaposition`, so the two fixes above are on the branch. Recording the merged-tree result so reopening is mechanical rather than exploratory.

`internal/sem/fsharp.go` is **not** on `main` — the juxtaposition scanner is unmerged. What did land is the language matrix, via #171, which is what this branch now collides with.

Merging current `main` (`7b3252aa`, 35 commits ahead) into `07d4392c`:

**One conflict** — `internal/sem/language_matrix_test.go`, add/add. Both sides added the same file; #171's copy is the one on `main`, so the resolution is to take it wholesale. The `testdata/langmatrix/**` fixtures merge without conflict.

**One test then fails**, and it is the failure `main`'s own comment on that test predicts:

```
--- FAIL: TestLanguageMatrixCallsGapsAreExactlyAsAudited/fsharp
    F# now extracts CALLS; remove its audited gap
    (F# function application by juxtaposition (`add ledger amount`) has no scanner,
     so idiomatic F# call sites are invisible) from languageMatrix
```

That is the assertion working as designed — the F# entry pins the gap as a literal with no `callsGapHolds`, and this branch closes it. The delta is one entry in `languageMatrix`:

```go
{
    dir:       "fsharp",
    symbols:   []string{"module:Ledger", "type:Ledger", "function:add", "function:double"},
    relations: []string{"DEFINES", "CONTAINS", "CALLS", "PARAM_TYPE", "USES_TYPE"},
},
```

`"CALLS"` added to `relations`; the `callsGap` field and its comment removed.

Verified, not assumed: with that resolution and that one entry changed, the merged tree is green —
`TestFSharp*`, `TestLanguageExtractionMatrix`, `TestLanguageMatrixCallsGapsAreExactlyAsAudited`, `TestLanguageMatrixCoversEverySemanticLanguage`, `TestDocumentedLanguageListsMatchTheRegistry`, `TestProviderGoldenSnapshots`, `TestCapability*` all pass, with `gofmt` and `go vet` clean. No golden fixture under `testdata/` moves.

For completeness, the other four branches in this batch (#161, #163, #170, #174) merge cleanly onto `7b3252aa` and pass their targeted tests **and** the full language-matrix set on the merged tree — #163's pipeline scanner does not fire on the F# matrix fixture, so only this branch closes that gap.

PR #179, opened 2026-08-29
fix(sem): stop an anonymous C struct being named after its first field

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/137

Stacked on #173 (`audit/lang-c-struct-references`), not on `main`. #173 adds the `cFamilyTagReference` guard to the exact two `case` arms this touches — the `struct_specifier` and `enum_specifier` arms of `entityFromNode` — and this is the second guard on the same line.

## The defect

`typedef struct {int x; int y;} Point;` is the ordinary way a C type is declared. It has **no tag**: the type's only name is the typedef's, and the specifier itself is anonymous. `nodeName` had no name node to read, so it fell through to the first identifier in pre-order — the first **field** — and emitted a struct definition called `x`:

```
type   "Point"
struct "x"        <- names nothing that exists
```

A search for the type finds a field name, a search for the field finds a struct definition, and the phantom sits at the type's own line so it competes with it.

The same shape hits three more forms:

| source | phantom emitted |
|---|---|
| `typedef enum {RED, GREEN} Color;` | `enum:RED` |
| `typedef struct {int w;} *Handle;` | `struct:w` |
| `struct Outer { struct {int nested;} anon; };` | `struct:nested` |

## The fix

Skip a C-family struct/enum specifier that has a body but no tag, on the same line as #173's tag-reference check. It has no name of its own to be emitted under, and the enclosing `type_definition` already emits the type under its real name.

## This is not just a deletion

`"type"` scopes its children, so with the phantom gone the typedef becomes the fields' nearest named container. Verified on a scratch merge with #174 (which is what makes C fields symbols at all):

| | with #174 alone | with #174 + this |
|---|---|---|
| | `type "Point"` | `type "Point"` |
| | `struct "x"` | — |
| | `field "x.x"` | `field "Point.x"` |
| | `field "x.y"` | `field "Point.y"` |

So the two changes compose: without this one, #174 would scope every anonymous-typedef struct's fields under a container named after one of those same fields.

## Verification

Three tests. `TestCNamedStructDefinitionsSurviveTheAnonymousCheck` is the counterweight — a tagged definition names itself and must still be extracted, with or without a typedef around it (`struct Plain`, `typedef struct Named {…} NamedAlias`, `enum Mode`, `typedef enum Level {…} LevelAlias`).

Two mutations:

| mutation | result |
|---|---|
| anonymous check removed | 5 member-named definitions reappear across 2 tests |
| check hardcoded to true | `…SurviveTheAnonymousCheck` loses `struct:Plain`, `struct:Named`, `enum:Mode`, `enum:Level` |

`go test ./...`, `go vet ./...` and `gofmt -s -l` are clean, and no golden fixture moved.

## Separate residue found while here

`typedef struct {int v;} A, B;` emits only `type:B` — a typedef with multiple declarators loses every name but the last. Not touched here; it is a different code path (the `type_definition` arm, not the specifier arm) and wants its own change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,12 @@ jobs:
           persist-credentials: false
       - uses: actions/setup-go@v6
         with: { go-version: "1.26.x" }
-      - run: go test ./...
+      # internal/sem runs ~550s on the Windows runner against Go's 600s default
+      # per-package wall, with run-to-run swings of about 40s. That is 92% of an
+      # unset budget: the suite was one ordinary new test away from failing for a
+      # reason unrelated to the change that tripped it, which is what happened.
+      # Name a budget with headroom instead of inheriting one nobody chose.
+      - run: go test -timeout 30m ./...
 
   bm25-adapter:
     runs-on: ubuntu-latest
--- a/docs/language-support.md
+++ b/docs/language-support.md
@@ -58,6 +58,20 @@ Current generated counts:
 - Zig
 - Zsh
 
+Every language in this list has a minimal fixture under
+`internal/sem/testdata/langmatrix/` and is asserted by
+`TestLanguageExtractionMatrix` to extract real symbols and relations. A grammar
+that loads and parses without error but yields nothing is invisible in
+aggregate symbol counts, so the per-language assertion is the only thing that
+catches it. `TestLanguageMatrixCoversEverySemanticLanguage` keeps that fixture
+set in lockstep with the registry, and `TestDocumentedLanguageListsMatchTheRegistry`
+keeps both lists on this page in lockstep with the registry.
+
+Languages that advertise `CALLS` but extract none from their fixture's
+idiomatic call site are pinned in `languageMatrix` with the reason. That list is
+enforced in both directions: a language going dark fails, and so does a fix
+landing without the audited gap being retired.
+
 Protocol Buffers support covers proto3 and legacy proto2 declarations,
 including files that omit the syntax declaration, proto2 field labels, and
 groups. Compatibility parsing preserves original source locations, signatures,
--- a/docs/semantic-provider-requirements.md
+++ b/docs/semantic-provider-requirements.md
@@ -207,6 +207,13 @@ When a move has multiple equally similar destinations (within 0.05), the
 provider reports the pair as remove/add and emits a `W_MOVE_AMBIGUOUS` warning
 in the diff `warnings` array rather than guessing.
 
+A file rename whose content is byte-identical on both sides has no entity-level
+delta to reconcile, but the file path is a component of every symbol ID in it,
+so the rename still re-identifies every symbol. Such a file is reported with a
+single module-scope `moved` change carrying `old_path`/`new_path` and
+`reconciliation: MOVED`, so a pure rename is never indistinguishable from an
+empty diff.
+
 ## Relations
 
 Relations should include:
--- a/internal/cli/search.go
+++ b/internal/cli/search.go
@@ -468,11 +468,15 @@ func writeNdjsonSearch(out interface{ Write([]byte) (int, error) }, response sem
 		encoder.SetEscapeHTML(false)
 		if err := encoder.Encode(map[string]any{
 			"record_type": "search_header",
-			"query":       response.Query,
-			"repo_root":   response.RepoRoot,
-			"commit":      response.Commit,
-			"tree":        response.Tree,
-			"profile":     response.Profile,
+			// The header record carries the envelope version for the same reason
+			// the JSON payload does: a consumer reading the stream must be able to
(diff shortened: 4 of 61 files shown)


ashtom:

**Verdict: clean**

- Live PR head `d688e472`; current `main` `bb41f223`; **2 ahead, 13 behind**.
- **Every green check on this PR ran against `d688e472` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking.

**Note (§6):** stacked on **#173** (`6f601f5f`), so #173 lands first. See #170 for the broader
C-family identity-churn point — this is one of the eight.

### Independent review state

- Entire trail **#137** exists but is still `phase=building` — the agent review has **not completed**.

suhaanthayyil:
Thanks — nothing to change here.

Confirming the stacking: this is on `6f601f5f` (#173), so #173 lands first. I have replied there confirming the one trade that review raised (opaque forward-declared types now produce no symbol) — verified, agreed with, and filed as a schema-level follow-up rather than a correction to either branch, so it does not block this one.

suhaanthayyil:
### Checked against #173 for a duplicated helper — not duplicated, and no conflict

This PR and #173 were flagged as reimplementing the same C-family specifier helper and as being on a collision course. Both halves are wrong.

**This PR does not redefine #173's helper — it calls it.** Both call sites read:

```go
if cFamilyTagReference(node, language) || cFamilyAnonymousSpecifier(node, language) {
	return Entity{}, false
}
```

and the two helpers test two different tree-sitter fields:

```go
// #173
return !validNode(node.ChildByFieldName("body"))   // no body  -> tag reference / forward decl
// this PR
return !validNode(node.ChildByFieldName("name"))   // no name  -> anonymous specifier
```

The only shared code is the five-line `switch language { case "C", "C++", "Objective-C": … }` gate, which is presumably what made them look like the same helper.

**They disagree in both directions.** Probing both predicates over every `struct_specifier`/`enum_specifier`/`union_specifier` node:

| fixture | node | `cFamilyTagReference` | `cFamilyAnonymousSpecifier` |
|---|---|---|---|
| `int f(struct Ledger *l)` | `struct Ledger` | **true** | false |
| `struct Ledger;` (forward decl) | `struct Ledger` | **true** | false |
| `typedef struct {int x; int y;} Point;` | `struct {…}` | false | **true** |
| `typedef enum {RED, GREEN} Color;` | `enum {…}` | false | **true** |
| `typedef union {int i; float f;} Value;` | `union {…}` | false | **true** |
| `struct Outer { struct {int nested;} anon; };` | inner `struct {…}` | false | **true** |
| `typedef struct Named {int a;} Alias;` | `struct Named {…}` | false | false |

Worth noting the union row: this PR's check covers `union_specifier` as well, since anonymous unions reach the same `struct_*` case arm — `typedef union {int i; float f;} Value;` used to emit a `struct:i`.

Neither predicate subsumes the other, so there is no merged helper to write. Deduping in each direction on the merged tree, both mutations compiling:

- keep only `cFamilyTagReference` → this PR's `TestCAnonymousTypedefStructIsNotNamedAfterItsFirstField` and `TestCPlusPlusAnonymousSpecifiersAreNotNamedAfterMembers` FAIL (`emitted "struct:x", a definition named after a member`).
- keep only `cFamilyAnonymousSpecifier` → #173's `TestCHeaderPrototypesEmitNoPhantomStructs` and `TestCPlusPlusTagReferencesAreNotDefinitions` FAIL (`header with no struct body emitted a struct definition at line 1 … 3 … 4`; `struct Point has 3 records, want exactly 1`).
- restore the `||` → all six pass.

**Merged-tree verification, not `merge-tree`.** Merged `audit/lang-c-anonymous-typedef` (which carries `6f601f5f` with it) into `main` in a scratch worktree: clean `ort` merge, no conflicts. On that tree `go build ./...` is clean, the 157 C/struct/typedef/enum/Objective-C tests in `internal/sem` pass, and `TestProviderGoldenSnapshots` / `TestProviderGoldenFixtureQualityCoverageReport` / `TestCapabilityMatrixCoversEmittedRelations` pass with no golden churn.

Both defects reproduce on `main` today, confirming the stack is doing two distinct jobs:
- `typedef struct {int x; int y;} Point;` → `[type:Point, struct:x]` (this PR removes `struct:x`)
- `struct Ledger;` + a prototype → `[struct:Ledger@1, struct:Ledger@2]` (#173 removes both)

No change requested. Land #173 first, then this.

One unrelated pre-existing gap seen while probing, present on `main` and unaffected by either PR: `typedef struct {int a;} A, *PA;` emits only `type:PA` — a multi-declarator typedef yields just its last declarator, so `type:A` is dropped.

PR #181, opened 2026-08-29
fix(sem): declare the relations Dart, Erlang, OCaml and SQL emit

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/139

## Impact

`capabilities --json` is the contract AGENTS.md tells agents to feature-detect against ("before assuming a language is semantically parsed"), and `.entire/graph-agent.md` repeats it as hard rule 7. It reports **Dart, Erlang, OCaml and SQL as call-and-containment only**, while all four reach the generic signature-type and data-flow passes and emit nine relation families between them:

| language | emitted | declared before |
|---|---|---|
| Dart | ASYNC_CALLS, DATA_FLOWS, PARAM_TYPE, RETURNS_TYPE, USES_TYPE | CALLS, CONSTRUCTS, CONTAINS, DEFINES |
| Erlang | DATA_FLOWS | CALLS, CONSTRUCTS, CONTAINS, DEFINES |
| OCaml | PARAM_TYPE, USES_TYPE | CALLS, CONSTRUCTS, CONTAINS, DEFINES, IMPORTS |
| SQL | DATA_FLOWS | CALLS, CONSTRUCTS, CONTAINS, DEFINES |

An agent that follows the report skips relation families the provider does produce. Dart is the widest gap found so far — five families.

## Repro (on `main`, live CLI)

```
$ entire graph edges --repo . --format ndjson --worktree   # emitted
  Dart     ['CALLS', 'DATA_FLOWS']
  Erlang   ['CALLS', 'DATA_FLOWS']
  OCaml    ['CALLS', 'PARAM_TYPE', 'USES_TYPE']
  SQL      ['CALLS', 'DATA_FLOWS']

$ entire graph capabilities --json                          # declared
  Dart     ['CALLS', 'CONSTRUCTS', 'CONTAINS', 'DEFINES']
  Erlang   ['CALLS', 'CONSTRUCTS', 'CONTAINS', 'DEFINES']
  OCaml    ['CALLS', 'CONSTRUCTS', 'CONTAINS', 'DEFINES', 'IMPORTS']
  SQL      ['CALLS', 'CONSTRUCTS', 'CONTAINS', 'DEFINES']
```

## Root cause

`ooRelationSupport` is a hand-maintained map, but nothing consults it before emitting these families. `DATA_FLOWS` is gated by `spec.emits("DATA_FLOWS")` — the **profile**, not the language — and the flow extractor (`returnFlowCalls`) is regex-driven over a callable's body, so it fires wherever a parameter is forwarded into a call. That includes languages with no `return` keyword at all (Erlang `total(A, B) -> add(A, B).`) and SQL function bodies. The signature-type pass is the same shape: it reads any language whose parameters carry annotations, so Dart `int a` and OCaml `(a : int)` both produce `PARAM_TYPE`/`USES_TYPE`.

So the map is a *report* that can drift from behaviour, not a *gate* that constrains it. (It is used as a gate only for `CONFIGURES` and `RESOURCE_DEPENDS_ON`, which these languages do not emit.)

`TestCapabilityMatrixCoversEmittedRelations` is the guard for exactly this divergence, but it is driven by `goldenFixtures`, and none of them carries an annotated signature or a forwarded argument in these four languages, so nothing ever tripped it. This is the same defect the `ooRelationSupport` comment already records for C, C++, Groovy and Julia, and the same one #168 reports for ten more — found here by probing the passes directly instead of waiting for a fixture to happen to contain the construct.

## Fix

Add the four entries, each listing only what is observed. **Pure addition: 15 inserted lines, zero deleted** — no existing map line is touched and no extraction behaviour changes.

Add `TestCapabilityMatrixDeclaresDartErlangOCamlAndSQLRelations` as a companion guard that does not depend on golden-fixture coverage: it builds one annotated, argument-forwarding file per language and asserts the invariant directly. It also asserts each language produced symbols **and** at least one non-structural relation, so the check cannot pass by extracting nothing or by the fixture silently ceasing to reach the passes it guards.

## Mutation verification

| mutation | result |
|---|---|
| all four entries removed | compiles; FAIL with all 8 observable violations |
| `Dart` drops `ASYNC_CALLS` | FAIL — `Dart emits ASYNC_CALLS but capabilities declares neither it per-language nor as a global heuristic` |
| `OCaml` drops `PARAM_TYPE` | FAIL — `OCaml emits PARAM_TYPE ...` |
| `SQL` drops `DATA_FLOWS` | FAIL — `SQL emits DATA_FLOWS ...` |
| `Erlang` drops `DATA_FLOWS` | FAIL — `Erlang emits DATA_FLOWS ...` |
| restored | PASS |

Verified after the fix with the built binary: for all four languages, emitted is now a subset of declared.

## Local gate

CI may not start (the org's Actions billing limit was hit), so this was gated locally:

- `go test ./internal/sem -count=1` — 2801 passed
- `go test ./internal/cli -count=1` — 748 passed
- `go vet ./...` clean, `gofmt -s -l` clean, `go build ./cmd/entire-graph` OK
- No golden fixture moved.

## Overlap with #168

#168 adds ten entries for a disjoint set of languages with **one exception: `Erlang`**. It declares `Erlang: {USES_TYPE, PARAM_TYPE}`; this PR declares `Erlang: {DATA_FLOWS}`. Merging both produces a **duplicate map key**, which Go rejects at compile time rather than silently dropping one, so the conflict cannot land unnoticed — whichever merges second should resolve to the union `{USES_TYPE, PARAM_TYPE, DATA_FLOWS}`. The other three languages here (Dart, OCaml, SQL) appear in neither #168 nor any other open PR.

### DIFF
--- a/internal/sem/capability_relation_gaps_test.go
+++ b/internal/sem/capability_relation_gaps_test.go
@@ -1,0 +1,160 @@
+package sem
+
+import (
+	"sort"
+	"strings"
+	"testing"
+)
+
+// TestCapabilityMatrixDeclaresDartErlangOCamlAndSQLRelations is a companion to
+// TestCapabilityMatrixCoversEmittedRelations that does not depend on a golden
+// fixture happening to contain the right construct.
+//
+// That guard walks goldenFixtures, so it only ever sees the relation families
+// those fixtures exercise. None of them carried an annotated signature or a
+// forwarded argument in Dart, Erlang, OCaml or SQL, so four languages emitted
+// nine relation families that `capabilities --json` said they could not
+// produce, and nothing tripped. AGENTS.md tells agents to feature-detect
+// against that report before trusting a language, so an under-declaration
+// silently steers them off relation families the provider does emit.
+//
+// This test builds the constructs directly instead of hoping for them: each
+// source below carries an annotated signature and forwards a parameter into a
+// call, which is what the generic type and data-flow passes key on.
+//
+// The direction of the check matches the golden guard: emitted-but-undeclared
+// is the failure. A declared-but-unseen relation is fine, because absence from
+// one small file is not evidence a language cannot produce it.
+func TestCapabilityMatrixDeclaresDartErlangOCamlAndSQLRelations(t *testing.T) {
+	sources := map[string]string{
+		// Dart: annotated params and return, an async call, and both a
+		// return-value flow and a forwarded argument.
+		"lib/flow.dart": `class Point {
+  int x = 0;
+  int y = 0;
+  Point(this.x, this.y);
+}
+
+int add(int a, int b) { return a + b; }
+
+int total(int a, int b) { return add(a, b); }
+
+Point make(int a, int b) { return Point(a, b); }
+
+int usePoint(Point p) { return add(p.x, p.y); }
+
+Future<int> later(int a) async { return await add(a, 1); }
+`,
+		// Erlang has no return keyword; the flow pass still sees the parameter
+		// forwarded into the callee's argument list.
+		"src/flow.erl": `-module(flow).
+-export([add/2, total/2]).
+
+add(A, B) -> A + B.
+
+total(A, B) -> add(A, B).
+`,
+		// OCaml annotates parameters and returns positionally (`(a : int)`),
+		// which the signature-type pass reads like any other annotation.
+		"src/flow.ml": `type point = { px : int; py : int }
+
+let add (a : int) (b : int) : int = a + b
+
+let total (a : int) (b : int) : int = add a b
+
+let make (a : int) (b : int) : point = { px = a; py = b }
+
+let point_sum (p : point) : int = add p.px p.py
+`,
+		// A SQL function body forwards its arguments into another function the
+		// same way, and the call pass already resolves it.
+		"db/flow.sql": `CREATE FUNCTION add_nums(a integer, b integer) RETURNS integer AS $$
+BEGIN
+  RETURN a + b;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE FUNCTION total_nums(a integer, b integer) RETURNS integer AS $$
+BEGIN
+  RETURN add_nums(a, b);
+END;
+$$ LANGUAGE plpgsql;
+`,
+	}
+
+	repo := t.TempDir()
+	for path, source := range sources {
+		writeFile(t, repo, path, source)
+	}
+
+	snapshot, err := BuildProviderSnapshot(t.Context(), repo, "test-version")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	capabilities := Capabilities()
+	global := map[string]bool{}
+	for _, relation := range capabilities.HeuristicRelationTypes {
+		global[relation] = true
+	}
+	declared := map[string]map[string]bool{}
+	for language, relations := range capabilities.RelationSupportByLanguage {
+		set := make(map[string]bool, len(relations))
+		for _, relation := range relations {
+			set[relation] = true
+		}
+		declared[language] = set
+	}
+
+	// Every language must have produced symbols, so the check cannot pass by
+	// extracting nothing at all — a parser regression would otherwise read as
+	// a clean run.
+	languages := []string{"Dart", "Erlang", "OCaml", "SQL"}
+	symbolCount := map[string]int{}
+	for _, symbol := range snapshot.Symbols {
+		symbolCount[symbol.Language]++
+	}
+	for _, language := range languages {
+		if symbolCount[language] == 0 {
+			t.Fatalf("%s produced no symbols; the fixture no longer exercises the passes this test guards", language)
+		}
+	}
+
+	// ...and each must have produced relations beyond the structural ones, or
+	// the fixture has stopped reaching the type and flow passes.
+	interesting := map[string]bool{}
+	violations := map[string][]string{}
+	for _, relation := range snapshot.Relations {
+		// from_id is repoKey:language:path:kind:name for symbols and
+		// repoKey:file:path for files; only the former carries a language.
+		parts := strings.Split(relation.FromID, ":")
+		if len(parts) < 3 || parts[1] == "file" {
+			continue
+		}
+		language := parts[1]
+		switch relation.Type {
+		case "CONTAINS", "DEFINES", "CALLS", "CONSTRUCTS", "IMPORTS":
+		default:
+			interesting[language] = true
+		}
+		if global[relation.Type] || declared[language][relation.Type] {
+			continue
+		}
+		if !contains(violations[language], relation.Type) {
+			violations[language] = append(violations[language], relation.Type)
+		}
+	}
+	for _, language := range languages {
+		if !interesting[language] {
+			t.Errorf("%s emitted no type or data-flow relation; the fixture no longer exercises the passes this test guards", language)
+		}
+	}
+
+	for _, language := range languages {
+		relations := violations[language]
+		sort.Strings(relations)
+		for _, relation := range relations {
+			t.Errorf("%s emits %s but capabilities declares neither it per-language nor as a global heuristic", language, relation)
+		}
+	}
+}
--- a/internal/sem/capability_relation_union_test.go
+++ b/internal/sem/capability_relation_union_test.go
@@ -1,0 +1,295 @@
+package sem
+
+import (
+	"sort"
+	"testing"
+)
+
+// capabilityRelationProbes pins one single-language source file per language
+// whose relation declaration was derived empirically rather than from the
+// golden fixtures.
+//
+// Each source deliberately carries the two constructs the generic passes key
+// on: a signature a type name can be read out of, and a parameter forwarded
+// into a call. That is what makes the expected set below reachable from this
+// one file, with no dependency on which other fixtures happen to exist.
+//
+// Single-language repositories are load-bearing, not incidental. USES_TYPE
+// resolves a signature identifier against every type symbol in the repository
+// by short name, so a polyglot fixture can hand one language an edge that
+// points at another language's type: an `-record(point, ...)` in an Erlang
+// file is enough to make an unrelated R or Clojure function look like it uses
+// a type. Probing one language at a time keeps each expectation a property of
+// that language's extractor.
+var capabilityRelationProbes = map[string]struct {
+	path   string
+	source string
+}{
+	// A type hint in the signature is Clojure's only route to USES_TYPE; the
+	// bare `[p]` parameter of an idiomatic accessor carries no type at all.
+	"Clojure": {"src/fixture/core.clj", `(ns fixture.core)
+
+(defrecord Point [x y])
+
+(defn add [a b] (+ a b))
+
+(defn make-point [x y] (Point. x y))
+
+(defn point-sum [^Point p] (add (:x p) (:y p)))
+`},
+	"Dart": {"lib/flow.dart", `class Point {
+  int x = 0;
+  int y = 0;
+  Point(this.x, this.y);
+}
+
+int add(int a, int b) { return a + b; }
+
+int total(int a, int b) { return add(a, b); }
+
+Point make(int a, int b) { return Point(a, b); }
+
+int usePoint(Point p) { return add(p.x, p.y); }
+
+Future<int> later(int a) async { return await add(a, 1); }
+`},
+	"Elixir": {"lib/point.ex", `defmodule Fixture.Point do
+  def add(a, b) do
+    a + b
+  end
+
+  def sum(x, y) do
+    add(x, y)
+  end
+end
+`},
+	// Erlang reaches the signature passes through a record pattern in the head
+	// and the flow pass through a forwarded argument, and it needs both in one
+	// file: a fixture with only one of them under-reports the language.
+	"Erlang": {"src/flow.erl", `-module(flow).
+-export([add/2, total/2, sum/1]).
+
+-record(point, {x = 0, y = 0}).
+
+add(A, B) -> A + B.
+
+total(A, B) -> add(A, B).
+
+sum(#point{x = X, y = Y}) -> add(X, Y).
+`},
+	"F#": {"src/Flow.fs", `module Fixture
+
+type Point = { X: int; Y: int }
+
+let add a b = a + b
+
+let sum (p: Point) = add p.X p.Y
+`},
+	"Haskell": {"src/Flow.hs", `module Fixture where
+
+data Point = Point
+  { px :: Int
+  , py :: Int
+  }
+
+add :: Int -> Int -> Int
+add a b = a + b
+
+pointSum :: Point -> Int
+pointSum p = add (px p) (py p)
+
+main :: IO ()
+main = print (pointSum (Point 1 2))
+`},
+	"Julia": {"src/flow.jl", `struct Point
+    x::Int
+    y::Int
+end
+
+function add(a, b)
+    return a + b
+end
+
+function pointsum(p::Point)
+    return add(p.x, p.y)
+end
+`},
+	"Lua": {"src/flow.lua", `local function add(a, b)
+  return a + b
+end
+
+local function total(x, y)
+  return add(x, y)
+end
+
+return total
+`},
+	// OCaml annotates parameters and returns positionally, which the
+	// signature-type pass reads like any other annotation.
+	"OCaml": {"src/flow.ml", `type point = { px : int; py : int }
+
+let add (a : int) (b : int) : int = a + b
+
+let total (a : int) (b : int) : int = add a b
+
+let make (a : int) (b : int) : point = { px = a; py = b }
+
+let point_sum (p : point) : int = add p.px p.py
+`},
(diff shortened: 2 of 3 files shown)


ashtom:

**Verdict: changes required — the Erlang declaration is incomplete, and it collides with #168**

- Live PR head `d17230ee`; current `main` `bb41f223`; **1 ahead, 3 behind**.
- **Every green check on this PR ran against `d17230ee` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

**1. (blocking) `Erlang` is under-declared here, and #168 declares it differently.** This PR sets

```go
"Erlang": {"DATA_FLOWS"},
```

while **#168** sets, in the same map:

```go
"Erlang": {"USES_TYPE", "PARAM_TYPE"},
```

Both PRs describe their entry as empirically derived, and neither is wrong about what it *saw* —
each only probed its own fixture. I ran both fixtures and their union through
`BuildProviderSnapshot` on current `main`:

```
PR168-fixture   languages=[Erlang]  relations=[CALLS DEFINES PARAM_TYPE USES_TYPE]
PR181-fixture   languages=[Erlang]  relations=[CALLS DATA_FLOWS DEFINES]
UNION           languages=[Erlang]  relations=[CALLS DATA_FLOWS DEFINES PARAM_TYPE SIMILAR_TO USES_TYPE]
```

**Erlang emits all three.** The correct entry is:

```go
"Erlang": {"USES_TYPE", "PARAM_TYPE", "DATA_FLOWS"},
```

Both PRs assign the whole slice for the same key, so whichever merges second silently discards the
other's relations, and neither guard test catches it because each sees only its own fixture. Your PR
description makes the right argument — probe the passes directly rather than wait for a fixture to
happen to contain the construct — and that argument applies to the *signature* passes here too, not
only the flow pass.

```
$ git merge-tree --write-tree refs/prs/168 refs/prs/181   ->  CONFLICT
```

**2. (unverified, worth checking the same way) Dart, OCaml and SQL.** I probed Erlang because two
PRs disagreed about it and that disagreement was the signal. I did **not** independently probe your
Dart / OCaml / SQL entries — please apply the same union check to them before merge, since the
failure mode (a fixture that triggers one pass but not another) is not Erlang-specific.

### Independent review state

- Entire trail **#139** exists but is still `phase=building` — the agent review has **not completed**.

suhaanthayyil:
## Finding reproduced — fixed on both branches

The Erlang collision is real. Confirmed by trial merge of the two PR heads:

```
$ git merge-tree --write-tree d17230ee 98c80088   -> exit 1 (CONFLICT)
```

Both branches assign the whole slice for `"Erlang"`, so the second to land would have
discarded the first's relations, and neither guard would have noticed: each only probes its
own fixture.

### The declarations were measured, not assumed

I ran each fixture through `BuildProviderSnapshot` on `bb41f223`, then re-ran every language
in a repository containing **only that language**. Structural relations
(`CONTAINS/DEFINES/CALLS/CONSTRUCTS/IMPORTS`) and the global heuristics are excluded.

| language | this PR (#181) | #168 | measured in isolation | agreed |
|---|---|---|---|---|
| Clojure | — | `USES_TYPE` | `USES_TYPE` (needs a `^Point` hint) | `USES_TYPE` |
| Dart | `USES_TYPE, PARAM_TYPE, RETURNS_TYPE, ASYNC_CALLS, DATA_FLOWS` | — | same | unchanged |
| Elixir | — | `DATA_FLOWS` | `DATA_FLOWS` | unchanged |
| **Erlang** | **`DATA_FLOWS`** | **`USES_TYPE, PARAM_TYPE`** | **`USES_TYPE, PARAM_TYPE, DATA_FLOWS`** | **union — both were wrong** |
| F# | — | `USES_TYPE, PARAM_TYPE` | same | unchanged |
| Haskell | — | `USES_TYPE, PARAM_TYPE` | same | unchanged |
| **Julia** | *(left at main's `DATA_FLOWS`)* | `USES_TYPE, PARAM_TYPE, DATA_FLOWS` | `USES_TYPE, PARAM_TYPE, DATA_FLOWS` | widened here too |
| Lua | — | `DATA_FLOWS` | `DATA_FLOWS` | unchanged |
| OCaml | `USES_TYPE, PARAM_TYPE` | — | same | unchanged |
| Objective-C | — | `DATA_FLOWS` | `DATA_FLOWS` | unchanged |
| Perl | — | `DATA_FLOWS` | `DATA_FLOWS` | unchanged |
| **R** | — | **`USES_TYPE, DATA_FLOWS`** | **`DATA_FLOWS`** | **narrowed — see below** |
| SQL | `DATA_FLOWS` | — | `DATA_FLOWS` | unchanged |

**Your Dart / OCaml / SQL entries hold.** You asked for the same union check on them — all
three are exactly what their language emits in isolation, and none of them is touched by
#168, so there was nothing to union.

### The union check turned up a second problem

`UNION` is not the right operator here, because emission is not monotone in the fixture.
Adding files *removes* relations:

```
Clojure alone            extra=[]                  # with #168's own fixture file
Clojure in #168 fixture  extra=[USES_TYPE]
R alone                  extra=[DATA_FLOWS]
R in #168 fixture        extra=[DATA_FLOWS USES_TYPE]
R + Erlang only          extra=[DATA_FLOWS USES_TYPE]   <- reproduces it with two files
```

`usesTypeRelations` (`internal/sem/provider.go:9097`) resolves each identifier in a
signature against every type symbol in the repository **by short name**. Dumping the edges
shows what those two extra `USES_TYPE`s actually were:

```
USES_TYPE: R/make_point     -> Erlang/point (struct) resolution=name_only scope=module
USES_TYPE: Clojure/point-sum -> Erlang/point (struct) resolution=name_only scope=module
```

They point at the `-record(point, {x = 0, y = 0}).` in #168's *Erlang* fixture file. Neither
is an R or Clojure type edge. In the 14-file union both disappear again, because a second
`point` (OCaml's) makes the short name ambiguous — which is why a "run the union" check would
have quietly given a third answer.

Clojure does reach `USES_TYPE` for real, via a `^Point` hint, so its declaration stands. R
defines no type symbols at all (`setClass` produces none), so every `USES_TYPE` seen from R
had resolved to a foreign type — R is declared `DATA_FLOWS` only.

### What changed on this branch (`b0f15e4c`)

- `internal/sem/provider.go:135-154` — the whole `ooRelationSupport` block is now
  **byte-identical to #168's**, carrying the agreed value for all thirteen languages.
  Erlang is the union; Julia is widened here too. Both PRs assigning the same text means
  merge order can no longer drop anything.
- `internal/sem/capability_relation_union_test.go` — new, also byte-identical on both
  branches.
- Your `TestCapabilityMatrixDeclaresDartErlangOCamlAndSQLRelations` is unchanged and still
  passes.

### The new guard

`TestCapabilityRelationDeclarationsMatchIsolatedExtraction` asserts **equality** between
`ooRelationSupport[lang]` and what that language emits over a single-language fixture, rather
than the one-directional emitted-implies-declared check the existing guards use. Equality is
what makes divergence impossible: a narrowed entry fails on the relation the fixture still
emits, a widened one fails on the relation it cannot produce. It probes one language per
repository deliberately, so a foreign type symbol can never justify an entry again.

### Verification

```
$ git merge-tree --write-tree d17230ee 98c80088          -> exit 1  (before)
$ git merge-tree --write-tree b0f15e4c a69440cd          -> exit 0  (after)
$ git merge-tree --write-tree origin/main b0f15e4c       -> exit 0
main + #181 + #168 merged, go vet + gofmt -s -l          -> clean
  TestCapabilityMatrixDeclaresDartErlangOCamlAndSQLRelations  PASS
  TestCapabilityMatrixDeclaresTypeAndFlowRelations            PASS
  TestCapabilityMatrixCoversEmittedRelations                  PASS
  TestCapabilityRelationDeclarationsMatchIsolatedExtraction   PASS (13 subtests)
```

Mutation (each compiles, then the new test is run alone):

| mutation | result |
|---|---|
| `Erlang` back to `{DATA_FLOWS}` (this PR's old value) | **FAIL** — `Erlang declares [DATA_FLOWS] but emits [DATA_FLOWS PARAM_TYPE USES_TYPE]` |
| `Erlang` back to `{USES_TYPE, PARAM_TYPE}` (#168's old value) | **FAIL** — `declares [PARAM_TYPE USES_TYPE] but emits [DATA_FLOWS PARAM_TYPE USES_TYPE]` |
| `Julia` back to main's `{DATA_FLOWS}` | **FAIL** — `declares [DATA_FLOWS] but emits [DATA_FLOWS PARAM_TYPE USES_TYPE]` |
| `Dart` minus `ASYNC_CALLS` | **FAIL** |
| restored | **PASS** |

### Still open

- **`name_only` cross-language type resolution is a real defect, not just a fixture smell.**
  An `-record(point, ...)` in one file produces `USES_TYPE` edges from unrelated R and
  Clojure functions in another language. Narrowing R's declaration stops it being advertised,
  but the false edges are still emitted. Worth its own change; not touched here.
- #168 still conflicts with #165 in `provider.go` (`git merge-tree 978f29ad a69440cd` -> exit
  1). That conflict predates this fix and is unrelated to the capability table.

Nothing merged.

suhaanthayyil:
The cross-language `name_only` resolution found while auditing this PR now has its own fix: #184.

Root cause: `resolveTypeReference` (`internal/sem/provider.go:9262`) filtered short-name candidates by kind but never by language, so an Erlang `-record(point, ...)` resolved as the type of an R function and a Clojure function that merely reused the bare name. Because the impossible declaration also counted towards ambiguity, a second same-named type anywhere in the corpus deleted the edge again — and could delete a real same-language one.

The fix is a language *compatibility* relation rather than same-language equality, so C/C++/Objective-C/Swift, the JVM languages, TypeScript/JavaScript, Clojure/ClojureScript and C#/F# keep resolving across the boundary; all 11 of those edges are pinned by test. Mutation-verified in both directions.

Left open there, and noted in that PR: the same unfiltered index leaks into CALLS, CONSTRUCTS, EXTENDS/INHERITS, DATA_FLOWS and TESTS via different resolvers.

ashtom:

**Verdict: clean and mergeable — but settle the Zig difference with #168 first**

Head moved `d17230ee` → `b0f15e4c`. I re-ran the whole gate rather than carrying the previous result forward.

| Check | Tree | Result |
|---|---|---|

### My earlier Erlang finding is fixed

`Erlang` is now `{"USES_TYPE", "PARAM_TYPE", "DATA_FLOWS"}` — the union I measured. Dart, OCaml,
SQL and Julia now agree with #168 as well.

### One difference from #168 remains

```
Zig   #181 = [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE]   <- main's current value
      #168 = [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE]
```

This PR keeps `main`'s value; #168 drops `READS_FIELD`. The two still conflict, so whichever merges
second overwrites the other.

I verified this is a reporting question rather than a behavioural one — `ooRelationSupport` gates
only `RESOURCE_DEPENDS_ON` and `CONFIGURES` through `languageSupportsRelation`, not `READS_FIELD`.
On the evidence, **#168 is probably the correct side**: a Zig fixture with struct field reads
emitted `[DEFINES PARAM_TYPE USES_TYPE]` and no `READS_FIELD`, and the golden fixture's only
`READS_FIELD` is Groovy's. Nothing currently protects this either way, since the guard test only
catches relations emitted *without* a declaration.

This PR has **no open, current findings** and a green gate; the Zig line is the one thing to settle.

### Entire trail findings

Re-queried immediately before posting (`--status open --freshness current`): **no open, current findings**.

### Verified vs. assumed

suhaanthayyil:
## The Zig difference, settled empirically — **#168 is right; `READS_FIELD` is wrong for Zig**

This merged with `main`'s value, so `main` now over-declares. Recording the evidence here because the question was left open on this PR, and because #168 — which carries the correction — is still open.

### It is reporting-only, confirmed

`languageSupportsRelation` has exactly two callers: `provider.go:1420` (`recordsByRelationSupport`, whose only call sites are `provider.go:3906` `RESOURCE_DEPENDS_ON` and `provider.go:3914` `CONFIGURES`) and `types.go:379` (`serviceBoundaryScanLanguage` → `HANDLES_GRAPHQL`). `ooRelationSupport` itself is read at `provider.go:735` for the capability report and at `provider.go:1432`. No path consults it for `READS_FIELD`.

### Zig cannot reach the emitter — source argument, not an empty fixture

`READS_FIELD` has a single emission site, `provider.go:9476` in `fieldAccessRelations`, gated at `provider.go:9472`:

```go
field, ok := fieldsByContainer[containerID][access.Field]
if !ok { continue }
```

`fieldsByContainer` is populated only from symbols of `Kind == "field"` (`provider.go:2774-2778`). Zig produces none: Zig struct members parse as `container_field`, and `fieldEntities` (`parser.go:4131-4136`) accepts only `field_declaration`, `public_field_definition`, `field_definition`, `property_signature`, `property_declaration`. `grep -rn container_field --include='*.go' internal/` returns **zero** hits outside the vendored grammars. The field index can never hold a Zig container.

### Fixtures

Seven Zig field-read shapes — field read off a signature-typed parameter, `self.field` in a method, pointer receiver read and write, nested `o.inner.v`, `@field(o, "inner")`, optional `o.maybe.?.v`, field read off a local `const` — indexed on current `main` (`7b3252aa`):

```
[zig-only] TYPES = [CONTAINS DEFINES PARAM_TYPE USES_TYPE]
```

Zero `READS_FIELD`, zero `WRITES_FIELD`, and zero `field`-kind Zig symbols in the dump.

### Where the appearance of Zig `READS_FIELD` came from

It is borrowed from another language. A Zig file with no local `Point`, beside a Go one:

```zig
// a.zig
fn readX(p: Point) f64 { return p.x; }
```
```go
// p.go
package p
type Point struct { x float64 }
```

On current `main`:

```
READS_FIELD  from-lang=Zig  readX(a.zig) -> x(p.go, lang=Go)
```

The short-name type lookup (`firstTypeLikeNamedPreferFile`, `provider.go:9459`) resolves the Zig signature onto a Go struct and attributes the edge to the Zig symbol. That is the only path by which Zig ever produces a `READS_FIELD`, and it is exactly the cross-language contamination this PR's own doctrine excludes by probing one language at a time. (It is also the class of edge #184 is about.)

### Verdict

Correct Zig value: `{DATA_FLOWS, PARAM_TYPE, RETURNS_TYPE, USES_TYPE}` — **#168's**. No change is needed on #168 for Zig; landing it corrects `main`.

### Why nothing currently catches this

`TestCapabilityMatrixCoversEmittedRelations` (`golden_test.go:560`) asserts *emitted ⇒ declared* only. It is structurally blind to over-declaration and stayed green with the wrong Zig value.

The equality guard this PR added, `TestCapabilityRelationDeclarationsMatchIsolatedExtraction`, does catch both directions — but only for languages in `capabilityRelationProbes`, and that map has 14 entries with **no Zig**, so Zig was unguarded. Verified by mutation on both trees: adding `READS_FIELD` back produces

```
Zig declares [DATA_FLOWS PARAM_TYPE READS_FIELD RETURNS_TYPE USES_TYPE]
     but emits [DATA_FLOWS PARAM_TYPE RETURNS_TYPE USES_TYPE]
```

and dropping `RETURNS_TYPE` fails the other way, so the probe is not vacuous. Both mutations compiled and were grep-confirmed applied; both trees restore green.

**Honest negative on a fixture-free guard:** there isn't a sound one. Over-declaration is a claim about what a language *cannot* produce, and the only sound evidence is running that language's extractor over source containing the construct — a fixture. Deriving it from node-type tables is unsound (the same node-type string is reused across grammars) and deriving it from the golden fixtures is coverage-by-coincidence, which is precisely why the golden guard is one-directional. The right meta-guard is #168's `TestCapabilityRelationProbesCoverEveryCallExtractionLanguage`, which fails when a language `supportsCallExtraction` admits has no probe — but that forces a fixture per language rather than avoiding one, and it cannot be ported here: this PR has 14 probes against 32 call-extraction languages.

### Five more divergences, not just Zig

Comparing `Capabilities().RelationSupportByLanguage` across `main` and #168, Zig is **not** the only remaining difference. #168 declares strictly more in five languages, each pinned by one of its probes:

| language | #168 adds |
|---|---|
| ClojureScript | `USES_TYPE` (no entry at all on this side) |
| Groovy | `RETURNS_TYPE`, `WRITES_FIELD` |
| JavaScript | `USES_TYPE` |
| Ruby | `USES_TYPE` |
| Swift | `WRITES_FIELD` |

Same class of silent-overwrite as the Zig line. #168 carries the complete 32-language probe set plus the meta-guard, so if its `capabilityRelationProbes` and `TestCapabilityRelationProbesCoverEveryCallExtractionLanguage` are the surviving version, all six are settled at once.

### Golden delta

**None.** `TestProviderGoldenSnapshots`, `TestProviderGoldenFixtureQualityCoverageReport` and `TestCapabilityMatrixCoversEmittedRelations` pass unchanged with `testdata/` clean — goldens record emitted relations, and no golden Zig fixture ever emitted `READS_FIELD`, so removing the declaration cannot move them.

### Footnote

Zig's `WRITES_FIELD` and `ACCESSES` are unreachable for the same reason and are correctly absent on both sides — noting it so they are not "fixed in" later.

PR #141, opened 2026-08-19
fix(sem): surface repo-controlled ignore exclusions in search (MITIGATION — makes hidden files visible, does not prevent hiding)

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/101

## MITIGATION, not a fix — read this first

This makes a silent exclusion **visible**. It does not stop a hostile contributor
from hiding code, and it does not tell you whether the hidden file was the answer
to your query. What it stops is the payload claiming a completeness it does not
have. The "What this does NOT do" section below is the honest boundary.

## The problem, reproduced against `origin/main`

`.graphignore` and `.gitignore` live in the repository, so whoever can commit
decides part of what the graph sees. One committed line naming a **tracked**
source file removes it from every search answer, and nothing in the response said
so.

Fixture: `internal/auth/auth.go` holds the real bearer-token validator;
`internal/auth/auth_stub.go` is a permissive stand-in. Both tracked and committed.
Then one line is committed:

```
$ printf 'internal/auth/auth.go\n' > .graphignore && git add -A && git commit -m "chore: trim the graph corpus"
```

`origin/main`, `--format agent`:

```
Coverage: notice (1 language/2 files; 1 warning; 0 partial failures)
- warning W_WORKTREE_SNAPSHOT
1. internal/auth/auth_stub.go:2-7 ValidateTokenStub s=22.0 [focus:4]
```

The real file is gone, the decoy is ranked #1, and the coverage line reports the
survivors as the coverage. `.gitignore` on a tracked file does the same thing in
worktree mode (Git still tracks it; the graph's own reading of the file drops it).
`--head` only loads `.graphignore`, so it is exposed through that file alone —
verified, `.gitignore` does not hide anything under `--head`.

The doctrine `entire-graph` itself installs is what makes the omission stick:
`.entire/graph-agent.md` rule 1 says search first and never grep before you
search. An agent told to trust the graph has no reason to look for what is
missing.

## Failing-first test, failing at runtime on `origin/main`

`TestSearchPayloadNamesWhatRepoIgnoreRulesRemoved` uses only API that already
exists, so it **compiles unchanged against `origin/main`** and fails there:

```
$ git worktree add --detach /tmp/fov-mainbase origin/main    # cceb4de
$ cp <the test> /tmp/fov-mainbase/internal/sem/ && go test ./internal/sem/ -run TestSearchPayloadNamesWhatRepoIgnoreRulesRemoved -v
[...]

PR #183, opened 2026-08-29
fix(sem): name an Objective-C typedef by its alias, not by its source

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/141

Objective-C's grammar is a C superset, so a typedef parses to the same
`type_definition` node C uses. `entityFromNode`'s naming branch for that node was
gated to C and C++ only, so Objective-C fell through to `nodeName` — which finds
no `name` field on a typedef and descends to the first identifier in pre-order.
That identifier always belongs to the type being aliased, never to the alias.

Measured on `origin/main`, the same source in a `.c` and a `.m` file:

| source | `.c` | `.m` (before) | `.m` (after) |
|---|---|---|---|
| `typedef struct {int v;} A, B;` | `type:B` | `type:v` — a field | `type:B` |
| `typedef struct {int w;} P;` | `type:P` | `type:w` — a field | `type:P` |
| `typedef struct Node {int q;} NodeAlias;` | `type:NodeAlias` | `type:Node` — the tag | `type:NodeAlias` |
| `typedef enum {RED, GREEN} Color;` | `type:Color` | `type:RED` — an enum constant | `type:Color` |
| `typedef NSInteger MyInt;` | `type:MyInt` | `type:NSInteger` — a framework type | `type:MyInt` |

Both directions are wrong. The declared name was not a symbol at all, so a search
for it found nothing and a declaration using it had no definition to resolve
against. And the name that *was* emitted is a phantom definition sitting on the
typedef line: for `MyInt` the graph claimed this file defines `NSInteger`, a type
it only mentions — so "where is `NSInteger` defined?" answered with an unrelated
alias in the user's own repo.

C parses the identical source correctly, so the fix is parity. The language set
moves into `cFamilyTypedefLanguage` and Objective-C joins it. No other behaviour
changes: `.c` and `.cpp` output is byte-identical, and the forms the alias regex
cannot match (`typedef void (^Handler)(int);`, `typedef int (*FnPtr)(int);`) keep
their existing `nodeName` fallback, which already named them correctly.

Objective-C++ (`.mm`) is inventory-only and unaffected.

## Test

`TestObjCTypedefIsNamedByItsAliasNotItsSource` builds a provider snapshot from one
file of typedefs, run twice: once as `.m`, once as `.c` as the control. It asserts
name by name — every alias is present as a `type`, and no field, tag, enum
constant or aliased-type name is — rather than on the whole symbol set, so it
stays orthogonal to the other open work on this node.

## Verification (local — see note below)

- New test fails on `origin/main`, passes with the fix.
- Mutation, both compiling:
  - drop `"Objective-C"` from `cFamilyTypedefLanguage` → 10 assertion failures, `.m` subtest fails. Caught.
  - drop `"C", "C++"` from it → `.c` control subtest fails, `.m` passes. Caught.
- Full `go test ./internal/sem/` green (75s), `gofmt -s`, `go vet ./internal/sem/`, `go build ./cmd/entire-graph` all clean.

CI may not start on this PR — the Actions billing limit was reached on 2026-08-29.
A red or missing check should be read as "the workflow could not run", not as a
failure. The gate above was run locally.

## Relationship to the other open work on this node

Independent of, and non-overlapping with, both:

- **#182** (multi-declarator typedefs) adds a new function and a call site in
  `walkEntitiesScoped`; this changes the language gate in `entityFromNode`.
  Different hunks, no textual conflict. Note for #182: its
  `cFamilyTypedefAliasEntities` carries its own `language != "C" && language != "C++"`
  gate, so once both land Objective-C will still keep only the last declarator of
  `typedef struct {…} A, B;`. Adopting `cFamilyTypedefLanguage` there on rebase
  closes that remaining gap. This PR deliberately leaves it alone — `A` is absent
  in C on main today too, so it is #182's finding, not this one's. The test here
  asserts `B` present and `v` absent, and stays green either way.
- **#179** (anonymous specifiers) suppresses the `struct_specifier` /
  `enum_specifier` entity. It does not touch `type_definition`, so it does not
  cover this: with #179 applied the bogus `struct:v` goes away but `type:v`
  remains. This test asserts only on `type` symbols, so it is unaffected by #179
  landing.

### DIFF
--- a/internal/sem/parser.go
+++ b/internal/sem/parser.go
@@ -3384,6 +3384,19 @@ var cFamilyNonFunctionNames = map[string]bool{
 	"while":   true,
 }
 
+// cFamilyTypedefLanguage reports whether a language's grammar spells a typedef
+// as C does, so that a `type_definition` node's name must be read from its
+// declarator rather than by descending to the first identifier. Objective-C is
+// a strict C superset and shares the node, so it shares the naming rule.
+func cFamilyTypedefLanguage(language string) bool {
+	switch language {
+	case "C", "C++", "Objective-C":
+		return true
+	default:
+		return false
+	}
+}
+
 func cFamilyTypedefAliasName(node *sitter.Node, src []byte) string {
 	text := strings.TrimSpace(stripCodeLiteralsAndComments(node.Content(src)))
 	if !strings.HasPrefix(text, "typedef ") {
@@ -4852,7 +4865,19 @@ func entityFromNode(node *sitter.Node, src []byte, language, scope string) (Enti
 			// *_type_defn child; the generic name descent would instead latch onto
 			// a leading attribute ([<CustomEquality>] type SemVerInfo -> "CustomEquality").
 			name = fsharpTypeName(node, src)
-		} else if (language == "C" || language == "C++") && node.Type() == "type_definition" {
+		} else if cFamilyTypedefLanguage(language) && node.Type() == "type_definition" {
+			// Objective-C belongs here with C and C++: its grammar is a C
+			// superset and emits the same type_definition node. Left out of
+			// this branch it fell through to nodeName, which finds no `name`
+			// field on a typedef and descends to the first identifier in
+			// pre-order — the SOURCE of the alias, never the alias. So every
+			// Objective-C typedef was published under the wrong name:
+			// `typedef struct {int v;} A, B;` became `type:v` (a field),
+			// `typedef struct Node {int q;} NodeAlias;` became `type:Node`
+			// (the tag), `typedef enum {RED, GREEN} Color;` became `type:RED`
+			// (an enum constant), and `typedef NSInteger MyInt;` became
+			// `type:NSInteger` — a phantom definition of a framework type this
+			// file does not define, while `MyInt` was not a symbol at all.
 			if alias := cFamilyTypedefAliasName(node, src); alias != "" {
 				name = alias
 			} else {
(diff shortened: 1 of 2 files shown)


ashtom:

**Verdict: clean**

- Live PR head `b7a87112`; current `main` `bb41f223`; **1 ahead, 3 behind**.
- **Every green check on this PR ran against `b7a87112` with an older `main`.** CI has not tested this change against `bb41f223`. I treated the green ticks as unproven and re-ran the gate myself on the merged tree.

### Verification matrix

| Check | Tree | Result |
|---|---|---|

### Findings

Nothing blocking. The fix is well-scoped: Objective-C joins C/C++ in reading a `type_definition`'s
name from its declarator, and the comment enumerates the four concrete mis-namings it removes.

**Note (§6):** one of the eight open C-family PRs that change symbol identity or membership
(#170, #172, #173, #174, #177, #179, #182, #183). See #170.

### Independent review state

- Entire trail **#141** exists but is still `phase=building` — the agent review has **not completed**.

suhaanthayyil:
Thanks — nothing to change here, and agreed on the framing: this brings Objective-C into line with C/C++ by reading a `type_definition`'s name from its declarator rather than its source type.

Nothing in the wider C-family sweep touched this one. For the identity-churn note, see my comment on #182 — the measured golden delta across the #174/#177 fixes was zero, so the set churns less than eight PRs implies, though that is partly because the fixtures do not exercise the constructs.

PR #184, opened 2026-08-29
fix(sem): stop bare type names resolving across language boundaries

suhaanthayyil:
https://entire.io/gh/entireio/entire-graph/trails/142

## The bug

`resolveTypeReference` resolves a bare type name from a signature against the workspace-wide short-name index. That index is keyed by **name alone**, and the candidate filter checked only `typeLikeKind` — never the language. Any workspace-unique name therefore bound confidently across a language boundary.

Repro (three files, nothing else):

```erlang
% geo.erl
-module(geo).
-export([origin/0]).
-record(point, {x = 0, y = 0}).

origin() ->
    #point{x = 0, y = 0}.
```
```r
# make.R
make_point <- function(point, scale) {
  point
}
```
```clojure
;; sum.clj
(defn point-sum [point other]
  (+ point other))
```

Indexing that directory on `main` emits:

```
USES_TYPE   R/make_point   -> Erlang/point (struct)  res=name_only scope=module conf=0.75
PARAM_TYPE  R/make_point   -> Erlang/point (struct)  res=name_only scope=module conf=0.75
USES_TYPE   Clojure/point-sum -> Erlang/point (struct) res=name_only scope=module conf=0.75
```

R cannot name an Erlang record; neither can Clojure. The edges exist purely because a bare name matched across a language boundary.

The wrong answer is also **unstable**: because the impossible declaration counts towards ambiguity, adding a second `point` anywhere in the corpus makes the short name ambiguous and the invented edges silently disappear. The same mechanism deletes *real* edges — a Go `type point struct{}` used by a Go function in the same package resolves to nothing on `main` as soon as an Erlang record named `point` exists anywhere in the repo.

## Diagnosis

`internal/sem/provider.go:9262` `resolveTypeReference` — candidate collection:

```go
for _, sym := range symbolsByShortName[name] {
    if sym.ID != from.ID && sym.Name == name && typeLikeKind(sym.Kind) {
        candidates = append(candidates, sym)
    }
}
```

No language predicate anywhere in the function. `scope=module` is a label written onto the emitted record (`provider.go:9294`, `:9305`), not a constraint that was enforced. `name_only` is not "too permissive" in general — it is the correct fallback for a workspace-unique declaration; the defect is that the candidate set it draws from is not restricted to declarations the referring language could name. The same function backs `USES_TYPE` (`usesTypeRelations`, `provider.go:9065`) and `PARAM_TYPE`/`RETURNS_TYPE` (`signatureTypeRelations`, `provider.go:9122`), so one filter fixes all three.

The codebase already recognises this defect class in one place — `firstTypeLikeNamedInLanguage` (`provider.go:9535`) exists precisely because "a cross-language short-name collision would produce a confidently wrong edge rather than a missing one" — but it is used only by `extensionMemberRelations`. The generic type resolver was the remaining hole.

## The rule, and why not "same language only"

A blanket same-language rule would destroy real edges. Measured on the pre-fix build, these cross-language type edges exist today and are legitimate:

| from → to | shared how | verdict |
|---|---|---|
| C++ → C, C → C++ | C++ compiles C headers verbatim | keep |
| Objective-C ↔ C/C++ | same headers; `.h` is sniffed to C, C++ or Objective-C | keep |
| Swift → C | Clang importer / bridging headers | keep |
| Kotlin → Java, Java → Kotlin | JVM classes | keep |
| Scala → Java | JVM classes | keep |
| Groovy → Java | JVM classes | keep |
| Clojure → Java | JVM class names in type hints | keep |
| ClojureScript → Clojure | `.cljc` is read by both readers | keep |
| TypeScript → JavaScript, JavaScript → TypeScript | one type system; `.d.ts`, JSDoc | keep |
| F# → C# | CLR types | keep |
| Elixir → Erlang | Erlang records are macros to Elixir, never a bare type reference | **drop** |
| Dart → Java | platform channels are string-keyed | **drop** |
| Ruby → C | the extension is C; Ruby never names the struct | **drop** |
| Go → Python, R → Erlang, Clojure → Erlang | nothing shared | **drop** |

So the guard is a language **compatibility relation**, not equality and not a partition. `internal/sem/langcompat.go` defines it as groups a language may belong to more than one of, closed symmetrically:

```
{C, C++, Objective-C, Objective-C++, Swift}
{Java, Kotlin, Scala, Groovy, Clojure}
{Clojure, ClojureScript}
{JavaScript, TypeScript}
{C#, F#}
```

Reflexive and symmetric, deliberately **not transitive**: ClojureScript ~ Clojure and Clojure ~ Java, but ClojureScript ≁ Java. Languages named in no group are compatible only with themselves; an unattributed (empty) language anchors nothing. Vue and Svelte are not listed because they are inventory-only — they produce component/document symbols and no signature-type edges at all, verified empirically.

`Go → C` is excluded on purpose: cgo references are qualified (`C.Foo`), and the qualified path is import-authoritative already.

## The fix

`internal/sem/provider.go:9271` — one predicate on candidate collection:

```go
if sym.ID != from.ID && sym.Name == name && typeLikeKind(sym.Kind) && languagesShareTypes(from.Language, sym.Language) {
```

Filtering **candidates** rather than emitted edges is deliberate: it also removes impossible declarations from the ambiguity count, so a foreign same-name declaration no longer suppresses the real edge. The invented answer and the missing answer were the same bug, and both are fixed.

New file `internal/sem/langcompat.go` holds the table, its symmetric closure and `languagesShareTypes`.

## Tests

`internal/sem/langcompat_test.go`:

- `TestTypeReferenceSkipsLanguagesThatCannotShareDeclarations` — the repro; asserts the R and Clojure edges are gone, with a same-language Go control edge in the same snapshot so it cannot pass by emitting nothing.
- `TestTypeReferenceKeepsSharedDeclarationLanguages` — 11 cross-family edges, both directions, each observed on the pre-fix build; a regression here means the fix destroyed a real edge.
- `TestForeignDeclarationNoLongerHidesTheRealType` — the instability half: a Go type shadowed by an Erlang record now resolves.
- `TestLanguagesShareTypesRelation` — the relation itself: reflexive, symmetric, non-transitive (`ClojureScript ≁ Java`), and every language named in the table is one the parser can actually attribute.

## Mutation verification

|---|---|---|
| revert the `languagesShareTypes` guard in `resolveTypeReference` | yes | `TestTypeReferenceSkipsLanguagesThatCannotShareDeclarations` FAILS (`R/make_point->Erlang/point was emitted (resolution=name_only scope=module)`), `TestForeignDeclarationNoLongerHidesTheRealType` FAILS (`Go/Shift->Go/point missing`); the "keeps" test still passes, as it must |
| restore | yes | all four PASS |
| split C and C++ into separate compatibility groups | yes | `TestLanguagesShareTypesRelation` FAILS (`languagesShareTypes("C", "C++") = false`) and `TestTypeReferenceKeepsSharedDeclarationLanguages` FAILS (`C++/renderWidget->C/Widget was dropped`) |
| restore | yes | all four PASS |

The compatibility rule is pinned in both directions: removing the guard reintroduces the invented edge, and weakening the table drops a real one.

Regression runs (`gofmt`, `go vet ./internal/sem/`, `go build ./...` all clean): the 12 existing tests that assert on `USES_TYPE`/`PARAM_TYPE`/`RETURNS_TYPE`, the 145 language-suite tests (`Kotlin|Swift|ObjC|Cpp|TypeScript|Java|Clojure|Erlang|Groovy|Scala|FSharp|CSharp|…`), and the `Uses|Signature|Param|Return|Resolve|Relation|Profile|Provider|Snapshot|Build` set — all green.

## Not fixed here, but measured

The same unfiltered short-name index leaks into other relation families. On the pre-fix build, a fixture of unrelated files also produced:

```
CALLS       Go/Run          -> Ruby/crunch_numbers  res=package
CONSTRUCTS  Python/mk       -> C#/ZebraBox          res=name_only
DATA_FLOWS  C#/ZebraBox     -> Python/mk            res=name_only
EXTENDS     Python/SubWidget-> Java/BaseWidget      res=name_only
INHERITS    Python/SubWidget-> Java/BaseWidget      res=name_only
TESTS       Python/test_quokka -> Erlang/quokka     res=name_only
```

Those go through different resolvers (`resolveCallTargets`, `resolveTestSubject`, the extends/constructs paths) and are left untouched by this PR to keep the change reviewable and the blast radius provable. `languagesShareTypes` is the reusable predicate for them; call resolution may want a slightly different relation than type resolution (a JVM language calling a Java method is real, an R function calling an Erlang function is not), which is why it is not bolted on here without its own empirical table.

Separately: `usesTypeRelations` emits `USES_TYPE` for languages that `ooRelationSupport` does not advertise it for (R, Clojure), so the capabilities matrix and the emitted relations disagree. Also out of scope here.


suhaanthayyil:
Extended to the four relation types I flagged as still leaking when I opened this. `b4007ad0`.

## The measurement, on `origin/main`

A corpus where Go, Python and Ruby each declare their **own** `Point.process`, plus Kotlin naming a Java type. All four `CALLS` edges resolve by `type_inferred`, and two are wrong:

```
Go/run     -> Go/Point.process     correct
Kotlin/... -> Java/Shape.area      correct  (compatible languages)
Python/run -> Go/Point.process     WRONG
Ruby/run   -> Go/Point.process     WRONG
```

These are not low-confidence guesses at a missing target. **Python and Ruby each declare the method the call actually reaches** — the resolver picked Go's because the index is keyed by bare name and Go's happened to come first.

Worth stating plainly: **the first commit on this branch did not fix this.** Relation counts on `main` and on `21d60e9d` were byte-identical over that corpus. `resolveTypeReference` was one reader of the index; call, constructor, inheritance and test resolution are four more.

## After

```
Go/run     -> Go/Point.process        unchanged
Kotlin/... -> Java/Shape.area         unchanged   ← the compatible edge survives
Python/run -> Python/Point.process    REDIRECTED
Ruby/run   -> Ruby/Point.process      REDIRECTED
```

**Relation totals are unchanged** — `CALLS 4`, 42 relations, 20 symbols, before and after. Nothing was deleted. Two edges moved from a wrong target to the right one, which is the whole argument for filtering *candidates* rather than emitted edges: with the impossible declarations out of the candidate set, the real ones resolve. Filtering edges would have deleted two and left the correct ones unfound.

## Scope

Every lookup into the global short-name index is filtered. Same-file and same-package candidate lists are **not** — they cannot cross a language boundary, so a pass over them could not change a result.

Four sites are deliberately left unfiltered because no referring symbol is in scope to filter against: `resolveQualifiedType`, `goMethodSymbolByID`, `methodReturnTypes`, and the extension-member lookup (which already restricts to an *exact* language, so it is stricter than this rule, not looser). Naming them rather than leaving them for the next person to rediscover.

## Mutation verify (all compiling, all confirmed applied by `grep` after the edit)

| mutation | result |
|---|---|
| make `sharedTypeCandidates` a pass-through (undo the fix) | **FAIL** |
| same-language-**only** instead of compatibility | **FAIL** — kills the real Kotlin→Java edge |
| drop the filter from `resolveCallTargets` only | **not caught** |
| unmutated | pass |

Row two is the one that matters for the design: it proves the rule has to be *compatibility*, not equality, and that the test would catch a naive fix.

**Row three is an honest negative and I am not dressing it up.** I could not build a fixture where the `resolveCallTargets` placement changes any edge — the bare-name path appears to be constrained by same-file preference and import resolution before it reaches the global list, and I tried Lua/R same-named free functions specifically to provoke it. I kept the filter because that path reads the same unfiltered index and the guard is free, but it is **defence in depth with no test that discriminates it**, not verified behaviour. If you would rather not carry unexercised code, say so and I will drop that one line.

`go test ./internal/sem/` — **pass, 79.0s**, full package.

## Still open, same root cause

`usesTypeRelations` emits `USES_TYPE` for languages the capability matrix does not advertise. Being handled separately on #168, since that is where the capability table lives.

suhaanthayyil:
Resolving the unverified filter I flagged in my last comment. **It was redundant. Dropped, and the path it was meant to protect is now covered by a test that discriminates.** `c43c83c8`.

## Why nothing caught it

All four `resolveCallTargets` call sites already wrap their lookup:

```go
targets = resolveCallTargets(name, from, sharedTypeCandidates(from, symbolsByShortName[name]), ...)
```

The referring symbol is known at the call site, which is why the filter belongs there. The line inside the function was a second application of an idempotent function — it could not change a result. Worse, it *looked* like the guard while the call sites were doing the work, which is exactly the kind of thing I would refuse in review.

## The path is real, and my earlier fixtures were wrong in two specific ways

A bare free-function call with no receiver ends at the globally-unique gate — `"direct call expression matched globally unique symbol name"`, `name_only`, confidence 0.68 (`provider.go:1962`). It fires when **exactly one** symbol in the workspace carries the name, so a name defined only in a foreign language is globally unique *by construction* and resolves straight across.

Both earlier attempts missed it for reasons worth recording, because each looks like a reasonable fixture:

1. The Lua/R attempt defined `helper` in **both** languages. Two candidates, so the name was never globally unique and the gate never fired — no edge either way, with or without the filter.
2. A later attempt put the call and the definition in **one file**. The same-file branch returns before the global candidate list is ever consulted.

The shape that reaches it: the name exists **exactly once**, in a **foreign** language, in a **different file**.

## The discriminating mutation

Unwrapping the bare-name call site reproduces the defect precisely:

```
a bare-name call resolved across a language boundary:
  local/001:Python:py/caller2.py:function:driver2_py (Python)
    -> local/001:Ruby:rb/defs.rb:function:beta_only (Ruby),  resolution name_only
```

| mutation | result |
|---|---|
| unwrap the bare-name call site (the real guard) | **FAIL** — edge above appears |
| unmutated | pass |

`TestBareNameCallDoesNotCrossLanguages` asserts both halves on purpose: a Python→Python call **across files** must still resolve (so a filter that broke cross-file bare-name resolution entirely would fail), and the Python→Ruby call must not. Without the first half a wrecking-ball filter passes; without the second half a no-op filter passes.

`go test ./internal/sem/` — **pass, 78.8s**, full package.

Net effect of this commit: one line of dead code removed, one previously-untested guard now tested. The filter count is unchanged where it matters.

ashtom:

| Check | Tree | Result |
|---|---|---|

### First review of this PR

This one opened after my first pass, so this is its first review from me.

```
01M17JEVQRXM  Medium  open    current    internal/sem/langcompat.go:47  This group cannot distinguish shared `.cljc` declarations from JVM-only `.clj` d…
```

### Verified vs. assumed

suhaanthayyil:
**VALID and fixed** (`b941de3b`, corrected by `ea278b94`).

## Reproduced

`parser.go:63-65` maps **both** `.clj` (JVM-only) and `.cljc` (portable) to the language `"Clojure"`, and `.cljs` to `"ClojureScript"`. So a rule stated in language names cannot tell the two Clojure dialects apart. Measured before the fix:

```
CROSS USES_TYPE: ClojureScript(src/app.cljs) -> Clojure(src/jvmonly.clj)    WRONG
CROSS USES_TYPE: ClojureScript(src/app.cljs) -> Clojure(src/portable.cljc)  correct
```

The `.clj` edge is the same class as the Erlang-record case this PR exists to fix — a binding the source language cannot express.

## Fix

The refinement reads the **declaring file's extension**, and only for a cross-language pair: within one language the question does not arise, and `.clj` referring to `.clj` is exactly what should resolve. It is symmetric in the useful direction too — a ClojureScript symbol always comes from `.cljs`, so a JVM Clojure file no longer binds a ClojureScript-only declaration either.

The language relation stays a pure language relation; the dialect narrowing lives in `candidateSharesDeclarations`, so the group table does not have to lie about what a language name means.

## Mutation verify (both compile, both confirmed applied by `grep`)

| mutation | result |
|---|---|
| drop the extension refinement | **FAIL** — the JVM-only `.clj` edge returns |
| sever the Clojure/ClojureScript pair entirely | **FAIL** — the portable `.cljc` edge is lost |

The second matters: severing the pair also removes the wrong edge, so a test that only asserted its absence would bless the over-correction. Both halves are asserted.

## Two mistakes of mine, recorded because they changed the outcome

**1. My first version of the new test failed at baseline, and the fix was right — the test was wrong.** It counted every relation targeting `src/jvmonly.clj`, including that file's own `DEFINES`/`CONTAINS` edges, which point into itself and are correct. Scoped to edges *from* the `.cljs` consumer, it passes. That is the third time in this campaign a guard test of mine was defective while the production change was sound, and mutation is what separated them each time.

**2. I pushed `b941de3b` with the suite red.** `TestTypeReferenceKeepsSharedDeclarationLanguages` was failing and I did not catch it, because my command chained `git commit` after a `grep` whose exit status masked the test's. The test itself turned out to assert the defect — it used `tick.clj` and required ClojureScript to bind it — so `ea278b94` moves that fixture to `tick.cljc`, the only Clojure source both readers accept, leaving the assertion unchanged. Corrected within minutes, but the branch was briefly red and that was avoidable.

`go test ./internal/sem/` — **pass, 82.4s**, verified on the exit code before pushing this time.

PR #211, opened 2026-09-04
perf(sem): make snapshot, search and analyze 2-3x faster and stop allocating 7.6 GiB

pjbgf:
https://entire.io/gh/entireio/entire-graph/trails/164

Profiler-led performance work across the three command families. Every change is
output-preserving: verified as byte-identical NDJSON and `diff --json` rather
than by a passing test suite.

Measured on `entireio/cli` (1,589 tracked files, 1,376 Go), 24 cores, Go 1.26.7.

| command | before | after | |
| --- | --- | --- | --- |
| `snapshot --profile full` | 40.1s / 10,508 MiB | **14.2s / 2,914 MiB** | −65% / −72% |
| `search --profile full` (cold worktree) | 20.9s / 7,591 MiB | **9.3s / 4,294 MiB** | −55% / −43% |
| `search` (cold worktree, fast) | 9.5s / 4,433 MiB | **6.6s / 3,723 MiB** | −31% / −16% |
| `diff` — one commit, 4 files | 3.41s / 770 MiB | **1.05s / 503 MiB** | −69% / −35% |
| `diff` — 125 files | 7.25s / 1,707 MiB | **2.11s / 1,329 MiB** | −71% / −22% |
| `diff` — 612 files | 12.05s / 3,039 MiB | **5.28s / 2,694 MiB** | −56% / −11% |

`snapshot` mallocs drop from 124.6M to 39.4M and GC cycles from 340 to 196.

## What was wrong

Two things dominated, and neither was subtle once profiled.

**96% of a snapshot ran in one goroutine.** The relation phase was a plain
`for _, file := range files` loop carrying 79.9% of CPU samples, while the parse
phase beside it already used a worker pool. Whole-process concurrency was
1.17× on a 24-core machine. `analyze` had the same shape twice over: its
dependent scan was 92% of a single-commit diff's wall clock, and its parse phase
was another 4.4s on a large range.

**70% of every byte allocated was regex compilation.** 180 `regexp.MustCompile`
calls with string-literal patterns sat inside function bodies, recompiling on
every call. Two of them — `forwardedParameterProperty` and `parameterVarTypes`,
both on per-symbol paths — produced 86% of that 7.22 GiB.

The rest is repeated derivation: the same symbol body stripped sixteen times,
the same alias map built five times, the same file's string constants derived
twice, and a whole-file regex tokenization that allocated a slice per identifier
to answer a membership question.

## Commits

Each one builds, vets, and was checked for identical output before the next.

| commit | change |
| --- | --- |
| `0c089adf` | **One mount-table snapshot per operation.** A cold search read `/proc/self/mountinfo` **24,948 times** — `newSameVolumePathResolver` parsed the whole table per resolver, and the git-directory walk builds one per path. |
| `d14789f4` | **Derive each symbol body and file's constants once.** `symbolBody` carries the stripped copy; `parameterAliasMap` and `staticStringConstants` computed once instead of 5× and 2×. |
| `bea0538d` | **Hoist constant regexes to package vars.** 180 sites, declared beside the function that reads them. `TestRegexpLiteralsAreCompiledOnce` fails on any new one. |
| `3e32d6fe` | **Resolve file relations in parallel.** `runProviderFilePipeline` generalized to `runIndexedPipeline[T]` so both snapshot phases share one coordinator. |
| `c49584be` | **Scan dependent candidates in parallel.** The `analyze` hot path. |
| `cb39df81` | **Parse changed files in parallel.** Sequential prime windows, parallel within each. |
| `e2eeab73` | **Scan identifier tokens without a regex.** Allocation-free byte scan. |
| `fb61340b` | **Run tests against an empty Git config.** See below — this one is not a micro-optimisation. |

## How the parallel phases stay deterministic

Ordering is the contract, not a side effect. `runIndexedPipeline` reduces
results in exact index order regardless of which worker finishes first, so
worker timing cannot reach the emitted bytes. Each loop body became a closure
that buffers what it would have emitted; the reducer replays it in file order.

Three things needed care beyond moving the loop body:

- **Budget stops are decided by the reducer, not the worker.** A worker
  reporting "over budget" carries a flag; only the reducer sees files in order,
  so only it can stop at the index the sequential loop would have. The
  skipped-file warnings therefore name the same files.
- **Shared memos got locks.** The JS cross-file namespace cache, the Python
  module-existence cache, and the five maps the dependents read closure writes
  for its later warnings. The last group is not visible from the loop body and
  would have raced.
- **One real bug, caught by the race detector and not the compiler.** The
  analyze read step assigned to an `err` declared in the enclosing function,
  shared across every worker. Now declared locally, with an AST check confirming
  neither closure writes anything it does not declare.

## Verification

- **Byte-identical output.** `snapshot` across 4 repositories × `full`/`fast`/
  `syntax-only` × committed-tree and worktree. `diff --json` across 10
  range/repository combinations, plus `commit HEAD`.
- **Race detector clean.** A full 1,376-file snapshot, three analyze ranges, and
  the whole analyze/dependents test set under `-race`.
- **Tests.** Failure set identical to the unmodified tree. Worker-count
  determinism, emission ordering, golden fixtures, and compact round-trip pass.
- **Budget path.** Still stops and warns correctly under `--max-seconds`; it
  simply reaches more files in the same wall clock (612 vs 308 at 2s).

### Caveat on the local test environment

The machine this was developed on has no C++ compiler, so the
`go-tree-sitter/yaml` grammar cannot build. Testing used a mirror with that
grammar stubbed. Three tests fail because of the stub —
`TestLanguageExtractionMatrix/yaml`,
`TestLanguageMatrixCoversEverySemanticLanguage`, and the `services-config`
golden fixture, which carries a Kubernetes manifest. **All three fail identically
on the unmodified tree**, and substituting a real grammar for the stub clears
them. `mise run check` still needs a run on a machine with a full toolchain.

## Two pre-existing defects found while verifying, not fixed here

Both predate this branch and both are reproducible on `main`.

**`reconcileMoves` is nondeterministic.** Rename pairings among tied candidates
— identical markdown code fences, and some struct fields — follow map iteration
order. On `cli HEAD~100` the unmodified binary produces a different result on
every run, and the variance reaches `dependents_count`. This is why several
equivalence checks above are reported over a restricted set of ranges: on the
affected ones there is no stable baseline to compare against. It also
contradicts the documented contract that the same repository view and options
yield the same graph.

**Heap corruption in the cgo parsing path.** Snapshotting
`internal/sem/testdata/langmatrix` and `testdata/fixtures` combined into one
repository aborts with `corrupted size vs. prev_size` in roughly a quarter of
runs. It reproduces on `main`, without the race detector, at `-workers 1`, so it
is neither a race nor anything this branch introduced. Removing the YAML files
does not stop it and no single fixture directory triggers it alone; bisection
narrows it to a group including the vendored `groovy` and `haskell` grammars,
but a heap overflow only surfaces when the allocation layout cooperates, so
subsetting is unreliable. It also aborts the test binary partway through
`internal/sem`, which is what stops a full run from completing. Worth its own
investigation under ASan.

## Why the build change is in here

`mise run test` inherits the developer's global and system Git config, and test
fixtures create real repositories and commit to them. Anything that makes
`git commit` wait is paid once per fixture commit — `commit.gpgsign` against a
hardware key is the one that bites, since every signature wants a touch and a
refused one is retried.

Restricted to the dependents tests, the same task takes **118 seconds and fails**
without `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` set, and **0.27 seconds** with
them. It reads as a slow machine rather than a misconfiguration, which is how it
went unnoticed: CI calls `go test` directly on runners with no user config to
inherit.

## Scope

26 files, +1,561 −510. No change to the `1.x` schema, the no-egress guarantee,
or `compound-v1` symbol ID stability.

---

> [!NOTE]
> **Medium Risk**
> Touches concurrency, wall-clock budgets, and shared Git readers in core analyze/dependents paths; behavior is intended to be output-preserving but regressions would show up as ordering, skip warnings, or races rather than compile failures.
> 
> **Overview**
> This PR speeds up **snapshot**, **search**, and **analyze** by parallelizing hot loops and removing repeated work, while keeping emitted NDJSON/`diff --json` byte-identical.
> 
> **Parallelism with ordered reduction:** `analyze` now parses changed files inside each metadata prime window via `runIndexedPipeline`, and the dependents reference scan runs candidates on workers the same way. Workers return buffered deltas/warnings/hits; a reducer applies them in original index order so warnings, progress, and budget stops match the old sequential behavior (including fixing **per-file** `MaxDuration` stops mid-window). Shared maps (dependents read outcomes, file string-constant memo) are mutex-protected; analyze workers use a **local** `err` instead of a shared outer variable.
> 
> **Cheaper hot paths:** Identifier tokenization for dependents uses allocation-free `forEachIdentifierToken` instead of per-match regex slices. Per-file static string constants are derived once via `fileStringConstants`. Many in-function `regexp.MustCompile` calls move to package-level vars (plus a Kotlin supertype regex). Mount-table reads go through a per-operation `mountTableCache` (`beginMountTableScope` at provider start) so path guards don’t re-parse `/proc/self/mountinfo` for every resolver.
> 
> **Tests:** Adds coverage for mount-table caching, identifier/regex parity, and budget exhaustion inside one prime window.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7335e75a65d8ed6edecdc19d985c131d41616a22. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pjbgf:
@cursor review

pjbgf:
@cursor review

PR #230, opened 2026-09-06
feat: initialize entire-graph project structure

tiybok:
- Set up Next.js project with Tailwind CSS and TypeScript
- Define core data models for code graph entities and relationships
- Implement semantic search and neighbor lookup API routes


tiybok:
wrong pr

Issue #247, opened 2026-09-10
search: prose queries lose to code-spelled terms (abbreviation gap + substring name matching)


karthik-rameshkumar:
## Summary

A plain-language query loses to a code-spelled one for the same concept. Asked for the login path in prose, `search` returns the **logging** package and no auth code in the top 5.

```
entire graph search --repo . --profile full \
  --query "the main authentication function that logs a user in"
```

Top hits, measured against `entireio/cli`:

| Rank | Score | Symbol |
| ---- | ----- | ------ |
| 1 | 22.3 | `ManualCommitStrategy.RestoreLogsOnly` |
| 2 | 22.3 | `handleLogsOnlyRewindNonInteractive` |
| 3 | 20.5 | `newDoctorLogsCmd` |
| 4 | 20.1 | `restoreResumeSessions` |

Nothing in the head is authentication. `runLogin`, `persistLogin` and the whole `auth` package are absent.

## Why

Single-term probes isolate it:

| Query | Top score | Head quality |
| ----- | --------- | ------------ |
| `auth` | **34.6** | clean — `addInsecureHTTPAuthFlag`, `runAuthenticatedDataAPI`, `NewAuthenticatedAPIClient` |
| `login` | **31.0** | clean — `persistLogin`, `RecordLoginContext`, `NewRefreshingLoginProvider` |
| `logs` | **31.1** | logging |
| `authentication` | **16.6** | noise — `slugifyTitle`, `HTTPErrorMessage`, `newRepoVisibilityCmd` |

`authentication` scores **half** what `auth` scores on the same concept, so in the combined sentence `logs` (31) simply outbids it and drags the ranking into logging. The prose spelling of the most meaningful word in the query contributes almost nothing.

## Defect 1 — no abbreviation expansion (fix proposed)

`searchNameContainsTerm` (`internal/sem/search.go`) is a substring test with plural tolerance and nothing else:

```go
if strings.Contains(lowerName, term) { return true }
// ... strips trailing "es", then trailing "s"
```

`strings.Contains("runlogin", "authentication")` is false, so every correct answer scores **zero** name coverage on the query's key term, while `"restorelogsonly"` contains `"logs"` and scores a full hit.

[...]

PR #248, opened 2026-09-10
fix(search): make prose queries reach the code spelling

karthik-rameshkumar:
https://entire.io/gh/entireio/entire-graph/trails/181

Fixes the ranking defects in #247. A plain-language query lost to a code-spelled one for the same concept: asked for the login path in prose, `search` returned the **logging** package with no auth in the top five.

```
entire graph search --repo . --profile full \
  --query "the main authentication function that logs a user in"

  1  22.3  ManualCommitStrategy.RestoreLogsOnly
  2  22.3  handleLogsOnlyRewindNonInteractive
  3  20.5  newDoctorLogsCmd
```

After this branch, ranks 1-4 are `addInsecureHTTPAuthFlag`, `newLoginCmd`, `newAuthUseCmd`, `ensureMirrorWizardAuth`.

## The four commits

**1. `e3fec8ea` — resolve prose abbreviations to the spellings code uses.** Prose says "authentication", code says `auth`. Name matching tolerated only the plural, so the prose spelling scored a miss against every correct answer. A long-to-short alias table, applied at **retrieval** (as a query term, so preselection and BM25 see it) and at **scoring** (token-scoped against identifier tokens). Both stages are needed: scoring alone re-ranks candidates already in the pool and cannot recall one that was never retrieved.

**2. `2923d52e` — join prose compounds that code spells as one identifier.** `search` is documented to take one plain sentence, and English splits compounds code writes as a single identifier. Phrasal verbs are separable ("logs **a user** in"), so the scan looks ahead up to three tokens for a particle, gated on an object phrase. Adjacent joins require a part-of-speech cue: a determiner before the head marks it a noun.

> **Corrected by `1bd66098` / `51a1628e` / `f52f6088`.** The determiner cue alone was not sufficient, and this description originally claimed it was. It catches "write **the log** in json" but does nothing when `log` is genuinely a verb and its object sits between verb and particle: "log a message in json format" still returned `newLoginCmd` at 29.6. See the correction comment below.

**3. `36b2d8cf` — match query terms at identifier token boundaries.** `strings.Contains("runlogin", "log")` is true, so plural-stripped `logs` matched `runLogin` and `RestoreLogsOnly` identically. Tokenize first, compare whole tokens, expand plural variants on both sides.

**4. `7de5a305` — add an offline retrieval eval.** Fixture corpus, table of queries, one assertion each: the gold cluster must outrank the distractor cluster. ~0.2s per case, offline, unlike `bench/`.

**5-7. `1bd66098`, `51a1628e`, `f52f6088` (@ashtom)** — two real defects in the four commits above.

*Plural queries never reached the alias table.* The retrieval loop iterated a snapshot taken **before** morphological expansion and then did an exact table lookup, so a plural long form never matched a key. `searchNameMatchesAbbreviation` had the same exact-key bug. `configurations` scored 10.4 on noise where `configuration` scored 26.5 on the right code. Fixed by expanding word variants before the lookup and taking the snapshot after morphology, which also picks up `-ies`: repositories → repository → repo.

*The format preposition.* `searchCompoundFormatPreposition` rejects the join when `in` introduces an output format rather than completing a phrasal verb. Scoped to `login` on the adjacent path and applied to every verb on the separated path, because in the separated case the object already precedes `in`, so a trailing format describes that object for any verb ("sign the request in JSON"), while adjacent "check in a JSON file" puts its object after `in` and must still join.

## Measurements, all on `entireio/cli`

| Query | Before | After |
| ----- | ------ | ----- |
| `authentication` | 16.6, noisy head | **36.7**, all auth — above the 34.6 that `auth` itself scores |
| `auth` (control) | 34.6 | 34.6, undisturbed |
| `the main authentication function that logs a user in` | 22.3, logging | **27.4**, auth/login |
| `the user cannot log in` | — | 37.2, `newLoginCmd` |
| `write the log in json format` (control) | — | JSON/log code, no auth pollution |
| `log a message in json format` (control) | 29.6 `newLoginCmd` | JSON/log code — fixed by `f52f6088`, not by the determiner cue |
| `configurations` | 10.4, noise | **26.5** `LoadCheckpointsConfig`, `Config` |
| `repositories` | 35.4 | **41.1**, alias now firing |
| `restore only the session logs` (control) | — | `RestoreLogsOnly` rank 1 at 49.1 |

## On the eval

It was validated as a **tripwire**, not merely as a passing test: reverted against `main`'s `search.go`, the separated-phrasal-verb case fails and passes again with the fix.

It also earned itself during construction, by distinguishing re-ranking from recall. That distinction is why commit 1 applies aliases at retrieval and not only at scoring, and it is not something reading the code surfaced.

Its limits are recorded in the file and worth repeating: a four-file fixture does not reproduce real BM25 dynamics. An early draft of one case gave one answer on the fixture and a different one on `entireio/cli`, neither correct. **A failure there is worth chasing; a pass is not proof of quality on real code.** Corpus-scale eval over real repositories is still missing and is tracked in #247.

## Verification

- `go test ./internal/...` green at HEAD (exit 0, six packages).
- Validated against GraphMark's Fidelity and Tasks suites over the pinned 18-repo Go corpus: **zero metric differences** vs `main`. See the validation comment. Note those suites never invoke `search`, so that is a negative control, not evidence the ranking improved.
- The hugo#12171 name-coverage calibration holds, in its unit test and in the eval.
- Queries containing no compound and no mapped abbreviation are provably unaffected: nothing enters the term set unless a table fires.

## Not in this PR

Defect 2b (body-level BM25 / `codeTokenBonus` saturation) is deliberately left open. It is the highest-blast-radius change remaining, and three mechanisms proposed for #247 have already been contradicted by measurement after reading as plausible prose. It should wait until the corpus-scale eval question is settled.

Refs #247.

### DIFF
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -628,7 +628,12 @@ func GrepIndexMatches(ctx context.Context, repo string, patterns []string, maxPe
 // case-sensitive POSIX expressions, returning only the first match per file.
 // Callers needing separate term evidence must scan each term independently.
 func GrepIndexPatternLines(ctx context.Context, repo string, patterns []string, visit func(GrepMatch) error) error {
-	return grepPatternLines(ctx, repo, "", patterns, visit)
+	return grepPatternLines(ctx, repo, "", patterns, 1, visit)
+}
+
+// GrepIndexPatternSample retains the legacy bounded sample for ordinary terms.
+func GrepIndexPatternSample(ctx context.Context, repo string, patterns []string, visit func(GrepMatch) error) error {
+	return grepPatternLines(ctx, repo, "", patterns, 32, visit)
 }
 
 // GrepTreePatternLines streams matches from the specified immutable tree and
@@ -637,15 +642,15 @@ func GrepTreePatternLines(ctx context.Context, repo, treeish string, patterns []
 	if treeish == "" || strings.HasPrefix(treeish, "-") || strings.ContainsRune(treeish, '\x00') {
 		return fmt.Errorf("invalid git grep treeish %q", treeish)
 	}
-	return grepPatternLines(ctx, repo, treeish, patterns, visit)
+	return grepPatternLines(ctx, repo, treeish, patterns, 1, visit)
 }
 
-func grepPatternLines(ctx context.Context, repo, treeish string, patterns []string, visit func(GrepMatch) error) error {
+func grepPatternLines(ctx context.Context, repo, treeish string, patterns []string, maxPerFile int, visit func(GrepMatch) error) error {
 
 	if len(patterns) == 0 {
 		return nil
 	}
-	args := []string{"grep", "--no-recurse-submodules", "--no-line-number", "--no-column", "--no-color", "--no-full-name", "-z", "-I", "-E", "-m", "1", "-f", "-"}
+	args := []string{"grep", "--no-recurse-submodules", "--no-line-number", "--no-column", "--no-color", "--no-full-name", "-z", "-I", "-E", "-m", strconv.Itoa(maxPerFile), "-f", "-"}
 	if treeish != "" {
 		args = append(args, treeish)
 	}
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -628,12 +628,29 @@ func GrepIndexMatches(ctx context.Context, repo string, patterns []string, maxPe
 // case-sensitive POSIX expressions. There is no shared per-file match cap:
 // callers can retain compact per-term evidence without buffering every line.
 func GrepIndexPatternLines(ctx context.Context, repo string, patterns []string, visit func(GrepMatch) error) error {
+	return grepPatternLines(ctx, repo, "", patterns, visit)
+}
+
+// GrepTreePatternLines streams matches from the specified immutable tree and
+// strips Git's treeish prefix from each returned path.
+func GrepTreePatternLines(ctx context.Context, repo, treeish string, patterns []string, visit func(GrepMatch) error) error {
+	if treeish == "" {
+		return fmt.Errorf("treeish is required")
+	}
+	return grepPatternLines(ctx, repo, treeish, patterns, visit)
+}
+
+func grepPatternLines(ctx context.Context, repo, treeish string, patterns []string, visit func(GrepMatch) error) error {
+
 	if len(patterns) == 0 {
 		return nil
 	}
 	args := []string{"grep", "--no-recurse-submodules", "--no-line-number", "--no-column", "--no-color", "--no-full-name", "-z", "-I", "-E"}
 	for _, pattern := range patterns {
 		args = append(args, "-e", pattern)
+	}
+	if treeish != "" {
+		args = append(args, treeish)
 	}
 	args = append(args, "--")
 	cmd := newGitCmdWithCallerLocale(ctx, repo, args...)
@@ -647,7 +664,16 @@ func GrepIndexPatternLines(ctx context.Context, repo string, patterns []string, 
 		return err
 	}
 	reader := bufio.NewReader(stdout)
-	readErr := readGrepPatternLines(reader, visit)
+	readErr := readGrepPatternLines(reader, func(match GrepMatch) error {
+		if treeish != "" {
+			prefix := treeish + ":"
+			if !strings.HasPrefix(match.Path, prefix) {
+				return fmt.Errorf("git grep missing treeish prefix")
+			}
+			match.Path = strings.TrimPrefix(match.Path, prefix)
+		}
+		return visit(match)
+	})
 	if readErr != nil {
 		_ = cmd.Process.Kill()
 	}
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -625,8 +625,8 @@ func GrepIndexMatches(ctx context.Context, repo string, patterns []string, maxPe
 }
 
 // GrepIndexPatternLines streams whole matching lines for caller-built,
-// case-sensitive POSIX expressions. There is no shared per-file match cap:
-// callers can retain compact per-term evidence without buffering every line.
+// case-sensitive POSIX expressions, returning only the first match per file.
+// Callers needing separate term evidence must scan each term independently.
 func GrepIndexPatternLines(ctx context.Context, repo string, patterns []string, visit func(GrepMatch) error) error {
 	return grepPatternLines(ctx, repo, "", patterns, visit)
 }
@@ -645,7 +645,7 @@ func grepPatternLines(ctx context.Context, repo, treeish string, patterns []stri
 	if len(patterns) == 0 {
 		return nil
 	}
-	args := []string{"grep", "--no-recurse-submodules", "--no-line-number", "--no-column", "--no-color", "--no-full-name", "-z", "-I", "-E", "-f", "-"}
+	args := []string{"grep", "--no-recurse-submodules", "--no-line-number", "--no-column", "--no-color", "--no-full-name", "-z", "-I", "-E", "-m", "1", "-f", "-"}
 	if treeish != "" {
 		args = append(args, treeish)
 	}
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -634,8 +634,8 @@ func GrepIndexPatternLines(ctx context.Context, repo string, patterns []string, 
 // GrepTreePatternLines streams matches from the specified immutable tree and
 // strips Git's treeish prefix from each returned path.
 func GrepTreePatternLines(ctx context.Context, repo, treeish string, patterns []string, visit func(GrepMatch) error) error {
-	if treeish == "" {
-		return fmt.Errorf("treeish is required")
+	if treeish == "" || strings.HasPrefix(treeish, "-") || strings.ContainsRune(treeish, '\x00') {
+		return fmt.Errorf("invalid git grep treeish %q", treeish)
 	}
 	return grepPatternLines(ctx, repo, treeish, patterns, visit)
 }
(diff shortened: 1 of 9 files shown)


karthik-rameshkumar:
## Validated against GraphMark — no collateral regression

Ran the `entirehq/graphmark` **Fidelity** and **Tasks** suites over their pinned 18-repository Go corpus, with two binaries built from this repo: `main` (`a8ae590f`) and this branch (`7de5a305`). Same oracle, same corpus at pinned refs, same harness.

**Zero metric differences.** Not "within noise" — byte-identical across all 18 repos and every metric.

Fidelity (graph correctness vs the `go/types` compiler oracle), micro-averaged:

| metric | main | this branch |
| --- | --- | --- |
| symbols P/R/F1 | 0.9407 / 1.0000 / 0.9695 | 0.9407 / 1.0000 / 0.9695 |
| calls_all P/R/F1 | 0.8406 / 0.8334 / 0.8370 | 0.8406 / 0.8334 / 0.8370 |
| calls_exact_only P/R/F1 | 0.9899 / 0.2026 / 0.3364 | 0.9899 / 0.2026 / 0.3364 |

Tasks (oracle-derived lookups), micro-averaged:

| task | main | this branch |
| --- | --- | --- |
| definitions P/R/F1 | 0.9632 / 1.0000 / 0.9813 | 0.9632 / 1.0000 / 0.9813 |
| change_impact P/R/F1 | 0.9896 / 0.8521 / 0.9157 | 0.9896 / 0.8521 / 0.9157 |

Corpus: httprouter, mux, errors, pflag, uuid, websocket, logrus, color, chi, go-cmp, mapstructure, xxhash, multierror, zerolog, cobra, testify, jwt, ulid — each at its pinned tag.

### What this does and does not establish

**Does:** the changes are confined to search ranking. Identical rather than merely close numbers is the strong form of that claim — symbol extraction, call-graph construction, definition lookup and change-impact are provably untouched.

**Does not:** validate the ranking improvements themselves. Grepping the harnesses for the subcommands they invoke returns `symbols`, `stats` and `snapshot` only. **Neither suite ever calls `search`.** So this is a clean negative control, not evidence the ranking got better.

Evidence for the improvement remains the hand measurements in the commit messages and the fixture eval in `7de5a305`. GraphMark has four suites and free-text search ranking is the one thing none of them scores; that gap is tracked in #247, and it belongs in graphmark rather than here.

Reproduce:

```sh
go build -o /tmp/graphmark-bin/oracle ./cmd/oracle          # in graphmark
export GRAPHMARK_ORACLE=/tmp/graphmark-bin/oracle
export GRAPHMARK_ENTIRE_GRAPH=<binary under test>
python3 run.py corpus/<repo> --system entire-graph --json out.json
```

`run.py --system entire-graph` is the CBM-free path, so the head-to-head competitor binary is not needed for a main-versus-branch comparison.

karthik-rameshkumar:
## Correction: two claims in the description were wrong

@ashtom pushed `1bd66098`, `51a1628e` and `f52f6088` fixing two real defects in the first four commits. Recording the corrections visibly rather than only editing the description, since the description asserted one of these as fixed when it was not.

### 1. The determiner cue did not fix the false positive, it fixed one instance of it

The description for `2923d52e` claimed a determiner before the head was the part-of-speech cue that keeps a logging query from becoming a login query. It catches `write **the log** in json`. It does nothing when `log` is genuinely a verb and its object sits between verb and particle:

```
query: "log a message in json format"          (before f52f6088)
  1  29.6  newLoginCmd
  2  29.1  RecordLoginContext
```

The exact false positive the commit message declared fixed, still present in a form I did not test. I verified the one example I had written the guard from, then generalised from it.

`searchCompoundFormatPreposition` is the actual fix: reject the join when `in` introduces an output format (`json`, `yaml`, …, or any word followed by `format`/`encoding`) rather than completing a phrasal verb. After it, the same query returns `log` and `formatMessageSuffix`.

The scoping across `51a1628e` and `f52f6088` is the part worth reading. The guard applies to `login` only on the adjacent path, and to every verb on the separated path, because the two shapes put the object in different places: separated, the object already precedes `in`, so a trailing format describes *that object* for any verb ("sign the request in JSON"); adjacent, "check in a JSON file" puts its object after `in` and must still join. The tests pin both directions, including that `log the user in and return JSON` still recovers login.

### 2. Plural queries never reached the alias table at all

A defect in `e3fec8ea` and `aa545311` (now folded into `e3fec8ea`). The retrieval loop iterated `originalTerms`, a snapshot taken **before** the morphological expansion ran, and then did an exact `searchTermAbbreviations[term]` lookup. A plural long form therefore never matched a table key. `searchNameMatchesAbbreviation` had the identical exact-key bug, so plurals missed at both retrieval and scoring.

| query | before | after |
| --- | --- | --- |
| `configurations` | 10.4, `Configured` then noise | **26.5** `LoadCheckpointsConfig`, `Config`, `globalConfigDirPath` |
| `repositories` | 35.4 | **41.1**, same hits, alias now firing |

`configurations` was returning essentially nothing while `configuration` worked, which is the failure mode the abbreviation work existed to remove, reintroduced one inflection over. `searchAbbreviations()` expands word variants before the lookup and the snapshot moved to after morphology, which also reaches `-ies`: repositories → repository → repo.

### Why the eval missed both

The fixture eval added in `7de5a305` passed throughout. Its cases were written from the same examples that drove the code, so it could only confirm what I had already checked by hand. The new cases — `plural configuration reaches config`, `log a message in json format`, `check in a JSON file`, `sign the request in json format` — are the queries I did not think to write.

That is an argument for the corpus-scale ranking suite discussed in #247 that has nothing to do with regression gating: its value is supplying queries no one on the team would have invented. SWE-bench `problem_statement` text gives exactly that, written by people who did not know the codebase.

### Status

Full `./internal/...` green at `f52f6088`, exit 0. Eval now six cases, all passing. GraphMark Fidelity and Tasks still show zero metric differences against `main`.

karthik-rameshkumar:
## Corpus study: 83 SWE-bench instances, sealed held-out, per-mechanism attribution

The fixture eval missed two real bugs (see the @ashtom corrections above), so this branch was put through the measurement it lacked: gold-file retrieval over real SWE-bench Multilingual issue text, gold = files touched by the human fix, two query forms (issue title, first two sentences), nine languages. 53-instance tuning set (`agentic-swebench/.tune_recall.json`), then a **30-instance held-out slice hash-selected with a recorded seed before any held-out result was seen**.

### Headline numbers (gold-file recall, `main` vs this branch at `1c466155`)

| set | form | metric | main | branch |
| --- | --- | --- | --- | --- |
| tuning 53 | title | recall@5 / @10 | 52.8% / 66.0% | **54.7%** / 66.0% |
| tuning 53 | title | MRR | 0.3925 | 0.3890 |
| tuning 53 | first2 | recall@5 / @10 | 56.6% / 60.4% | 56.6% / 60.4% |
| held-out 30 | title | hit@5 / @10 | 23/30 / 24/30 | 23/30 / 24/30 |
| held-out 30 | title | MRR | 0.5575 | 0.5409 |
| held-out 30 | first2 | hit@5 / @10 | 24/30 / 24/30 | **23/30** / 24/30 |
| held-out 30 | first2 | MRR | 0.5511 | 0.5372 |

Verdict in one line: **@10 flat everywhere, @5 ±1 instance per set, MRR -0.004 to -0.015; the targeted prose-vocabulary class is dramatically better; every residual loss is diagnosed.**

### What the study forced

**1. The abbreviation table is now frequency-derived (`1c466155`).** The query "return 4xx on invalid input for **delete** method" (caddy-5870) lost its gold file because `delete → del` pulled every `Del()` helper into the head; verified by single-entry removal. The fix is a keep rule measured over 1.4M identifier tokens from 40 repos: alias ≥ 2:1 over the long form AND long form ≤ 50 per 100k. 43 of 74 hand-written pairs failed it, including `transaction→txn` (txn occurs once in 1.4M tokens) and `database→db` (code prefers the long form 3:1). No hand exceptions.

**2. The token-boundary commit's two rank-1 losses are an artifact ending, not a regression to fix.** vue-11899 and preact-2896 dropped from rank 1 because the old substring scoring counted morphological variants of one query word as separate coverage terms: "scoped" spawned three variants (`scoped, scope, scop`), all substring-matching `scopedPlugin`, giving gold 3/3 coverage from a single word while `htmlNesting` got 2/3 from "nesting". Token matching counts each word once, both files tie at 1/3, and the accidental lead evaporates. There is no principled tweak that restores those rank-1s; they were morphology arithmetic. Against that, the same commit **improves** gold rank on nine other (instance, form) pairs across the two sets (ruff 4→2, fmt-3272 2→1, jq 6→5, vue-11739 8→6 and 3→2, axios 8→6, fmt-3248 3→2, php-cs-fixer 4→3, prometheus 4→3).

**3. One held-out loss belongs to the alias mechanism itself and is accepted:** fmt-2317 (4→6, first2), where the query genuinely discusses documentation and `documentation → doc` boosts doc-adjacent files. The entry survives the frequency rule and earns its keep elsewhere.

### Also found: a measurement bug in the existing recall harness

`tools/recall_agent_query.py` runs `--format text --max-context-bytes 3000`; at `--top-k 10` that emits **4** ranked paths (json/unbounded emits 10, measured on jq-2235). Its published "recall @ top-5" was recall over whatever survived the payload byte budget. This study used json/unbounded. Will be reported to the benchmark repo separately.

### Where this leaves the merge

The gate proposed earlier was "corpus flat-or-better". The strict reading fails on MRR by 0.004-0.015, entirely explained by two accidental rank-1s ending and one alias-adjacent shuffle. The case for merging anyway: the defect this PR exists to fix (prose vocabulary never reaching code spelling) is measured and large where it applies (`authentication` 16.6 → 36.7 on entireio/cli, above the 34.6 of the code spelling itself; the headline login query fixed; plurals fixed), the SWE-bench corpus is dominated by queries already written in code vocabulary where these features correctly stay out of the way, and every negative delta now has a name and a mechanism attached.

karthik-rameshkumar:
## Re-measured on the current head (`3016b04e`): recall flat, first2 MRR positive, every delta named

The corpus study two comments up measured a superseded build; the branch has since gained nine commits of alias-mechanics hardening plus the frequency-derived table. Re-ran the 53-instance tuning set against `main` on the actual head:

| form | metric | main | head `3016b04e` |
| --- | --- | --- | --- |
| title | recall@5 / @10 | 52.8% / 66.0% | 52.8% / 66.0% |
| title | MRR | 0.3925 | 0.3853 |
| first2 | recall@5 / @10 | 56.6% / 60.4% | 56.6% / 60.4% |
| first2 | MRR | 0.4253 | **0.4260** |

Gold-rank movements, all of them: **8 improvements** (laravel 2→1, nushell 3→2, jq 6→5, axios 8→6, immutable-js 3→2, fmt-3248 3→2, vue-11739 8→6 and 3→2) vs **2 regressions**: fluentd 5→6, and vue-11899 1→5, which is the diagnosed substring-variant-multi-counting artifact ending (see the corpus-study comment). The caddy delete-method regression is gone: the frequency-derived table no longer carries `delete → del`.

### Table criterion revised after review feedback

The first derivation used a code-prefers-the-short-form ratio and wrongly pruned `database → db`; `TestSearchReviewRoundFourRegressions` caught it (`openDB` must be reachable from a "database" query, and 183 real `openDB`-style tokens exist). Final rule, measured over 1.4M identifier tokens from 40 repos: **long form ≤ 50 per 100k** (catches every entry with measured harm: delete 80, context 139, error 345, source 683) **AND alias ≥ 25 occurrences** (kills the dead aliases: txn appears once in 1.4M tokens, authz seven times, sess zero). 51 of 74 pairs survive. The derivation and the do-not-add-by-taste rule are in the table's doc comment.

### One open thread: perf

On a cold cache against `entireio/cli`, this head ran the `authentication` query in 155s where the pre-rebase build took 113s, both measured under corpus-run contention, so neither number is clean. Worth one quiet-machine timing of the boundary-aware git scan before merge; filed as a review question, not a blocker claim.

Full `./internal/...` green at `3016b04e`, including the round-four regression tests and the fixture eval.

PR #249, opened 2026-09-11
fix(sem): verify Gradle project ownership before deriving tests

### LINKED ISSUE #219: verify: narrow Gradle tier emits a project path without the settings-inclusion check
The SUITE tier (`deriveSearchVerifySuiteGradle`) now refuses to name a Gradle project path unless a
settings script literally declares it, using `searchVerifyGradleSettings` /
`searchVerifyGradleSettingsIncludes`. The NARROW tier (`deriveSearchVerifyGradle`,
`internal/sem/search_verify.go`) still emits `./gradlew :<dir>:test --tests <Class>` from the
repository root on the strength of a `build.gradle` existing in `<dir>` — nothing checks that the
root build declares `<dir>` as a project.

## Reproduction

Against `fix/verify-search-medium-findings`:

```go
files := map[string]string{
    "gradlew":                      "",
    "settings.gradle":              "include ':app'\n",
    "app/build.gradle":             "",
    "lib/build.gradle":             "",
    "lib/src/main/java/A.java":     "",
    "lib/src/test/java/ATest.java": "",
}
ev := searchVerifyTestEvidence(files)
got := deriveSearchVerifyCommand(searchVerifySubject{
    sourcePath:   "lib/src/main/java/A.java",
    testPath:     "lib/src/test/java/ATest.java",
    testEvidence: "covering test",
}, &ev)
// got.Command == "./gradlew :lib:test --tests 'ATest'"
```

The settings script declares `:app` and nothing else. Gradle answers
`Project 'lib' not found in root project`, so the VERIFY block advertises a hard-gate command that
cannot run — the same class the suite tier just closed.

## Why it is not in that PR

Deliberately deferred, twice now, for the reasons recorded on the trail:

- the failure is LOUD (Gradle names the missing project) rather than the suite tier's silent
  "ran a different build with unresolved sibling dependencies", so it is less severe;
- the narrow tier is the more heavily used path, and PR #197 already carries three behavioural
  changes to this file (`include` call-shape, the Maven reactor walk bound, Node workspace
  membership) on top of a two-deep stack.

## Fix when picked up

Reuse `searchVerifyGradleSettings` / `searchVerifyGradleSettingsIncludes` so both tiers share one
notion of what declares a project, and pin the DECLINE direction as well as the emit. Two questions
the suite tier had to answer and this one will too:

1. A directory carrying its OWN settings script is a separate build root; the suite tier spells that
   `-p <dir> test`, and the narrow tier currently has no `-p` form to put `--tests` on.
2. A settings script that COMPUTES its includes (a `listFiles()` loop) declares nothing this
   predicate can read, so the tier goes silent where it used to emit a correct command. The suite
   tier accepted that trade; whether the narrow tier should is the actual decision here.

MuskanPaliwal:
Closes #219

`entire graph search` can suggest a VERIFY command after locating relevant code. Before this change, its Gradle logic could return a command that failed outright or passed without testing the edited project.

## The bug

The narrow verifier treated a source directory as proof of Gradle project membership. Finding `lib/build.gradle` was enough to emit:

```sh
./gradlew :lib:test --tests 'ATest'
```

But the directory `lib/` does not establish a project named `:lib`. That membership comes from `settings.gradle` or `settings.gradle.kts`.

If the settings file omits `:lib`, Gradle rejects the task. If it remaps `:lib` to another directory, the command can pass while testing code other than the edited files.

Declining that narrow command was not sufficient. Suite discovery could continue upward and substitute `./gradlew test` from an ancestor build that did not own the nested project.

## What changed

The narrow and suite routes now share one Gradle target resolver. It:

- finds the nearest wrapper at or above the project;
- finds the nearest settings script between the project and that wrapper;
- accepts included projects only when a literal settings declaration proves membership;
- rejects projects whose `projectDir` is remapped elsewhere;
- recognizes a directory with its own settings file as an independent build root;
- addresses intermediate build roots through `-p`;
- stops fallback at a Gradle boundary that cannot prove a safe command.

This shared resolution supports filtered commands for nested builds. For an independent build under `modules/`, it emits:

```sh
./gradlew -p modules test --tests 'ATest'
```

For `modules/core/` declared as `:core` by `modules/settings.gradle`, it emits:

```sh
./gradlew -p modules :core:test --tests 'ATest'
```

The whole-suite route uses the same target and omits only the test-class filter.

The final fallback now says that no safe verification command was found instead of incorrectly claiming that no build manifest exists.

## Resulting behavior

| Repository evidence | Result |
|---|---|
| Root settings include `:lib` | `./gradlew :lib:test --tests 'ATest'` |
| Root settings omit `:lib` | Emit no Gradle command |
| Root settings remap `:lib` to another directory | Emit no command for files under `lib/` |
| Source belongs directly to the root project | `./gradlew :test --tests 'ATest'` |
| Nested directory has its own settings file | `./gradlew -p modules test --tests 'ATest'` |
| Intermediate settings include `:core` | `./gradlew -p modules :core:test --tests 'ATest'` |
| A nested Gradle boundary cannot prove ownership | Do not substitute an ancestor suite |

These checks fail closed. A settings script that computes its includes dynamically cannot be proven by the literal scanner, so it produces no command. That is the intended safety boundary: silence is preferable to a passing command that may test unrelated code.

## Verification

```sh
go test ./internal/sem -run '^TestSearchVerifyGradleNarrowNestedBuilds$' -count=1
go test ./internal/sem -run '^TestSearchVerifySuiteGradleNamesTheProjectUnderAnAncestorWrapper$'
go test ./internal/sem -run 'Gradle' -count=1
mise run check
```

All checks passed, including formatting, vet, build, status-line tests, and the full race-enabled Go suite.

The tests derive commands from repository fixtures rather than launching a real Gradle build. Shell-safety tests execute generated commands against local runner stubs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,28 +2,47 @@
 
 # Entire Graph
 
-Coding agents lose time before the edit, while they are still looking for the
-right code. Entire Graph is a plugin for the Entire CLI that gives an agent a
-precomputed map of one Git repository: ranked code search plus definitions,
-callers, types, routes, and change impact, each with `file:line` locations. The
-built-in analyzer parses the repository locally with tree-sitter and makes no
-network requests, model calls, or API-key lookups. Installing the plugin is the
-networked step.
+Agents are continually spending their budgets before they write the first line
+of code, rummaging through files, grepping for function names, and re-reading
+the same files and configs, as they attempt to understand a codebase fresh with
+each session. According to [OpenRouter's Head of Insights](https://www.linkedin.com/posts/peterjameswalker_february-6th-2026-potentially-the-last-share-7493029881841344512-IK89/?utm_source=share&utm_medium=member_desktop&rcm=ACoAABfX0nABz6sCWbPldiV_9liETVfz5fRLAD0), agentic
+token usage grew 14x between February and August 2026, up from 0.51 trillion tokens
+to 7.3 trillion.
 
-Setup happens once per repository. After that, the interface is your coding
-agent: you ask a code question in plain language, the agent runs graph queries,
-reads the code the graph points at, and answers with citations. A captured
-example is shown below.
+Entire Graph is a plugin for the Entire CLI specifically designed to enable your
+agents to stop paying that cost. It hands your agent a precomputed map of a Git
+repository: ranked code search plus definitions, callers, types, routes, and
+change impact, each with `file:line` locations. The built-in analyzer parses the
+repository locally with tree-sitter and makes no network requests, model calls,
+or API-key lookups.
+
+When running [LoCoMo](https://github.com/snap-research/locomo) against competitors,
+we measured the top score of 94.74% for Entire Graph. We also observed token savings
+up to 71% depending on the coding scenario. As always, your mileage may vary.
+
+## Setup
+
+[Entire CLI](https://github.com/entireio/cli#quick-start) is required. Then setup happens once per repository:
+
+```sh
+entire graph init-agents --repo .
+```
+
+If the plugin is not installed yet, the Entire CLI offers to install it on the
+spot. After that, the interface is your coding agent: you ask a code question in
+plain language, the agent runs graph queries, reads the code the graph points
+at, and answers with citations. A captured example follows.
 
 ## Benchmarks
 
-The Entire Graph retrieval engine ranked first in an eight-system LoCoMo comparison
-(1,540 questions, shared reader and judge, a 200-item retrieval budget requested for
-every arm) while building its index without model calls. Measured 2026-08-14 on the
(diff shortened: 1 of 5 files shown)


MuskanPaliwal:
A follow-up review caught two more problems, fixed in `75051c0e`.

When Gradle rejected a nested child, narrow discovery kept walking and could emit a parent project's test command. That command could pass without testing the child. The fallback also said "no build manifest found" when a manifest existed but did not prove a safe command.

The narrow and suite paths now stop at a declined Gradle boundary. The fallback now says "no safe verification command found." New tests cover both cases.

The focused Gradle tests pass. The full check hit the existing timeout in `TestHeadReadersShareDeepUnsafeMetadataAllowance`; that test passed alone under `-race` in 32.351 seconds.

Issue #225, opened 2026-09-05
verify: Rakefile test-task detection ignores namespace scope (all three patterns)


suhaanthayyil:
`searchVerifyRakefileDefinesTest` decides whether a Rakefile declares a top-level `test` task, which
is what licenses `rake test`. Rake scopes a task declared inside `namespace :foo do … end` as
`foo:test`, and none of the three patterns knows about namespaces, so all three license `rake test`
for a Rakefile that answers "Don't know how to build task 'test'".

Measured against `fix/verify-search-medium-findings`, `searchVerifyRakefileDefinesTest` returns true
for every one of these:

```ruby
namespace :foo do
  task :test do            # defines foo:test
  end
end
```
```ruby
namespace :foo do
  Rake::Task.define_task(:test)   # defines foo:test
end
```
```ruby
namespace :foo do
  Rake::TestTask.new do |t|       # defines foo:test
  end
end
```

Agent review raised this against the `define_task` pattern specifically, but that pattern is the
newest of the three: `searchVerifyRakeTestTaskPattern` and the `TestTask` generator pattern both
predate it and have the same hole. The anchor is not where the class lives.

## Why it is not a small fix

Deciding scope means tracking Ruby block structure — `namespace … do` paired with its `end`, against
every other construct that also ends with `end` (`if`, `def`, `class`, `do` blocks on other tasks),
plus the brace block form and heredocs. That is a parser, not a regex tweak.

The cheap approximations are not sound. "Decline when the file contains `namespace` and the matched
line is indented" also declines a top-level task written inside a conditional, and would drop
coverage for most large Rakefiles — a broad silence bought for a narrow correctness gain, which is a
trade to make deliberately rather than inside a findings round.

The shape worth considering instead is deriving the QUALIFIED task: a declaration found inside
`namespace :foo` licenses `rake foo:test`, which is a command that runs, rather than silence.

suhaanthayyil:
Duplicate of #205. The in-code comment at `internal/sem/search_verify.go:1533` asserting #205 is "behaving exactly as it did" should be re-checked as part of the fix.

Issue #224, opened 2026-09-05
verify: workspace globs need brace expansion; the unreadable-pattern default is wrong either way


suhaanthayyil:
`searchVerifyNodeWorkspaceCovers` decides whether an ancestor `yarn.lock`'s manager may be used for a
leaf package, by matching the root manifest's `workspaces` globs. Yarn globs with **micromatch**;
the matcher here is `path.Match`, whose language is smaller. Brace expansion (`{packages,tools}/*`)
and the extglob groups are ordinary characters to it, so such a pattern is unreadable and the code
must pick a side for the leaves it cannot answer for. **Both sides are wrong for some repository,
and agent review has now asked for each of them.**

## The two cases, and why neither side is free

| pattern | leaf | permissive (current) | strict |
|---|---|---|---|
| `{packages,tools}/*` | `tools/scratch` (a member) | `yarn test` — correct | `npm test` — **PnP cannot run it** |
| `{packages,tools}/*` | `standalone` (not a member) | `yarn test` — **Yarn rejects the leaf** | `npm test` — correct |

Both wrong answers are a hard-gate command that cannot run, which is the failure this whole block
exists to prevent. There is no ordering of the two that avoids both.

## Demonstration that the strict remedy is not free

Flipping the unreadable-pattern answer from `true` to `false` compiles and fails the cases pinned for
the opposite finding:

```
search_verify_runnable_test.go:978: command = "cd tools/scratch && npm test", want "cd tools/scratch && yarn test"
--- FAIL: TestSearchVerifyNodeAncestorLockfileNeedsWorkspaceMembership/brace_expansion_is_not_answered,_so_the_manager_is_kept
--- FAIL: TestSearchVerifyNodeAncestorLockfileNeedsWorkspaceMembership/an_extglob_group_is_not_answered_either
```

## The answer that is free

Stop having an unreadable class for the common case: **expand braces before matching.**
`{packages,tools}/*` is mechanically expandable to `packages/*` and `tools/*` (nested and multiple
groups included), after which `path.Match` answers both rows of the table correctly and neither
finding has a case. That leaves only the extglob groups unreadable — genuinely rare in a
`workspaces` array — and the permissive default is a defensible answer for a residue that small,
stated as such rather than as a shrug over the whole class.

Not done inside PR #197 because it is a new matching capability with its own bug surface (nesting,
escaping, an expansion bound), decided against a benchmark of real `workspaces` fields rather than
[...]

Issue #221, opened 2026-09-05
docs-and-fixtures section swallows program-text fixtures, but the obvious remedy breaks VERIFY


suhaanthayyil:
## What is true today

`searchDocsSectionClass` (internal/sem/search_section.go:137) admits
`searchFileClassFixture`, and `classifySearchFile` assigns that class by DIRECTORY
segment (`testdata`, `fixtures`, `snapshots`, `golden`, `baselines`, ...). So a file
that is program text gets labelled into the docs-and-fixtures section:

```
classifySearchFile("internal/sem/testdata/case.go")  = fixture
searchDocsSectionPath(q, "internal/sem/testdata/case.go") = true   -> Section = docs-and-fixtures
classifySearchFile("crates/ruff_linter/resources/test/fixtures/flake8_simplify/SIM201.py") = fixture -> docs-and-fixtures
classifySearchFile("pkg/golden/report.go") = fixture -> docs-and-fixtures
```

This contradicts the function's own contract — its doc comment says the section is for
classes that "hold no program text" — and `AssessSearchConfidence` will report the
reason string `"top hit holds no program text"` for a `.go` file. Being in the section
also excludes the hit from related-site expansion (search_related.go:277) and from
type-card / container-map / literal / callee enrichment.

This repo's own fixtures are exactly the case: `internal/sem/testdata` holds 9 `.go`,
9 `.ts`, 5 `.py`, 4 `.java`, 4 `.cs`, 3 `.rs`, ... alongside 29 `.golden` recordings.

## Why the obvious remedy is not free

The natural fix — restrict the section to RECORDINGS, i.e. keep a fixture in the
section only when it matches a fixture extension, is non-program-text, or maps to no
known language — was implemented and run. It passes the whole suite
(`go test ./internal/sem/ ./internal/cli/` green), but it moves the failure into VERIFY.

`searchVerifySubjectFor` only inspects results in the PRIMARY section, and
`searchTestArtifactPath("internal/sem/testdata/case.go")` is true (the `testdata`
segment). So the moment a program-text fixture returns to primary, VERIFY adopts it as
the test to run:

```
baseline: verify subject = {sourcePath: internal/sem/parser.go, testPath: "",                            testName: "",     testEvidence: ""}
remedy:   verify subject = {sourcePath: internal/sem/parser.go, testPath: internal/sem/testdata/case.go, testName: Case,  testEvidence: "ranked test"}
```

[...]

Issue #205, opened 2026-09-04
Rake detection matches a namespaced task as if it were top-level


suhaanthayyil:
Pre-existing, independent of the `task(:test)` over-strictness fixed on #197, and split out so it does not gate that PR.

`searchVerifyRakefileDefinesTest` is line-anchored, so an indented declaration inside a namespace matches:

```ruby
namespace :foo do
  task :test do
    # ...
  end
end
```

That defines **`foo:test`**, not `test`, so a `rake test` command derived from it will not find the task.

## Why it was not fixed alongside the regex widening

The regex must keep accepting a top-level `task :test` that happens to be indented — inside `if`, `begin`, or a conditional block — so indentation alone cannot distinguish the two cases. Telling them apart requires tracking `namespace ... do ... end` blocks, which is parsing, not pattern matching. Attempting it in the regex would trade this false positive for false negatives on valid top-level tasks, which is the exact mistake the #197 round already had to correct once.

## Severity

Low. The failure is **loud** — `rake` reports `Don't know how to build task 'test'` — rather than silently running the wrong thing, which is the direction this repo treats as acceptable.

## Fix when picked up

Track namespace blocks and qualify the emitted task name (`rake foo:test`), or decline when the only match is namespaced. Pin the decline direction and keep the indented-but-top-level cases accepted — the 30-row table in `TestSearchVerifyRakefileDeclarationFormsAreRecognised` is the place to extend.

suhaanthayyil:
Same defect as #225 — namespaced Rake tasks emit an unqualified `rake test`. Tracking together in v0.5.0; one PR will close the pair.

PR #261, opened 2026-09-20
fix(verify): derive only commands the repository can actually run

### LINKED ISSUE #219: verify: narrow Gradle tier emits a project path without the settings-inclusion check

suhaanthayyil:
Four VERIFY derivations printed a command that dies on invocation. Each is reproduced below with the real `entire-graph query` verb on a real fixture repository, before and after.

Closes #219, #203, #225, #205, #224, #221.

---

### #219 / #203 — Gradle narrow tier emits an undeclared project path

`deriveSearchVerifyGradle` derived `:lib:test` from the existence of `lib/build.gradle`. A project path is not a directory path: `:lib` resolves against the projects the settings script declares. `searchVerifyGradleSettingsIncludes` already existed, and the SUITE tier already used it; the narrow tier did not call it.

Fixture: `gradlew`, `settings.gradle` containing `include ':app'`, `app/build.gradle`, `lib/build.gradle`, `lib/src/main/java/A.java`, `lib/src/test/java/ATest.java`.

```
before  VERIFY: ./gradlew :lib:test --tests 'ATest'
          (from lib/build.gradle + gradlew + ranked test class) tier=narrow
after   VERIFY: none derivable (no build manifest found) - syntax-check the file you edited and stop.
```

Gradle's answer to the emitted command is `Project 'lib' not found in root project`.

With the same tree and `settings.gradle` amended to `include ':app'` + `include ':lib'`:

```
after   VERIFY: ./gradlew :lib:test --tests 'ATest'
          (from lib/build.gradle + gradlew + settings.gradle + ranked test class) tier=narrow
```

Both tiers now share one notion of what declares a project, and the narrow tier takes the suite tier's three answers with `--tests` on them:

| tree | command |
|---|---|
| directory carries its own settings script | `./gradlew -p lib test --tests 'ATest'` |
| root settings declares the project | `./gradlew :lib:test --tests 'ATest'` |
| root project | `./gradlew :test --tests 'ATest'` |
| nothing declares it | silence |
| `project(':lib').projectDir = file('other')` | silence |

The remap case is the one that is not merely loud: the command would run, and pass, about code the edit never touched.

The accepted trade is the one the suite tier already accepted: a settings script that COMPUTES its includes declares nothing either predicate can read, so the tier goes silent where it used to emit.

### #225 / #205 — Rake namespace scope

Rake scopes a task declared inside `namespace :foo do ... end` as `foo:test`. None of the three declaration patterns knew that.

Fixture Rakefile:

```ruby
namespace :foo do
  task :test do
    sh 'ruby -Ilib -e "require \"invoice\""'
  end
end
```

```
before  VERIFY: rake test          (from Rakefile test task (whole suite)) tier=suite
after   VERIFY: rake foo:test      (from Rakefile foo:test task (whole suite)) tier=suite
```

Both commands run against the fixture:

```
$ rake test
rake aborted!
Don't know how to build task 'test' (See the list of available tasks with `rake --tasks`)

$ rake foo:test
ruby -Ilib -e "require \"invoice\""
```

#225 notes that deciding scope means tracking block structure. It does, so the change tracks it rather than approximating: a line scanner keeping a stack of open blocks, naming each frame a `namespace` statement opens. Openers come out of one quote-aware pass as an ordered list rather than a count, so `namespace :foo do namespace :bar do` — which Ruby accepts and which `rake -AT` reports as `foo:bar:test` — names both frames. Comments and string bodies are removed before keywords are counted; `if`/`unless`/`while`/`until`/`for` count only as a line's first word, so the modifier forms open nothing; a word after `.`, `:`, `@` or `$` is not a keyword. What it does not read is a heredoc body, a brace-delimited block, and a `namespace` written with braces — and that is what the balance check is for: any file whose blocks do not balance reports no namespace anywhere, which is exactly the previous behaviour. A namespace whose name is not a literal (`namespace NAME do`) is reported unreadable and degrades the same way. A desync degrades to the old answer rather than inventing a namespace, and never to a half-qualified one.

A declaration that merely looks nested — indented inside `if`, `begin`, a `def` — is still top level and keeps `rake test`; a top-level declaration wins over a namespaced one.

The in-code comment on `searchVerifyRakeTestTaskGeneratorPattern` claiming #205 was "behaving exactly as it did" is no longer true and has been corrected in place.

### #224 — workspace globs need brace expansion

`workspaces` globs are micromatch; the matcher is `path.Match`, whose language is smaller. The issue's table shows both sides of the unreadable-pattern default are wrong for some repository, and names the fix that is free: expand the braces.

Fixture: root `package.json` with `"workspaces": ["{packages,tools}/*"]`, a `yarn.lock`, and a leaf `standalone/` that is **not** a member.

```
before  VERIFY: cd standalone && yarn test
after   VERIFY: cd standalone && npm test
```

Yarn rejects the `before` command: the leaf is not one of its workspaces.

The member direction is unchanged, on the same repository, with a leaf that **is** a member of the second alternative:

```
before  VERIFY: cd tools/scratch && yarn test
after   VERIFY: cd tools/scratch && yarn test
```

Bounded to what the manifest format allows, not to bash: only alternation (a group with a top-level comma) is expanded, nesting included, capped at 64 patterns. A sequence range `{1..3}`, a single-alternative group, an unbalanced brace and the extglob groups are all declined and keep the permissive answer — the residue the issue accepts, stated as such.

### #221 — docs-and-fixtures swallows program text

`searchDocsSectionClass` admitted the whole fixture class, which `classifySearchFile` assigns by directory segment. A `.go` file under `testdata/` was labelled into a section whose own doc comment says it holds no program text, excluded from related-site expansion and every enrichment block. This repository's own `internal/sem/testdata` is that case.

Fixture repository holding `internal/sem/parser.go`, `internal/sem/testdata/case.go` (program text) and `pkg/golden/report.go` (ordinary source under a directory named `golden`):

```
before
  1. internal/sem/parser.go:2-6 ParseUnaryOperand
  VERIFY: go test ./internal/sem
  DOCS & FIXTURES (matched the query; not fix sites):
  2. pkg/golden/report.go:4 RenderUnaryOperandReport
  3. internal/sem/testdata/case.go:8 ParseUnaryOperandFixture
  4. internal/sem/testdata/case.go:4 UnaryOperandCase

after
  1. internal/sem/parser.go:2-6 ParseUnaryOperand
  2. pkg/golden/report.go:2-6 RenderUnaryOperandReport   signals=...,fixture-prior
  3. internal/sem/testdata/case.go:8 ParseUnaryOperandFixture
  4. internal/sem/testdata/case.go:4 UnaryOperandCase
  VERIFY: go test ./internal/sem
```

Both halves, as the issue requires. The section now keeps a fixture only when it is a RECORDING — a fixture extension, a path `NonProgramTextPath` already answers for, or a file no parsed language claims — so `.snap`, `.golden` and `.ambr` are unaffected and the ruff snapshot case the class was built for is unchanged. Returning program-text fixtures to primary puts them in front of `searchVerifySubjectFor`, which reads primary and for which `searchTestArtifactPath("internal/sem/testdata/case.go")` is true; the paired half declines a fixture-class path as a test to run, in the primary and covering-test branches both. Keying on the fixture CLASS rather than on test trees in general is what keeps `internal/sem/parser_test.go` and every ordinary `tests/` file adopted exactly as before. The demotion prior is untouched — a fixture still ranks below the implementation, as `fixture-prior` on hit 2 shows.

---

## Tests

Failing-first, at runtime on `origin/main` (the two new-API unit tables are additional):

```
--- FAIL: TestSearchVerifyGradleNarrowTierNeedsSettingsInclusion
    a_settings_script_that_declares_a_SIBLING_declines: command = "./gradlew :lib:test --tests 'ATest'", want ""
    no_settings_script_at_all_declines:                 command = "./gradlew :lib:test --tests 'ATest'", want ""
    a_directory_carrying_its_own_settings_script...:    command = "./gradlew :lib:test --tests 'ATest'", want "./gradlew -p lib test --tests 'ATest'"
    a_remapped_project_declines...:                     command = "./gradlew :lib:test --tests 'ATest'", want ""
--- FAIL: TestSearchVerifyRakeNamespacedSuiteCommandRuns
    a_namespaced_task_is_advertised_by_its_qualified_name: command = "rake test", want "rake foo:test"
--- FAIL: TestSearchVerifyWorkspaceBraceExpansion
    a_package_the_expanded_pattern_does_not_reach_falls_back_to_npm:  command = "cd standalone && yarn test", want "cd standalone && npm test"
    a_nested_group_that_does_not_reach_the_leaf_falls_back_to_npm:    command = "cd packages/db/core && yarn test", want "cd packages/db/core && npm test"
--- FAIL: TestSearchProgramTextFixtureStaysPrimary
    a_Go_fixture:         searchDocsSectionPath("internal/sem/testdata/fixtures/go-basic/auth.go") = true, want false
    a_TypeScript_fixture: searchDocsSectionPath("internal/sem/testdata/fixtures/typescript-basic/api.ts") = true, want false
    a_Python_fixture:     searchDocsSectionPath(".../fixtures/flake8_simplify/SIM201.py") = true, want false
    ordinary_source_under_a_directory_named_golden: searchDocsSectionPath("pkg/golden/report.go") = true, want false
--- FAIL: TestSearchVerifyNeverRunsAFixtureAsATest
    testPath = "internal/sem/testdata/case.go", want none: a fixture is not a test to run
```

Every table pins both directions: the emit and the decline for Gradle, the qualified and the bare task name for Rake, the member and the non-member leaf for the workspace globs, recordings and program text for the fixture section, and a fixture versus an ordinary test file for the VERIFY subject.

Two pre-existing tests pinned behaviour these issues call wrong and were updated rather than worked around:

- `TestDeriveSearchVerifyCommandFromBuildEvidence/gradle needs the wrapper and a test class` asserted `./gradlew :lib:test --tests 'ATest'` for a tree with no settings script at all — the exact command #219 reports as unrunnable. It now supplies `settings.gradle.kts`, and a sibling case pins the decline when it is removed.
- `TestSearchVerifyCommandsDoNotExecuteRepositoryData` derived its two Gradle fixtures from trees with no settings script, so they stopped deriving a command and the injection guard stopped running. Both now declare their project; the malicious-directory case routes through the `-p` spelling, which guards the new `shellQuotePath` argument surface that the project-path spelling does not have.

## Gate

```
gofmt -s -l .            empty
go vet ./...             clean
go build ./...           clean
go test ./...            ok  (all 9 packages; sem 176s, cli 54s)
```


suhaanthayyil:
### Finding 01M2ZV8DBEGR (Medium) — upheld and fixed in 2ec6e8f1

Reachable, and worse than the case the scanner was added for. Confirmed against `rake 13.0.6`:

```
$ cat Rakefile
namespace :foo do namespace :bar do
  task :test do
    puts "RAN foo:bar:test"
  end
end end

$ ruby -c Rakefile
Syntax OK

$ rake -AT
rake foo:bar:test  #

$ rake foo:test
rake aborted!
Don't know how to build task 'foo:test' (See the list of available tasks with `rake --tasks`)
Did you mean?  foo:bar:test
```

The finding is exactly right about the mechanism: blocks balance numerically, so the balance guard did not catch it, and the derivation was **confident and wrong** — which is the failure mode this PR exists to end, not an instance of the "degrades to the previous answer" property it claims.

`entire-graph query` on that fixture, before and after:

```
before  VERIFY: rake foo:test        <- rake: Don't know how to build task 'foo:test'
after   VERIFY: rake foo:bar:test    <- RAN foo:bar:test
```

**The fix.** `searchVerifyRubyLineBlocks` replaces the split between a `^`-anchored regex over the raw line and a keyword count over a comment-stripped copy. One quote-aware pass now returns the line's openers as an ordered **list**, each carrying the name of the `namespace` whose `do` opened it, so names pair with blocks in token order however many share a line. Reading the name and the keywords in the same pass is what lets `namespace "bar" do` keep a name held inside a string literal while `desc 'run to the end'` still does not close a block.

**Second defect found while fixing it, same class.** A name that is not a literal — `namespace NAME do`, `namespace compute_name do` — was silently treated as absent, which half-qualified every task below it. It is now reported as unreadable and joins the unbalanced-blocks case in degrading the whole file to no namespace:

```
$ cat Rakefile
NAME = :foo
namespace NAME do
  task :test do ... end
end

after   VERIFY: rake test
$ rake test
rake aborted!
Don't know how to build task 'test'
```

Loud failure, not a wrong suite. That restores the stated safety property for both guards.

**Tests**, failing-first at runtime on the previous PR head:

```
--- FAIL: TestSearchVerifyRakeNamespacedSuiteCommandRuns/both_namespaces_opened_on_one_line_are_qualified
    command = "rake foo:test", want "rake foo:bar:test"
--- FAIL: TestSearchVerifyRakeNamespacedTaskIsQualified/two_namespaces_opened_on_one_line
    = "foo:test", want "foo:bar:test"
--- FAIL: .../three_namespaces_opened_on_one_line            = "a:test",  want "a:b:c:test"
--- FAIL: .../a_quoted_namespace_opened_second_on_the_same_line = "foo:test", want "foo:bar:test"
--- FAIL: .../a_namespace_opened_after_another_block_on_the_same_line = "test", want "foo:test"
```

Both directions, as before: the multi-namespace line and the unreadable-name degrade are pinned alongside the existing table, which still passes unchanged — one namespace per line, nested namespaces on separate lines, a namespace closed before the declaration, top-level-wins, and the four "indented but still top level" cases (`if`, `begin`, `def`, a namespace that already closed).

One limit worth stating rather than leaving implied: an unreadable name or an unbalanced file degrades the **whole file** to no namespace, including lines above the problem. That is deliberate — a partial stack is how this bug happened.

**Gate:** `gofmt -s -l .` empty · `go vet ./...` clean · `go build ./...` clean · `go test ./internal/sem ./internal/cli` ok (161s / 38s).

MuskanPaliwal:
hey @suhaanthayyil , i noticed the gradle fix here overlaps with #249, which i’d opened for #219. i’d put a fair bit of effort into that pr, so i’d really appreciate a heads-up when there’s overlapping work like this. totally okay if a different approach makes more sense.

a few cases from mine might be useful here too. both fix the original issue, but this branch still falls back to root or parent tests for an undeclared child. #249 prevents that and also covers renamed projects and intermediate build roots. i’ve also pushed a fix for a nested-wrapper bug i found while comparing them, and the full checks pass.

could we look at #249 alongside this one before deciding how to handle the gradle changes? maybe we could merge that first and keep the other fixes here. i’m happy to discuss whichever approach fits best.

